PR #28, opened 2019-12-19
Add algorithm

pavlis:
Ian please have a look at this branch and see if you think this structure is appropriate for future evolution.  I've added an "algorithms" directory and associated library.   The idea of that directory is a place top put C++ processing algorithms that are wrapped and then easily adapted to a map/reduce processing with a python front end (i.e. the run would use a python script that would do little more than call the C++ function and deal with the output.  

Note this branch is totally untested other than I know it compiles under macos.   The agc function probably works as it is a minor perturbation from an antelope seispp program I've run on a lot of data.   The only change was another feature this will test - the version here posts errors to the ErrorLogger object contained in Seismogram.  I discovered some obvious holes in the C api when I did that.  Those changes WILL need to be assimilated into master even if you think we should organize this differently.  See what you think

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,7 @@ dist: bionic
 jobs:
   include:
     - name: "Docker Compose"
-      sudo: required
-      language: minimal
+      language: shell
       services: 
         - docker
       before_install: 
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ before_install:
   - |
     if [ $TRAVIS_OS_NAME == "linux" ]; then
       pyenv versions
-      pyenv global 3.7.1
+      pyenv global 3.7
     fi
   - cd cxx 
   - mkdir build 
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,6 +70,10 @@ RUN cd /mspass/cxx \
     && make \
     && make install 
 
+# Add setup.py to install python components
+ADD setup.py /mspass/setup.py
+RUN pip3 install /mspass -vvv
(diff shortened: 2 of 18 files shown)


wangyinz:
I think you did not check in all the files here. I can only see a `algorithm.h` file that is new here. The build errors out because of missing `CMakeLists.txt`. I did changed the Travis file because my old configuration tied to a specific version of Python that got updated by Travis, so it passed to original error. 

That organization is exactly what we originally designed, so I think we have no problems there.

pavlis:
I did just pushed the cmake configuration file from the algorithm directory - woops.  Not sure how I missed that.   Hopefully this will do it.  

Did you merge this with master already?  Might make a second merge tricky

wangyinz:
I think the corresponding algorithm.cc file is still missing. 

The merge comment is because of the Travis updates that breaks both this branch and the master branch. I have updated the mater branch, and merged that change to this branch, too. It shouldn't affect any of the subsequent merges.

wangyinz:
and it seems the CMakeList.txt you added is actually not the one expected for the header file, so it throws error again: https://travis-ci.com/wangyinz/mspass/jobs/269082928

pavlis:
Woops, did miss the agc.cc file and related cmake configuration file.  Hopefully fixes my earlier mistake.   Always a problem adding a new directory in git as git status doesn' show files in other directories.

wangyinz:
btw, I think this one is ready to merge. We probably want to do a "squash and merge" here to keep the history clean.

PR #29, opened 2019-12-22
Eliminate nested

pavlis:
Ian can you have a look at this and see what you think of these changes.  If you think this is ok merge some or all of this to master.  Note that might not be trivial as master has evolved a bit since I forked this from it.   

There are two different changes here:

1. I gave up on getting an answer on the pybind11 issues question about nested classes.  Additional reading convinced me it is probably just not supported because it appears there may be some fundamental clashes with python.   I also realized I was being a purist and there was no real reason for requiring a nested class.  The only reason is the ErrorLogger relation to TimeSeries and Seismogram was a classic example of a "has a" relationship.   It is now an "is a" because I simply made ErrorLogger a parent to MsPASSCoreTS.  TimeSeries and Seismogram now have a pretty complex inheritance diagram with three trunk branches:  one from BasicMetadata, one from BasicTimeSeries, and now one from ErrorLogger.  There is a revised wrapper definition that seems to work based on my test script (more on that below)
2.  I have a partial implementation of an obspy2mspass converter.  It is partial because right now it is only a converter from Trace objects to TimeSeries.   Next step is a converter for three component data.  That presents a slightly different problem as there need to be some rational options for handling orientation.  obspy Trace does not demand orientation information be stored in the Stats header.   I want to make an optional "cardinal" argument that says assume assume you are give three Trace objects in E,N,Z order that are in standard coordinates.   There is also the issue of where you get the optional metadata from when you have three trace objects and the bundle object (can't remember the name right off).   

Finally, I got hung up on this setup for python thing and couldn't figure out how to implement my test scripts into the repository.   You will find these:

1. in mspass/python is the obspy2mspass.py file describe above.
2.  in mspass/python/tests are a set of files that are the test scripts I have been beating on.   They work for me when all these files are copied to mspass/cxx/build/python, but they need to be restructured to make them standard tests.   Look at the README file there for a few more details.

One final note about the tests directory.  There is a crude and partial file to build a test MetadataDefinitions object using a pf format.  Something we need to do soon is standardize that file and put it into the yml format you have (correctly I think) proposed for the schema.   We most definitely need to start filling out the namespace.   Hard for me, in fact, to do the conversion to Seismogram without a decision on how to define orientation - presume we will use hang and vang ala css3.0 as I think iris web services use the same symbols.

### DIFF
--- a/cxx/include/mspass/utility/ErrorLogger.h
+++ b/cxx/include/mspass/utility/ErrorLogger.h
@@ -22,7 +22,7 @@ public:
   \param alg is assigned to algorithm attribute.
   \param merr is parsed to fill the message and severity fields.
   Note p_id is always fetched with the system call getpid in the constructor.*/
-  LogData(int jid, std::string alg, mspass::MsPASSError& merr);
+  LogData(const int jid, const std::string alg,const mspass::MsPASSError& merr);
   friend ostream& operator<<(ostream&, LogData&);
 };
 /*! \brief Container to hold error logs for a data object.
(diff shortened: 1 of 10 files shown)


wangyinz:
The code looks good to me and they also past the CI, so I guess it is fine to merge them. However, I don't really want to give up on doing the nested class yet. I was looking into the issue, and do find a couple discussions related (e.g. [this](https://github.com/RosettaCommons/binder/issues/24#issuecomment-428399434), [this](https://github.com/pybind/pybind11/issues/210) and [this](https://stackoverflow.com/questions/48372537/pybind11-for-c-code-with-inner-struct-created-via-static-factory-method)). I am not sure if you have seen these links already, but it appears there has to be a way to do it. We don't want to break our original design of relation among the objects unless absolutely have to. Let me take a closer look before we decide to merge this branch.

pavlis:
After a fix was found for handling nested classes this approach was abandoned.  I am closing this pull request.

PR #34, opened 2019-12-30
Add yaml parser

pavlis:
This seems to work to parse a yaml file defining MetadataDefinitions.   There are a couple of residual issues before we should merge this with master:

1. The prototype yaml file can be found in mspass/data/yaml/mspass.yaml.   The structure contains a possibly unnecessary "Attributes:" at the top, but (a) I couldn't figure out how to make yaml-cpp work with a plain list and (b) this might not be wrong anyway.   I'm thinking we might want to change that keyword to CoreAttributes and then add something like "Extensions" that could be parsed if it wasn't empty.   The current code doesn't have that, but it might be smart to help separate user supplied and core definitions.   
2.  I created a simplied constructor with just a simple string that could be called like this:
mdef=MetadataDefinitions("mspass")
It works BUT it requires setting a MSPASS_HOME environment variable.   I used something similar before, but not in a manner like this one that will be a nearly universal file to load to run mspass.  I say that because all mongodb access should probably use this object to get types consistent with the database and all get and puts of metadata in python should use it to access the C code cleanly.   The alternative is to load this from the db, but that is probably better done as an alternate constructor when we have a better handle on db interactions.   If you accept this idea, we need a way to have cmake copy the data/yaml directory to the install directory - to me that is the only rational place to have this copied.    Hope that isn't to confusing.


cxx/src/lib/utility/MetadataDefinitions.cc
```diff
@@ -1,26 +1,22 @@
...
+using namespace std;
 MetadataDefinitions::MetadataDefinitions(const std::string mdname)
 {
     try{
-        /* This s temporary.  For now this defaults to pf format file input */
-        MetadataDefinitions tmp(mdname,MDDefFormat::PF);
-        *this=tmp;
+      string datadir=mspass::data_directory();
+      string path;
+      path=datadir+"/yaml/"+mdname+".yaml";
```
wangyinz: if a user explicitly give the full path in the input `mdname`, isn't it gonna break?

cxx/include/mspass/utility/MetadataDefinitions.h
```diff
@@ -28,16 +28,17 @@ data file.
 class MetadataDefinitions
 {
 public:
-  /*! Default constructor - defaults to mspass namespace. */
-  MetadataDefinitions();
+  /*Comment out default constructor for now - needs to default to mspass 
+   * and not sure how to do that.
+  MetadataDefinitions();*/
```
wangyinz: I guess this is already done by the the default of following constructor `MetadataDefinitions(const std::string mdname=std::string("mspass"))`, right?

cxx/python/mspass_wrapper.cpp
```diff
@@ -437,10 +437,11 @@ PYBIND11_MODULE(mspasspy,m)
   );
   py::enum_<mspass::MDDefFormat>(m,"MDDefFormat")
     .value("PF",MDDefFormat::PF)
-    .value("SimpleText",MDDefFormat::SimpleText)
+    .value("YAML",MDDefFormat::YAML)
   ;
   py::class_<mspass::MetadataDefinitions>(m,"MetadataDefinitions","Load a catalog of valid metadata names with types defined")
     .def(py::init<>())
```
wangyinz: Is this default constructor already commented out in the above header `MetadataDefinitions.h`?

wangyinz:
> 1. The prototype yaml file can be found in mspass/data/yaml/mspass.yaml.   The structure contains a possibly unnecessary "Attributes:" at the top, but (a) I couldn't figure out how to make yaml-cpp work with a plain list and (b) this might not be wrong anyway.   I'm thinking we might want to change that keyword to CoreAttributes and then add something like "Extensions" that could be parsed if it wasn't empty.   The current code doesn't have that, but it might be smart to help separate user supplied and core definitions.

I think that top level *Attributes* is necessary as that servers as the key of the following sequence item.  I agree that we need a better name for it. Instead of *CoreAttributes*, maybe we could just name it something simple as *Core*. The *Extensions* is definitely needed.

> 2. I created a simplied constructor with just a simple string that could be called like this:
>    mdef=MetadataDefinitions("mspass")
>    It works BUT it requires setting a MSPASS_HOME environment variable.   I used something similar before, but not in a manner like this one that will be a nearly universal file to load to run mspass.  I say that because all mongodb access should probably use this object to get types consistent with the database and all get and puts of metadata in python should use it to access the C code cleanly.   The alternative is to load this from the db, but that is probably better done as an alternate constructor when we have a better handle on db interactions.   If you accept this idea, we need a way to have cmake copy the data/yaml directory to the install directory - to me that is the only rational place to have this copied.    Hope that isn't to confusing.

Yes, I understand the issue. I guess we definitely don't want this fundamental schema definition to be easily (accidentally) changed by a user. I think it is fine to have it in a home directory for now, but in the future, we do want a more secure place to put it. Since the core schema should not expect much changes anyway, I wonder if we could even hard-code it in the source code. I guess that is OK, since we could always provide a way for the user to safely alter it after a MetadataDefinition object is created in the memory (although that should be considered a dangerous operation). In this way, we could have the *Extensions* stored in separate files that users can specify.

cxx/include/mspass/utility/MetadataDefinitions.h
pavlis: Yes, this is the issue I wasn't sure how to resolve so I pushed it up for now.   Let me experiment with this a bit today.   I think there may be a simple solution, but I need to test my idea.   Glad you caught this as I had forgotten this little titbit I had abandoned when trying to catch bigger fish.

cxx/python/mspass_wrapper.cpp
pavlis: Yes, good point but I do plan to restore the default constructor.   The goal is to have the default constructor automatically load mspass

pavlis:
made some changes so am closing this pull request.  A new one will appear momentarily.

PR #35, opened 2019-12-31
Add yaml parser

pavlis:
This update implements a feature to allow multiple group tags in the MetadataDefinitions file mspass.yaml.    Uses the keys only for looping and discards any memory of the actual names.  Did not update the mspass.yaml file to reflect this change because we have a bigger job there and need to resolve a couple issues before finalizing it.  For now, check this over and merge this when you are sure it passes all tests.


wangyinz:
Note this is the new pull request created from #34.

PR #38, opened 2020-01-10
Add yaml parser

pavlis:
This is a much improved prototype for the MetadataDefinitions and a new yaml file structure to build the schema.   It passed my tests.  Assuming Travis builds it with no issues, I think i should be assimilated into master.

You may run a test of the new features with this short script:
from mspasspy import MetadataDefinitions
mdef=MetadataDefinitions()
print("testing readonly on chan")
print(mdef.readonly("chan"))
print("testing is_normalized method on sta")
print(mdef.is_normalized("sta"))
print("testing normalization getters")
print("unique_id_key for sta=",mdef.unique_id_key("sta"))
print("return of table method for key of sta=",mdef.table("sta"))
print("return of method normalize_data for key=sta is ",mdef.normalize_data("sta"))


cxx/include/mspass/utility/MetadataDefinitions.h
```diff
@@ -104,20 +105,147 @@ class MetadataDefinitions
...
+  The idea is to first ask for the unique_id_key and then ask what table (collection)
+  contains the key returned by unique_id_key.   This provides a fast and
+  convenient lookup for normalized data.
+
+  \param key is the normally the return from unique_id_key
+  \return string defining the table(collection) the key can be used for locating the
+     unique tuple/document required to access related Metadata.  String will be
+     empty if the search fails.
+  */
+  string table(const string key) const;
```
wangyinz: Tha `table` is a relational database concept, and the `normalize` above is a MongoDB thing. This could be confusing to the users. I think we should either use `collection` & `normalize` or `table` & `join` so the concepts are consistent. I guess the question is whether we want to hide the new concepts of MongoDB with the more familiar relational ones.

cxx/src/lib/utility/MetadataDefinitions.cc
```diff
@@ -202,9 +208,116 @@ MetadataDefinitions& MetadataDefinitions::operator+=(const MetadataDefinitions&
...
+      this->unique_id_data.insert(*uptr);
+    }
   }
   return *this;
 }
+/* We assume the readonly list is smaller so the set of defined
+keys is those for readonly.  That inverts the logic of this function.
+i.e. roset contains keys of attributes marked readonly and this function
+is the not of that logic. */
+bool MetadataDefinitions::writeable(const string key) const
```
wangyinz: Another thing I realized while reading the code is that we may want to have additional code to handle the situation of the argument `key` being an alias.

cxx/include/mspass/utility/MetadataDefinitions.h
pavlis: Excellent point.  Easily fixed.  Will do so immediately and push the changes.

cxx/src/lib/utility/MetadataDefinitions.cc
pavlis: Not sure what you mean here?  MetadataDefinitions already has a set of methods for handling aliases.  Do you mean we should have a way to lock data stored with an alias?   That could get real tricky.   

This brings up a point worth emphasizing that may be relevant here.   I think the idea should be that dbmaster (normlized and readonly) attributes should not necessarily be treated such dogmatically.   The storage cost of duplicating such parameters in a wf document in intermediate results may not be justified.  Using mongo jargon the same attributes could be used in an embedded data model if we ignore the readonly flag.  I can conceive a need for three distinct readers that might be added as needed:

1. The first thing we should write is the normalized reader that uses the id's to do a find on documents with names like site and source to load attributes.  I would be dogmatic about that approach.
2.  We could have the complete reverse with a pure embedded model reading the same attribute from wf.   
3.  We could build a bombproof hybrid:   try first to always read an attribute from wf.  When that fails, use the normalized model.   

All readers need to exploit the ErrorLogger feature and never fail.   Callers should check the status of the error log and handle different levels of error differently.

pavlis:
Change from table to collection is done and just pushed.   

We have another branch pending that will definitely have conflicts with this one.   I think this one is now good to merge.   If you do that, I will take care of resolving the conflicts with the other branch pending.   I know exactly what they are and can resolve the differences quickly.

cxx/src/lib/utility/MetadataDefinitions.cc
wangyinz: Well, I definitely did not state the comment clearly. What I mean is that, for example, a call of `writeable(sta)` and `writeable(station)` should return the same value. In current implementation, the code only looks up the key in `roset`, which may not have aliases defined. This is really just a minor change to make the code more robust. 

And I couldn't agree more with this addition to the reader. Actually, I was concerned about having a dbmaster as it might impair the flexibility for some cases that we do not anticipate (although it is already more flexible than the one in antelope). Then, this new design addressed the problem perfectly. It makes possible building a quick and easy workflow without caring about the dbmaster at all.

pavlis: Ahh - I see.  Good point.  That should be pretty easy to fix.   I'll do so as soon as I resolve the likely conflicts between master with this branch and the one mentioned below.

PR #47, opened 2020-02-09
Cxx add historychaining

pavlis:
Have a look at this and see if you think this is an appropriate initial solution for processing history preservation.   I'm going to expand the concepts in the issues section on this topic, but look here for the implementation.

### DIFF
--- a/cxx/include/mspass/utility/Metadata.h
+++ b/cxx/include/mspass/utility/Metadata.h
@@ -337,6 +337,31 @@ other attributes.
   {
     this->put<long>(key,val);
   };
+  /*! Create or append to a chained string.
+   *
+   A chain conceptually is identical to a list of string data.  
+   We implement it in Metadata because sometimes (e.g. MongoDB 
+   interaction and some constructs like the unix shell PATH variable) 
+   handling a full scale container like list<std::string> would be
+   awkward.   If more extensive capability like that is needed it would
+   be better to add a class that inherits Metadata and does so.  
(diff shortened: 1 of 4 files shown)


wangyinz:
I was working on the Python code to build a Spark workflow. Then, I got an error when using the `obspy2mspass` routine. It says:
```
File "/usr/local/lib/python3.6/dist-packages/mspasspy/importer/obspy.py", line 68, in obspy2mspass
    dout.elog.set_algorithm("obspy2mspass")
AttributeError: 'mspasspy.ccore.ErrorLogger' object has no attribute 'set_algorithm'
```
The `set_algorithm` is definitely missing in current master branch. I was thinking this might be in this branch, as I never really get back on this history chaining issue. Turns out this is not the case. I wonder if you have any locally uncommitted code somehow.

pavlis:
I think that was a version skew problem.  That python function was written before we decided how to handle this issue.   Actually every single call to elog.log_error would throw an exception because the api changed.   I pushed the changes I made untested.  I suspect you can fix any residual problems you might find.  This is one of those dark sides of python not being compiled.  A compiler would have caught these errors.   

This brings up another issue about that particular script and others.  Another thing I added after this script and all the C++ algorithm functions were placed in the repository was the history preservation feature.   Recall as script should first register each function that will be applied to the dataset.  Then each function is supposed to post to some Metadata field that I don't recall right now.  I'll post something in the right issues section when I find it.

PR #48, opened 2020-02-17
Pickle ccore

pavlis:
Ian, I think this branch is now ready to merge.   pickle is working now for both Seismogram and TimeSeries objects with the caveats you will see in the issues section on serialization issues.  

Here is a little test script you can use to validate this - you will need to modify the sys.append.path as I'm using that approach right now for handling the ccore library.  Also you may need to set LD_LIBRARY_PATH to the location of boost libraries.    Here is the script:
```
import sys
sys.path.append('/home/pavlis/src/mspass/python')
from mspasspy.ccore import CoreSeismogram
d=CoreSeismogram(200)
d.put_double('delta',1.0)
d.put_double('dt',1.0)
d.put('npts',200)
d.ns=200
d.t0=100.0
d.live=True
from mspasspy.ccore import Seismogram
d2=Seismogram(d,'invalid')
import pickle
x=pickle.dumps(d2)
print("pickle succeeded")
print("trying to restore")
d3=pickle.loads(x)
print("pickle loads completed")
print("data npts=",d.get_int('npts'))
print('same stored in struct of BasicTimeSeries=',d.ns)
print('data t0=',d.t0)
print('Now testing pickle for TimeSeries data')
from mspasspy.ccore import TimeSeries
from mspasspy.ccore import CoreTimeSeries
d0=CoreTimeSeries(500)
d0.live=True
d0.dt=1.0
d0.ns=500
d=TimeSeries(d0,'invalid')
s=pickle.dumps(d)
print('Pickle dumps succeeded')
print('size of string returned by pickle=',len(s))
print('Trying loads')
dr=pickle.loads(s)
print('Finished - BasicTimeSeries attributes in output')
print('ns=',dr.ns)
print('dt=',dr.dt)
print('t0=',dr.t0)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ jobs:
           packages:
             - gfortran
             - libboost-dev
+            - libboost-serialization-dev
             - liblapack-dev
       compiler: gcc
       
@@ -50,6 +51,7 @@ jobs:
           packages:
             - gfortran
             - libboost-dev
+            - libboost-serialization-dev
             - liblapack-dev
       compiler: clang
       
@@ -71,6 +73,7 @@ jobs:
           packages:
             - gfortran
             - libboost-dev
+            - libboost-serialization-dev
             - liblapack-dev
             - doxygen
       compiler: gcc
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN apt-get update \
        python3-dev python3-pip \
        openjdk-8-jdk \
        git cmake gfortran gdb \
-       liblapack-dev libboost-dev libyaml-dev \
+       liblapack-dev libboost-dev libboost-serialization-dev libyaml-dev \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ RUN pip3 --no-cache-dir install pymongo
 RUN mkdir /home/data
 
 # Prepare the environment
-ENV SPARK_VERSION 2.4.4
+ENV SPARK_VERSION 2.4.5
 ENV SPARK_MASTER_PORT 7077
 
 ENV MSPASS_ROLE master
--- a/cxx/CMakeLists.txt
+++ b/cxx/CMakeLists.txt
@@ -31,8 +31,7 @@ message (STATUS "YAML_CPP_LIBRARIES   = ${YAML_CPP_LIBRARIES}")
 message (STATUS "YAML_CPP_INCLUDE_DIR = ${YAML_CPP_INCLUDE_DIR}")
 
 
-#find_package (Boost 1.64.0)
-find_package (Boost 1.64.0 REQUIRED COMPONENTS serialization )
+find_package (Boost 1.64.0 COMPONENTS serialization)
 if (NOT Boost_FOUND)
   message (STATUS "Building Boost")
   include (cmake/boost.cmake)
@@ -41,9 +40,8 @@ if (NOT Boost_FOUND)
     ${PROJECT_BINARY_DIR}/boost
     )
 endif ()
-message (STATUS "BOOSTROOT="${BOOSTROOT})
 message (STATUS "Boost_LIBRARIES="${Boost_LIBRARIES})
-message (STATUS "Boost_LIBRARY_DIR_RELEASE="${Boost_LIBRARY_DIR_RELEASE})
+message (STATUS "Boost_LIBRARY_DIRS="${Boost_LIBRARY_DIRS})
 message (STATUS "Boost_INCLUDE_DIRS="${Boost_INCLUDE_DIRS})
 
 find_package (BLAS)
--- a/cxx/CMakeLists.txt
+++ b/cxx/CMakeLists.txt
@@ -40,9 +40,9 @@ if (NOT Boost_FOUND)
     ${PROJECT_BINARY_DIR}/boost
     )
 endif ()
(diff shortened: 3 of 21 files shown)


wangyinz:
I have not got the script tested yet, as I've been fixing the cmake files. I think I have all the issues address, and the C++ code should build on systems with or without boost.serialization installed. You can take a look at what I changed. Basically, it will first try to find boost.serialization in the default paths of the system. If found, it will compile and link the libboost-serialization as shared library with rpath embedded, which will show up in the `ldd` command. If not found, it will download and compile the boost library with serialization enabled and link the static library with the `fPIC` flag. Then, the static libboost-serialization library can be correctly linked to the ccore python module (shared library), in which case it will not show the serialization library in the `ldd` command. Please try compile it locally on your machine and let me know if it works as expected.

wangyinz:
Just confirmed the test scripted worked fine for me.

pavlis:
Well, that procedure didn't work for me.   I pulled the pickle_ccore branch and ran cmake.  I got this output that hints at an issue but probably is just the informational message being posted:
```
CMake Warning (dev) at CMakeLists.txt:43:
  Syntax Warning in cmake code at column 35

  Argument not separated from preceding token by whitespace.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:44:
  Syntax Warning in cmake code at column 38

  Argument not separated from preceding token by whitespace.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:45:
  Syntax Warning in cmake code at column 38

  Argument not separated from preceding token by whitespace.
This warning is for project developers.  Use -Wno-dev to suppress it.

-- YAML_CPP_LIBRARIES   = yaml-cpp
-- YAML_CPP_INCLUDE_DIR = /home/pavlis/src/mspass/cxx/build/yaml-cpp/lib/cmake/yaml-cpp/../../../include
-- Boost 1.64.0 found.
-- Found Boost components:
   serialization
-- Boost_LIBRARIES=Boost::serialization
-- Boost_LIBRARY_DIRS=
-- Boost_INCLUDE_DIRS=/opt/boost/include
-- A library with BLAS API found.
-- A library with BLAS API found.
-- A library with LAPACK API found.
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.6.9", minimum required is "3") 
-- Found PythonInterp: /usr/bin/python3 (found version "3.6.9") 
-- PYTHON_LIBRARIES = /usr/lib/x86_64-linux-gnu/libpython3.6m.so
-- PYTHON_EXECUTABLE = /usr/bin/python3
-- PYTHON_INCLUDE_DIRS = /usr/include/python3.6m
-- pybind11_LIBRARIES = pybind11::pybind11
-- pybind11_INCLUDE_DIR= /usr/local/include
-- Configuring done
-- Generating done
-- Build files have been written to: /home/pavlis/src/mspass/cxx/build
```
Not sure what is generating those warning, but I don't think it is related to this issue.  

Now perhaps I need to do a proper install.   Should be running setup.py?   I just did the hack thing and from build/python ran this command:
```
cp ccore.cpython-36m-x86_64-linux-gnu.so  ~/src/mspass/python/mspasspy
```
when I tried to run the testpickle.py script I got the same message seen before about not resolving the serialization library:
```
python testpickle.py 
  File "testpickle.py", line 3, in <module>
    from mspasspy.ccore import CoreSeismogram
ImportError: libboost_serialization.so.1.71.0: cannot open shared object file: No such file or directory
```
ldd on the ccore file confirmed that.   Note I got the same result on the file in the build/python directory so something is not working right with this.   

I got this with ubuntu.   Were you using a mac?   May matter as the linkers in macos and linux are quite different.

pavlis:
Pushed a minor fix I should I put in yesterday with the C code for CoreTimeSeries.   Should not effect a decision to merge this branch.

pavlis:
Looking at buffer_info details and pybind11 I am not at all convinced it would be worth the effort to try to use that approach to serialize the dmatrix in the Seismogram object.   When I consider how I'd have to implement it, both cases will end up having to convert to some version of a string representation of the binary data.   I would be surprised if the overhead in pickle's serialization of a numpy array is any faster than boost's serialization.   They are pretty much doing the same thing with different formats as best i can judge without digging a lot deeper.  Hence, let's stick with this form now.   

I'm going to put this same message in the issues section - realize it should be there for closure on this topic.

wangyinz:
The syntax error was from these lines:
```
message (STATUS "Boost_LIBRARIES="${Boost_LIBRARIES})
message (STATUS "Boost_LIBRARY_DIRS="${Boost_LIBRARY_DIRS})
message (STATUS "Boost_INCLUDE_DIRS="${Boost_INCLUDE_DIRS})
```
I have the syntax fixed, and that shouldn't affect the build anyway.

Based on the log, it seems you have the boost serialization installed in the system, so the build that I defined is using that one as the first priority. In this case, it is going to link the shared library. However, it seems there must be something wrong (something different from a standard install) with this boost installation. That's why the CMake procedure cannot find the `Boost_LIBRARY_DIRS`. I wonder why you were still able to build it, as I think it should fail at the linking stage without that. I wonder if you have installed the boost library yourself and moved some of the library files. The setup I have passed Travis, so it should build on both [linux](https://travis-ci.com/wangyinz/mspass/jobs/288396577) and [mac](https://travis-ci.com/wangyinz/mspass/jobs/288396579), where you can see in the logs that the `Boost_LIBRARY_DIRS` are all identified by CMake. I think you can either install a standard boost library with `apt install libboost-serialization-dev`, or remove all boost related files in the PATH so that CMake can use what I defined in `boost-download.cmake`.

PR #56, opened 2020-04-13
unified the use of _id

wangyinz:


wangyinz:
This is already merged by #58

PR #63, opened 2020-07-15
ccore API changes

pavlis:
I think it is time to merge this branch to master.   I just checked in a C++ test program for the ProcessingHistory object that follows more of the guidelines you suggest for such tests.   It isn't perfect but it does test all the methods (I think - I think you have tools to show what it does and doesn't test).   Anyway, it is time we make this origin.   Suggest you (ian) work on fixing python code these changes break while I assimilate the deconvolution code that came from you original work in with the new api.   Decon operators will be a good core algorithm for timing tests.


wangyinz:
It seems I can just comment that out

wangyinz:
I fixed a couple minor issues, now the test_history code errors out with clang compiler somehow. If you look at the [output](https://travis-ci.com/github/wangyinz/mspass/jobs/361385973), it is the number of history record doesn't match after reduction call. Have you seen that before? Kind of weird that it works fine for gcc...

PR #67, opened 2020-08-31
Add graphics

pavlis:
I think this code is ready to be considered for assimilation into the master branch.   I implemented two different plotting objects with the same user API:  (1) SectionPlotter makes seismic reflection style plots and (2) SeismicPlotter makes plots in the normal seismology convention (time as the x axis).   These should be viewed as base classes for more advanced graphics that should be feasible through the maplotlib.pyplot engine.  

Note that although SectionPlotter and SeismicPlotter have the same public API, they are implemented quite differently.   SectionPlotter was built by (quite extensive) modification of the plotting code I found in the 
(apparently depricated) code from fatiando a terra.   SeismicPlotter has some efficiencies for handling 
irregular sample rate data that SectionPlotter lacks.   We may want to fix that some day, but normally 
section plotting doesn't use irregular start times anyway so the issue should be minor.   

SeismicPlotter has only a stub for now for normalization of data.   We need a set of standard functions for doing amplitude adjustments anyway (e.g. peak and rms normalizations) so I would prefer to fix that after that is stabilized in ccore.

### DIFF
--- a/python/mspasspy/graphics.py
+++ b/python/mspasspy/graphics.py
@@ -11,26 +11,26 @@ from mspasspy.ccore import SeismogramEnsemble
 def wtva_raw(section, t0, dt, ranges=None, scale=1., color='k',
                    normalize=False):
     """
-    Plot a numpy 2D array (matrix) in a wiggle trace, variable area 
+    Plot a numpy 2D array (matrix) in a wiggle trace, variable area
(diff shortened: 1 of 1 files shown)


pavlis:
I would like to merge this branch with master if you concur.  I would now but thought i should make sure Travis builds with no errors first - in progress.

PR #91, opened 2020-12-08
Jinxin

JiaoMaWHU:


JiaoMaWHU:
This commit 
1. adds channel support for both types
2. support new schema
3. change collection names
4. add magnitude 
5. change history to history_object

PR #94, opened 2020-12-21
Cleanup core c api

pavlis:
This is the revision that moves elog from ProcessingHistory to be a member attribute of atomic TimeSeries and Seismogram data objects.  Also makes some api changes for python.  Most notably Core objects now have a leading underscore:  _CoreTimeSeries and _CoreSeismogram.   This should serve as a clear hint to python programmers these should be used only with caution.

There are a few additional features cleaned up in the process. 

1. Constructors are regularized for Seismogram and TimeSeries to eliminate any need for intermediate Core objects in a python code.
2. Learned a trick to use the "virtual public" inheritance clause to clean up copy constructors for Seismogram and TimeSeries.  There may be a tiny increment in performance.

### DIFF
--- a/cxx/include/mspass/seismic/CoreSeismogram.h
+++ b/cxx/include/mspass/seismic/CoreSeismogram.h
@@ -37,7 +37,8 @@ done, for example, with std::basic_string made equivalent to std::string.
  through inheritance of a Metadata object.
 \author Gary L. Pavlis
 **/
-class CoreSeismogram : public mspass::seismic::BasicTimeSeries , public mspass::utility::Metadata
+class CoreSeismogram : virtual public mspass::seismic::BasicTimeSeries,
+              virtual public mspass::utility::Metadata
 {
 public:
  /*!
@@ -138,7 +139,8 @@ Initializes data and sets aside memory for
 
  \exception  Will throw a MsPASSError if required metadata are missing.
  */
-        CoreSeismogram(const mspass::utility::Metadata& md,const bool load_data=true);
+ CoreSeismogram(const mspass::utility::Metadata& md,const bool load_data=true);
+
 /*!
  Standard copy constructor.
 **/
--- a/cxx/include/mspass/seismic/CoreSeismogram.h
+++ b/cxx/include/mspass/seismic/CoreSeismogram.h
@@ -37,8 +37,8 @@ done, for example, with std::basic_string made equivalent to std::string.
  through inheritance of a Metadata object.
 \author Gary L. Pavlis
 **/
-class CoreSeismogram : virtual public mspass::seismic::BasicTimeSeries,
-              virtual public mspass::utility::Metadata
+class CoreSeismogram : public mspass::seismic::BasicTimeSeries,
+              public mspass::utility::Metadata
 {
 public:
  /*!
@@ -146,6 +146,7 @@ Initializes data and sets aside memory for
 **/
 
 	CoreSeismogram(const CoreSeismogram&);
+
 	/* These overload virtual methods in BasicTimeSeries. */
 	/*! \brief Set the sample interval.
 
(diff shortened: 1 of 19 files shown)


pavlis:
This version of this branch is most definitely broken.  The change to virtual public inheritance broke the copy constructors for TimeSeries and Seismogram.   I'm working on a fix.

pavlis:
Cleaned up some of the problems in this branch, but more may be coming after some additional tests.   There was a mysterious problem I only marginally understand with inheritance where the copy constructor was not running the CoreSeismogram or CoreTimeSeries copy constructor in this kind of construct:
```
Seismogram::Seismogram(const Seismogram& d) : CoreSeismogram(d),ProcessingHistory(d),elog(d.elog)
```
The problem seems to have been created by this declaration for CoreSeismogram:
```
class CoreSeismogram : virtual public mspass::seismic::BasicTimeSeries,
              virtual public mspass::utility::Metadata
```
For some weird reason this caused the Seismogram copy constructor above to call the default (i.e no argument) constructor for BasicTimeSeries and Metadata.  That is not at all what we want.  As best I can glean from google searches this is likely due to some flaw in the inheritance chain possibly from lower level objects lacking any virtual methods.  Without them the vtable has no link back to the desired constructor.  I think what is in this branch is working, but we may face this again.

wangyinz:
I guess there is still quite some clean-up needed on the Python side. Let me work on that tonight.

wangyinz:
Turns out there are more problems than I previously thought, and they are actually in the C++ code. I have fixed most of the python calls that are affected by the API change, however, there are still mysterious problems from various tests. I've looked into one of them, which can be reproduced with:
```
import pickle
from mspasspy.ccore.seismic import TimeSeries, Seismogram
from mspasspy.ccore.utility import ErrorSeverity

from datetime import datetime

test_ts = Seismogram()
test_ts['test'] = ' '  # empty key
test_ts['extra1'] = 'extra1'
test_ts['extra2'] = 'extra2'  # exclude
test_ts.elog.log_error("alg", str("message"), ErrorSeverity.Informational)
test_ts.elog.log_error("alg", str("message"), ErrorSeverity.Informational)
test_ts.erase('starttime')
test_ts['t0'] = datetime.utcnow().timestamp()

ts = pickle.loads(pickle.dumps(test_ts))

print(test_ts.elog.size())
print(ts.elog.size())
```
The problem here is that the serialized elog will always be empty as can be seen from the last two print commands. It took me a while to realized that the problem is actually from the `elog` member of the `ProcessingHistory` class. Turns out it is not removed, so somehow that `elog` was used instead of the one under TimeSeries or Seismogram. 

When trying to remove the elog member from ProcessingHistory, I found that the methods of ProcessingHistory were using the `elog` extensively to record various error messages within the call. The question here is how we should handle those error messages once elog being removed? I am not sure we can just throw them as exceptions here...

pavlis:
Hmmm - this is getting uglier all the time.  I had totally forgotten ProcessingHistory actually used the elog mechanism.  It might be easier to deal with except for the fact it uses the mechanism a lot and to log nonfatal errors that should not generate an exception.    

I recommend we abandon this branch and return to the previous design where elog was a member of ProcessingHistory.   It is largely an OOP purest view that it shouldn't belong there.   i.e. ErrorLogger is an example of something where the data "has a" ErrorLogger than "is a" ErrorLogger.   Remind me why we decided to make that change anyway?  My recollection is it was not an essential problem the way it has become in this context.   Not sure if I can jump from this page and not lose what I just composed so there may be a followup below here after I read our earlier github discussion that lead us down this dark alley.

pavlis:
When I reread our discussion of this topic I see no reason not to abandon this branch.   If you concur I will fork again from origin and this time only fix the problems with the clear ambiguity.   After learning some things about the use of the "virtual public" trick I think I will also clean up the copy constructors in the previous implementation in a way that should theoretically make them run slightly faster.   I won't geek out on the details in the dark alley where I think this set of changes are destined to die.

wangyinz:
sounds good to me

PR #95, opened 2020-12-24
Regularize ccore

pavlis:
This branch 
1.  Fixes inconsistency of clean methods for ProcessingHistory and Metadata changing the names to clear_history and erase respectively.
2. Regularizes the constructors for Seismogram and TimeSeries objects to have useful versions parallel to those in Core equivalents. 
3. In the process the copy constructor for TimeSeries and Seismogram were made slightly more efficient by declaring inheritance in Core objects to be virtual public.   That allowed calling Core copy constructors directly in Seismogram and TimeSeries copy constructors.  

Before merging this consider if all the constructors defined in the pybind11 definitions are necessary.   I commented out one already, but you may wish to consider others.

### DIFF
--- a/cxx/include/mspass/seismic/CoreSeismogram.h
+++ b/cxx/include/mspass/seismic/CoreSeismogram.h
@@ -252,7 +252,7 @@ to an exception if the the time requested is outside the data range.
 */
         std::vector<double> operator[](const double time)const;
 /*! Standard destructor. */
-	~CoreSeismogram(){};
+	virtual ~CoreSeismogram(){};
 /*!
  Apply inverse transformation matrix to return data to cardinal direction components.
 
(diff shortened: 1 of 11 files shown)


pavlis:
I see this is creating some errors in python code because of the changes to clear.   I need to be educated on how to run the same tests before I push a branch like this.  I could have fixed them earlier had I known how to do that.   For now, is it easier to repair these here or my local branch I pushed here?

JiaoMaWHU:
Hey Gary, to run the tests, first go to the mspass folder and run `export MSPASS_HOME=$(pwd)`. Then you can run all the test cases by running `pytest python/tests/`. You can also run a specific case, for example, the db, by 'pytest python/tests/test_db'.

wangyinz:
this is included in #97

PR #99, opened 2020-12-29
Regularize operators

pavlis:
This branch includes cleaning up ccore operators for data objects. Seismogram and TimeSeries with this branch have operator +=, -=, +, and -.   Also include  python test program to validate the operators.  

Finally, cleaned up a few errors I could find from api changes for Metadata::clear being mapped in pybind11 to erase.


pavlis:
Ian, can you try to sort out the errors that remain in this branch?  I think they are coming from the database test and I don't know how to run those locally as I think they require MongoDB running on the docker container.   If you can educate me on that I might be able to fix the problems in this branch so we can merge it clean.

Issue #78, opened 2020-11-02
Database api design


pavlis:
In our last call we discussed the fact that something we need to finalize soon is our database api for python.   Ian, I know you said you would start this but in working on raw data handling this week I was conscious of this imminent need and thought I'd pass along some insights I have gotten as a newbie to MongoDB and implications for mspass.   

- The syntax of MongoDB's query language is weird and will be a barrier to many people using our package if we don't give them a crutch to walk with when they are getting started.  Maybe a better analogy is one of those wheeled baby carts where a baby can get around before they learn to walk.  I think good examples of what we need to build on are the waveform readers and writers.   They require only the user know  how to create a MongoClient and construct a Database object.  That kind of abstraction of common database needs should be our primary design goal.
- We can't cover everything.  We need to come up with a set of initial common queries we can implement to simplify common needs.   
- Once someone sees how the db works, they will be more motivated to dig into the weeds of MongoDB's API.   We probably won't need a lot of help on that, but we might want to design a set of functions/methods that have a "query" argument that would be the typical py::dict used by pymongo. 
- Here is a list of common functionality I think might be necessary/useful:
 1.  Return a cursor to wf with data sorted by some key (might default to source_id)
 2.  Return the result of grouping wf in a form that could be used to construct an RDD of ensembles.  The atomic object level for some operations are ensembles so we either need a way to tell an algorithm how to group data or pass ensembles to provide a clean map operator.  
3.  We will need a preprocessing dbverify program that will apply generic tests of db integrity.  This functionality will be essential to validate large inputs.  What and how to structure this is not simple.   My perspective is it should be extensible with a core set of tests we build and a mechanism to allow a user to build and apply custom tests.   
4.  We need some ways to print and/or plot some common metadata and/or summaries of metadata properties.   Examples are:  station coordinates or a map of same (may be able to use obspy's plotting for this), source coordinates (same point as stations with obspy), summaries of waveform properties (list of unique stations, list of unique sources, waveform time coverage plots, and likely others.  This isn't an immediate need, but we will need to develop some of these along the way to validate test data.  I've already run into many cases where it is hard to figure out what is going wrong because the data volume can't just be listed.  
[...]

PR #102, opened 2021-01-05
Fix database

pavlis:
I pushed this branch to help resolve some parallel development by Jinxin and myself that created conflicts in the database.py file and a some smaller side effects created by the need to merge branches that had diverged for a few weeks.   I believe this branch resolves the conflicts in database.py.   I made changes that may break some test code and I haven't cracked your system yet to run tests on my local machine.  Hence, I'm going to push this and hopefully we can resolve any issue here on github.

In addition to resolving the branch conflicts this branch has the following  additional features:

1. It fixes a significant bug that was previously undetected in BasicTimeSeries handling of switching between relative and UTC times.  It was undetected until I looked at output from the database writer in MongoDB and realized there was a bug.
2. Part of the reason that bug was uncovered was because I also discovered a major hole in the Database.save_data method.   It was not saving the time shift data when data were time shifted to relative time, which in the test case was based on an arrival time.  I added a small python function to the database.py file to handle this called *_sync_time_metadata*.    This implementation required adding two new attributes to the wf schema that are in a revision to the mspass.yaml on this branch.  They are: boolean *CanBeConvertToUTC" and double *starttime_shift*.  It also sets the string attribute *time_standard* but I think that had been defined previously.   Note we may want to move this function to Database and use the schema to set these keys instead of hard coding them like they are now.   For now it is a standalone procedure instead of a method in the class.
3. I added a related function *_sync_ensemble_metadata* that is needed to avoid losing attributes stored in ensemble's global (ensemble) Metadata area.   The current implementation does something a little evil to reduce changes to the Database class.  It actually copies the ensemble's Metadata to every ensemble member before calling save_data.   Because of our implementation that means the data passed to Database will be altered when the ensemble writer is called.   This can be fixed but I elected to leave this as a side effect for now.   We may want to resolve this before merging this branch.
4. I implemented a little function *_repair_type_mismatch* to use the schema to attempt to fix any readily solvable type conversions like int to float.   You should look at this critically.   There are a couple of issues.  First and most important is that this routine can still throw an exception.    Lacking is a handler for the possibility that this can throw an exception.  I intentionally omitted that because it impacts the overall logic of the save_data method.  The reason is that I think what the handler should do is post a warning message to elog and blunder on.   What is less clear is if that message should then be immediately posted to the database or just pushed to the data and left to be handled downstream.   The other issue is simpler:  does this need logic for additional supported types.   Until I added a couple things the previous version, for example, didn't have a boolean attribute listed.   Are there others?
5. There are some minor changes to the preprocessing.seed.ensembles.py file to mesh with the Database changes.  I also added a small function to remove debris in data read from miniseed that obspy posts to their stats container and which gets mapped into our Metadata.  Once converted from seed these are meaningless debris that simply waste space.  They are *_format* which is always set to "mseed" that obspy converts to a dict and that end up mapping to a subdocument if written to mongodb.   The contents are meaningless junk once the miniseed file is converted of things like block size, blockette style, etc.).   I am 100% sure both should be deleted, but right now it requires a separate call.  We might eventually be dogmatic about clearing these in a miniseed reader.


pavlis:
Sorry, I need help on this.   I cannot figure out why it is failing where it is and remain unable to figure out how to run the tests on my local machine.   I don't know how to solve this problem without that capability and the capability of running test_db through some kind of debugger.  I am not understanding how these tests are run.   Can you attempt to fix the current issues that are causing test_db to error?  I think this is an example of something I said yesterday that you should use me to provide experience and background and be patient when I can't adapt quickly to an unfamiliar tool like this one.

JiaoMaWHU:
I'll take a look after 6 pm.

wangyinz:
Fixed two problems and there is still one remaining. This issue seems relate to how relative time and absolute time conversion is done, and it somehow conflicted with way the test code is using the time for. The test code seems is not really using the time in the right way, so it could be the test code's fault here. Anyway, I need to debug further to understand this behavior. Or, @JiaoMaWHU can take care of it assuming he got some time to look at it.

There is one thing in the code that I am confused.  i.e., https://github.com/wangyinz/mspass/blob/c48d8c13a85644d66208373edf39c8999602bbaa/python/mspasspy/db/database.py#L246-L253
First of all, this is the simpler way of handling type conversion, because the return value of type is the Class itself, so it can be called directly to initiate a new instance of that type. What I am not sure is that the previous `_repair_type_mismatch` method will not throw an exception, but return a key value pair of `'FAILURE' : error_message` instead. One of the original problem related to these lines is that the key got changed to `'FAILURE'` so was not correctly called. However, I am not sure what suppose to be the correct mechanism here. Are we suppose to save this entry of `'FAILURE' : error_message` into the database? That doesn't seem right and especially when multiple keys had the same conversion error, they will be overwritten because of the key is shared. Are we suppose to add this into the elog, but then shall we just ignore this problematic key? What if the key is the important ones such as the `_id`? It seems to me that we'd better just throw the error here and stop. What do you think?

JiaoMaWHU:
I'm looking into it.

JiaoMaWHU:
Regarding the time test case, I think the problem is the clear_alias method. @pavlis Did we change this method in this pull request? Previously, let's say for a timeseries object TS, with TS['t0'] = 2 and TS['starttime'] = 1, if we do TS.clear_alias, 'starttime' attribute will be overwritten to 2. But it seems it's not overwriting right now. If the change is expected, please let me know and I can change the test case itself.

wangyinz:
hmmm... that behavior you described does not seem right. If starttime is already defined, we shouldn't overwrite it with t0. Well, it is a user error to have inconsistent starttime and t0 in the first place. Also, the other problem here is that there is a difference between ts['starttime'] and ts.t0. In Python, modifying ts['starttime'] will not affect ts.t0, but modifying ts.t0 will also change ts['starttime']. It seems to me that this also contributes to some of the inconsistency.

wangyinz:
What I wrote above was not quite accurate, and this is how `set_t0` works:
https://github.com/wangyinz/mspass/blob/235f484d8c9a267235bdbe8bd4eeee13700df029/cxx/src/lib/seismic/CoreTimeSeries.cc#L201-L218
It actually also changes the `t0` and `time` in Metadata if they exist.

JiaoMaWHU:
ok, I need to dig more into this.

JiaoMaWHU:
@wangyinz Thanks for pointing it out. It turns out using the ".t0" method can simply fix this problem. But the codecov is failing on the patch difference.

wangyinz:
Yeah, actually this helps me identify a problem deep within the `change_key` method of `Metadata`, so I just got that fixed and let the `clear_aliases` in the schema to safeguard the unexpected change of unique name instead. This plus your fix should work now.

wangyinz:
Actually, it makes me wonder how the test worked previously... you must be lucky I guess

JiaoMaWHU:
I guess Gary must have fixed something in this branch. Otherwise, it doesn't make sense and that's what I'm asking in the first place.

python/mspasspy/db/database.py
```diff
@@ -1089,3 +1101,125 @@ def load_event(self, source_id):
...
+    involves posting some special attributes seen below to handle this issue.
+    Since Metadata is volatile we need to be sure these are consistent or
+    timing can be destroyed on data.
+    """
+    # this adds a small overhead but it guarantees Metadata and internal t0
+    # values are consistent.  Shouldn't happen unless the user messes with them
+    # incorrectly, but this safety is prudent to reduce the odds of mysterious
+    # timing errors in data
+    t0=mspass_object.t0
+    mspass_object.set_t0(t0)
```
wangyinz, line 1185: Note here, this is actually what revealed the bug with `t0`

wangyinz: @pavlis Actually, I am confused. Here we save the `utc_convertible` and `startime_shift` into the database, but how are we gonna use them when reading data? These will be retrieved to the Metadata, but the `ator` and `rtoa` methods won't use the Metadata field. Maybe we should make the `t0shift_is_valid` and `t0shift` of `BasicTimeSeries` visible and link them to the corresponding Metadata field?

JiaoMaWHU:
I've finished the documentation. But I don't think it's a good idea to append the commits in this request. I'll wait for this request to be merge, solve conflicts, and then pull a new request for the docs. @pavlis @wangyinz

wangyinz:
Oh, what I meant was to open a new pull request for the documentations. That update is orthogonal to this branch anyway.

JiaoMaWHU:
I don't think it's a good idea to have two pull requests for a module at the same time. We better close this at first. After that, I will solve the upcoming conflicts and pull a new one.

wangyinz:
well, there are still problems in this branch... Anyway, I guess it is fine to merge this first, but I need to at least fix the minor ones first.

python/mspasspy/db/database.py
wangyinz: We should continue this topic in #78

PR #104, opened 2021-01-08
Fix error handling in database

wangyinz:


pavlis (commented):
While this is fresh in your mind I would advise adding a few comments to the code for a 
few rather tricky things done here.  e.g. I hadn't seen the trick in line 248:  
```
insert_dict[k] = update_metadata_def.type(k)(copied_metadata[k])
```
With a line like this:
```
# This is like a cast operator in C to convert the actual value in a dict to a required type
```
Consider others less obvious to me.

PR #105, opened 2021-01-13
fix history object id type && testing

JiaoMaWHU:
This commit fixes the logic that the inserted id of the history object should be the UUID set in the processing history class.

Since deep-copy in testing will copy the same UUID and result in mongo duplicate-key-error, modifications are made, i.e. using logging_helper to create a new node.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -520,30 +520,33 @@ class Database(pymongo.database.Database):
         else:
             d.put_string('time_standard', 'UTC')
 
-    def _save_history(self, mspass_object, history_object_id=None, collection=None):
+    def _save_history(self, mspass_object, prev_history_object_id=None, collection=None):
         """
         Save the processing history of a mspasspy object.
 
         :param mspass_object: the target object.
         :type mspass_object: either :class:`mspasspy.ccore.seismic.TimeSeries` or :class:`mspasspy.ccore.seismic.Seismogram`
-        :param history_object_id: the previous history object id (if it has).
-        :type history_object_id: :class:`bson.objectid.ObjectId`
+        :param prev_history_object_id: the previous history object id (if it has).
+        :type prev_history_object_id: :class:`bson.objectid.ObjectId`
         :param collection: the collection that you want to store the history object. If not specified, use the defined
         collection in the schema.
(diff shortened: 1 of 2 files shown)


python/mspasspy/db/database.py
```diff
@@ -520,30 +520,33 @@ def _sync_metadata_before_update(self, d):
...
             # overwrite history
-            filter_ = {'_id': history_object_id}
-            history_col.find_one_and_replace(filter_, {'nodesdata': history_binary})
-            return history_object_id
+            history_col.delete_one({'_id': prev_history_object_id})
+            history_col.insert_one({'_id': current_uuid, 'nodesdata': history_binary})
         else:
-            return history_col.insert_one({'nodesdata': history_binary}).inserted_id
+            # new insertion
+            history_col.insert_one({'_id': current_uuid, 'nodesdata': history_binary})
```
wangyinz, line 548: Is the duplicate key error thrown by this line? I think that's potentially a bug because it is possible for a user to deep copy an object. Maybe we should add an exception handling here such that when the UUID is duplicated, we generate a new uuid. I guess we need to make sure the change is also applied to the `mspass_object` outside of this database operation. Do you think it is gonna work?

JiaoMaWHU: UUID is designed to be unique on the application-level right?

wangyinz: Yes. It is unique, so the problem should only happen when a deep copy is make and the `update_metadata` is called as a new insertion.

JiaoMaWHU: Sorry, I'm a little confused here. First, why new insertion is related here? Do you mean that when it's a new insertion, we won't delete the old object, so the duplicate error can happen? Then, I do think it's the user's problem to cause a duplicate key error since our processing history is designed like this. Deepcopy should never be recommended in users' case, I'm using it just for the convenience of testing. From another perspective, is it really a good practice that we change the _id implicitly when we find it may produce a duplicate error?

wangyinz: Good point! Let's just wrap the error with MsPASSError as you suggested. This should be fine here because no database operation is performed yet at this point.

PR #106, opened 2021-01-13
Add bundle algorithms

pavlis:
This branch includes a set of new algorithms for what I'm calling "bundling".  That is to combine a set of TimeSeries objects into a set of Seismogram objects.   The algorithms are mostly C++ with very thin python wrappers.  There is also a  new C++ test program with proper asserts to validate the algorithm is working.  The tests there are pretty complete, but may not exercise all the features.  In particular, they weren't updated when I added new code to properly handle failed bundling.  The code here creates dead Seismogram objects whose elog and Metadata can be used postmortem to identify the body.

I attempted to merge this with the origin branch with a rebase operation as you (Ian) had advised but I got into a serious can of worms with conflicting commits done in the last week mainly on test_db.py and database.py.   After suffering through about 20 demands by git rebase to resolve conflicts manually (required edits, git add, git commit, and git rebase --skip for each conflict) I decided I was doing something dangerous and aborted the rebase.   Thus, this branch diverges from master a lot and it is going to take some work to resolve the conflicts.  I think you, Ian, are better prepared to resolve those because all the complaints rebase was throwing at me were about test_db.py and database.py.  You may have a better way to resolve this than the tedious approach I was doing with direct interaction with git with the unix shell.


cxx/python/algorithms/basic_py.cc
```diff
@@ -16,6 +22,10 @@ PYBIND11_MODULE(basic, m) {
   m.attr("__name__") = "mspasspy.ccore.algorithms.basic";
   m.doc() = "A submodule for algorithms namespace of ccore with common algorithms";
 
+  py::bind_vector<std::vector<double>>(m, "DoubleVector");
+  py::bind_vector<std::vector<TimeSeries>>(m, "TimeSeriesVector");
+  py::bind_vector<std::vector<Seismogram>>(m, "SeismogramVector");
+
```
wangyinz, line 28: These bindings of vector seem unnecessary. Let me try removing them.

cxx/src/lib/utility/ProcessingHistory.cc
```diff
@@ -750,6 +750,7 @@ ProcessingHistory& ProcessingHistory::operator=(const ProcessingHistory& parent)
     mytype=parent.mytype;
     algorithm=parent.algorithm;
     algid=parent.algid;
+    elog=parent.elog;
```
wangyinz, line 753: This should be a bug. I guess our tests are outdated so it is never detected. I need to revise corresponding test.

wangyinz: This operator is never explicitly called on the python side. I added a test for the `elog` member instead.

cxx/python/algorithms/basic_py.cc
wangyinz: Removed fine as expected.

cxx/python/algorithms/basic_py.cc
```diff
@@ -108,6 +108,47 @@ PYBIND11_MODULE(basic, m) {
...
+  m.def("_bundle_seed_data",&bundle_seed_data,
+    "Create SeismogramEnsemble from sorted TimeSeriesEnsemble",
+    py::return_value_policy::copy,
+    py::arg("d") )
+  ;
+
+  m.def("_BundleGroup",&BundleGroup,
+    "Bundle a seed grouping of TimeSeries into one or more Seismogram objects",
+    py::return_value_policy::copy,
```
wangyinz, line 120: Why do we need to specify the return value policy here? Seems could be inefficient. @pavlis

cxx/python/algorithms/basic_py.cc
wangyinz: Previous comment was not true. It seems the bundle code is never called in any tests, so it is not clear whether they worked or not.

wangyinz: Further testing shows that with or without these "bind_vector" lines are equally problematic. With the lines here, we are redefining these "Vector" types under this `mspasspy.ccore.algorithms.basic` module. But because they are also defined in `mspasspy.ccore.seismic`, it turns out they will be treated as different classes. Basically, `mspasspy.ccore.algorithms.basic.TimeSeriesVector == mspasspy.ccore.seismic.TimeSeriesVector ` will be `false`.

However, if we remove these lines, the input and output type of `BundleGroup` will become `List[mspasspy.ccore.seismic.TimeSeries]` and `List[mspasspy.ccore.seismic.Seismogram]`, which is also different from the "Vector" types...

wangyinz: This is fixed by https://github.com/wangyinz/mspass/pull/106/commits/0f4f732c75de07c788357785f3ea538325ecef84

cxx/python/algorithms/basic_py.cc
pavlis: I was being conservative because I wasn't sure how this would behave.   BundleGroup does return a fairly large but for planned use normally not huge object (one or a few Seismogram objects).  My concern was memory management.   I was concerned that if we didn't copy the data it would vanish when BundleGroup when out of scope.   However, as a return I may be wrong about that.  This is deep in the subtleties of how python and C++ interact in this context.  In unknowns is why I intentionally made this copy.   I suspect my concern would not be difficult to test.   I could probably do so easily with the little python test program I hadn't put here because the data it was using is too huge for github.

pavlis:
I tested the revisions and they passed my test program.  However, I realized when finalizing the documentation that I should rethink the usage of BundleGroup.   That is, the original algorithm was supposed to be able to handle groups of multiple stations and return a vector of Seismograms.  I realized I hadn't tested that and found that the version at this moment does not do that anyway.   

I am going to change BundleGroup to return only a single Seismogram.  Further, to make it more generic rather than receive an Ensemble it will only receive a std::vector<TimeSeries>.   That will make it more useful as a building block because it won't care how the data are sorted - it just wants a range of index positions.   The only thing it adds is small safeties on time alignment and a consistent error handled (returning a dead object with an elog entry if it fails).  

My test program, by the way, is even less tranportable than I remembered.  It depends on an assembled data set indexed in MongoDB with the procedure in the raw_data_tutorial.   Nonetheless, here it is with the last test I did that will be no longer relevant removed (testing with a bundle of 6 instead of 3).
```
from mspasspy.db.client import Client
from mspasspy.preprocessing.seed.ensembles import (load_one_ensemble,
                                                   load_channel_data)
#from mspasspy.algorithms.signals import (bundle_seed_data,BundleGroup)
from bundle import(bundle_seed_data,BundleGroup)
from mspasspy.ccore.algorithms.basic import seed_ensemble_sort
def print_dead_logs(ens):
    count=0
    for d in ens.member:
        if d.elog.size() == 0:
            if d.dead():
                print("ERROR:  Ensemble member number ",count," is marked dead but has an empty elog")
        else:
            print("Member number ",count," has a elog entry")
            if d.live:
                print("DIFFERENT PROBLEM:  marked live with errors")
            for err in d.elog.get_error_log():
                print(err.algorithm,err.badness,err.message)
        count+=1
client=Client()
db=Database(client,'raw_data_tutorial')

dbwf=db.wf_miniseed
cursor=dbwf.find({}).limit(5)
n=0
for doc in cursor:
    print("Working on ensemble number ",n+1)
    ens=load_one_ensemble(doc,apply_calib=True,
                  ensemble_mdkeys=['source_id','starttime','endtime'])
    print("Number of TimeSeries in this ensemble=",len(ens.member))
    load_channel_data(db,ens)
    ens3c=bundle_seed_data(ens)
    print_dead_logs(ens3c)
    print("Number of Seismograms after bundle_seed_data=",len(ens3c.member))
    # quick test of sort and BundleGroup
    print("testing sort algorithm in isolation")
    seed_ensemble_sort(ens)
    print("Testing BundleGroup")
    # This explicitly uses first 3 channels as a group
    s=BundleGroup(ens,0,2)
    print("Type returned by BundleGroup=",type(s))
    print("Length of vector returned by BundleGroup (should be 1): ",len(s))
    n+=1
```
BTW:  I'm posting a related new discussion topic that emerged from the need for the small function in this test program

wangyinz:
Are you going to implement the changes? I guess after the vectors removed, the trick I add in https://github.com/wangyinz/mspass/commit/0f4f732c75de07c788357785f3ea538325ecef84 is no longer needed. We should probably remove that too, because making the bindings global may cause unexpected collision (although it is very unlikely).

pavlis:
Working on this.   The trick you used will simplify the binding code so I may try that.  Thanks for pointing that out - a detail there were no way I'd have seen otherwise

pavlis:
Just pushed some fairly significant changes to the BundleGroup function to make it do more what I think it should.  Now returns just a single seismogram and has a lot of tests for consistency of seed net,sta,chan,loc codes before it attempts to bundle a group.

One point is the name BundleGroup is probably not what we should call this.  The reason is it is not at all generic but depends completely on seed naming keys (net,sta,loc,chan).   Maybe I should rename it "BundleSEEDGroup"?  Let me know if you think that wise and I'll push that change.  

Related before we finalize this is where to put the python wrapper code?   I have a bundle.py file right now that just needs to be inserted into some other module.  I presume it should go in algorithms/signal although that isn't completely descriptive of what it does.   Maybe it belongs in preprocessing/seed since all the functions are seed centric BUT they aren't necessarily preprocessing functions either.   Not clear to me so let me know and I'll push that (hopefully) final change for this branch.

wangyinz:
Naming it `_BundleSEEDGroup` sounds good to me. This is gonna be a protected function, so I guess the naming doesn't matter that much.

For the python wrapper, the first place that I can think of is to put the bundle as a submodule of `mspasspy.util`. Or, if you don't want to make it a standalone module, maybe we could put it under `mspasspy.util.converter`. What do you think?

pavlis:
Ok, after one last test I'll run after the setup.py job finishes I'll check in what I hope is the final revision of this.  Two things are subject to your editing and I leave them at your discretion:
1.  I made the name BundleSEEDGroup - I'm assuming you will accept that one without any changes.
2. I'm going to put a file bundle.py in mspasspy/algorithms.   I decide that made more sense because the C++ code is in ccore/algorithms and it is descriptive of what it is.  That is, the python file is just wrappers for code that does match our model of a processing algorithm.

Worth pointing out, I think, that BundleSEEDGroup has two properties we might want to extend/exploit:
1.  It is a good type example of how to write a C++ code that does everything and requires only a minimal wrapper.   i.e. this example handles ProcessingHistory internally and does the proscribed approach to handling errors (kill followed by writing a message to elog).   
2. BundleSEEDGroup satisfies the reduce model, and I don't know how to map that into this framework.  The idea of this function is it should be the workhorse for reading data as TimeSeries (limited by data created from seed had having all the net, sta, loc, chan codes set) sorted properly into an RDD.   The C++ function just uses a range of  vector positions to define a bundle so a reduce wrapper should, I think, be able to take an RDD of sorted TimeSeries and produce an RDD of Seismogram objects.   I have only a rudimentary idea of how to do that so I'm passing the buck to you.

pavlis:
I would like to make  a strong argument we put the bundle.py code where I placed it in the last push (mspasspy/algorithms).   I'm working on documentation at the moment and the writing reminds me there will be a need for other *bundle* algorithms that would belong in that module.  e.g. I have collected a number of multichannel data sets that have 3C sensors attached and there the "bundle" algorithm would require a channel number to define the component number and direction.  That would be a very different, and much simpler algorithm than the SEED version in this branch.

wangyinz:
Looks good to me. 

Actually, in principle, any thing under the `algorithm` should accept Spark or Dask's parallel data format just like the `algorithm.signal` module, where we use the decorators to handle error logs and history. We still need to apply them to the other modules of `algorithm`. However, that reveals a deign problem of our current layout. First of all, the RFdeconProcessor is not defined as function, so we cannot apply the decorator directly. We will need to turn that into a function like we discussed before. Then, right now we have a dangling [reduce](https://github.com/wangyinz/mspass/blob/master/python/mspasspy/reduce.py) module, which should be under a submodule of algorithm. Finally, we have this new `BundleSEEDGroup` which does not directly take a mspass object but a list of mspass object as its input. I don't think our current decorator handles that. Also, to make that work properly, before calling the `BundleSEEDGroup` in a map function, users also need do a grouping to turn the distributed data of type TimeSeries into that of a list with three entries. This is something that should go to the document since there is no way for us to enforce that in our library. 

Anyway, I guess we should just merge this branch for now, and plan to implement the features later.

PR #119, opened 2021-01-20
Fix bundle

pavlis:
As the name of the branch implies this branch has fixes to bundle code to properly handle inputs marked dead.   Previous version did not always handle such situations properly.


pavlis:
Hmmm - not sure why all those commits are listed.  I did pull origin this morning before making the changes included in this branch.  Only the last two should be after I forked this branch.   I don't know why we have the relics in the chain

wangyinz:
Well, this branch should be abandoned, and I have deleted it already. It contains diverged history from the master that cannot be resolved automatically. Can you pull the latest master to you local repo, fork it to a new branch, and cherry-pick to apply the two commits to that new branch?

wangyinz:
hmm... so I was wrong. I assumed this was the add bundle branch. Weird...

pavlis:
Let me dig a bit here.  It would have been very easy to neglect to pull master.   I thought I had done that, but from what you suggest I think the most reasonable hypothesis is that I messed up.

This is exceptionally easy to fix because all the changes are limited to one file - bundle.cc

wangyinz:
Took a closer look at the commit history. I believe somehow you local master is diverged from the origin master (note the multiple merge commit in this pull request). I think the best way to resolve it is to do a hard reset on your local master so that it becomes in sync with the origin. Then, you should never see a merge commit when pulling from the origin.

pavlis:
Actually, look at the commit history.  You see the last two commits change only bundle.cc.   2 back from the tip logs my pull of origin master.  So, this may be a residual mess from the tangled branches.

Shall I just send you the updated bundle.cc file?   Not sure how to proceed here now

wangyinz:
yeah, let me deal with this pull request, but you still want to reset you local master to avoid such issue in the future.

wangyinz:
apply the changes in a new pull request #120

pavlis:
How do I do a "hard reset"?

wangyinz:
First make sure you are on the master branch, then you should can run:
```
git fetch --all
git reset --hard origin/master
```
Please be vary cautious because the reset will erase anything on your local working tree in the master branch.

PR #120, opened 2021-01-20
Fix bundle2

wangyinz:
This branch resolved the divergence issue in #119.

PR #121, opened 2021-01-21
Fix improper handling of input TimeSeries marked dead

pavlis:
This is the revision of bundle with what should be a clean commit history prior to these changes.   The "v2" on the branch name is to show it has a relationship to compare_fix_bundle.  Probably unimportant.  Only thing lost is an intermediate commit with some scaffolding I used for debugging saved.


wangyinz:
This is duplicated with #120

PR #125, opened 2021-01-24
Fix issues of database behavior

wangyinz:
This is to address various issues discussed in https://github.com/wangyinz/mspass/discussions/122

### DIFF
--- a/cxx/include/mspass/seismic/Ensemble.h
+++ b/cxx/include/mspass/seismic/Ensemble.h
@@ -95,6 +95,23 @@ public:
           (*mdmember)+=dynamic_cast<mspass::utility::Metadata&>(*this);
       }
   };
+  /*! \brief copy ensemble metadata to all members except for the ones excluded.
+    */
+  void sync_metadata(std::vector<std::string> exclude)
+  {
+    mspass::utility::Metadata sync_md(*this);
+    for (size_t i = 0; i < exclude.size(); i++)
+    {
+      if(sync_md.is_defined(exclude[i])) {
+          sync_md.erase(exclude[i]);
+      }
+    }
+    for (size_t i = 0; i < this->member.size(); ++i)
+    {
+      mspass::utility::Metadata *mdmember = &(this->member[i]);
+      (*mdmember) += sync_md;
+    }
+  };
 };
 /*! Useful alias for Ensemble<TimeSeries> */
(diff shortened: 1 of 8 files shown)


wangyinz:
I revisited some of my changes and find a couple of potential issues:

What should be the right behavior of `update_metadata` when called with a dead object? What if the operation is an insert v.s. an update? Right now, `update_metadata` does not check for dead object, but the `_save_elog` does, so the method will create an inconsistent database entry, in which an valid `wf` entry as well as a `gravestone` entry under the elog collection coexist. Maybe we should let `update_metadata` silently skip dead objects (by appending to its elog)?

We should probably always use the `exclude_keys` argument to skip keys. Right now `exclude` is used sometimes.

pavlis:
Well, let's maybe focus on how this should be handled.   We have all combinations of dead or live (boolean) plus the modes:  (1) add new, (2) update metadata only, and (3) update all.   That leaves 6 cases I'll tag as D/L for dead or live and 1,2, 3.  

L1 - normal case, keep everything
L2 - update metadata as requested and save any elog entries present
L3 - update metadata and save elog entries if present delete old metadata and waveform data but do not touch elog
(i.e. if L save everything requested
D1 - save metadata, save elog, but do not save waveform data
D2 - update metadata and save any elog entries, delete waveform data
D3 - The only tricky case to handle.   Update metadata and save elog for sure.   Only issue is what to do with the waveform data.  Probably best to delete the waveform data.

Is that helpful or only confuse things?

wangyinz:
I think current implementation does exactly what you listed except that the `update_metadata` would not delete waveform in any case. I think that should be the more reasonable behavior anyway. Also, when data object is dead, its metadata will always be saved as the "gravestone" entry of the elog, so nothing in the wf collection will be changed. I guess I still need to correct a couple issue in the `update_metadata` method so it conforms with `save_data` and `_save_elog`.

wangyinz:
I decided to remove the `_sync_ensemble_metadata` method, and not adding the call to `Ensemble.sync_metadata` to `save_ensemble_data`. I think this is the correct behavior since users may or may not want to save the metadata of the ensemble with the members. For those that wants that, they should just call `sync_metadata` before saving.

wangyinz:
note that I added an overloaded `sync_metadata(std::vector<std::string> exclude)` method to the ensemble class as well as corresponding tests.

PR #127, opened 2021-01-25
Improve preprocessing

pavlis:
This branch has three changes.  Two I doubt you will have problems with but item 3 below will require a decision from on how to handle it.  First, here are the things changed:
1.  In finalizing the raw data tutorial, which is a prototype for our benchmark workflow the the usarray data, I fixed a couple of minor bugs in the ensemble.py file.  
2. As an aid to sorting out some additional site and channel metadata problems I developed two functions I think are useful additions to the database api.  These functions are called *site_report* and *channel_report*.  The produce a nicely formatted display of key metadata for a selected equal match to one or more of the keys net, sta, chan, loc.  I put them in the database.py module because I thought they belonged there.  I do not, however, think they belong as methods in the Database class.   The alternative is to create a new module in preprocessing/seed because these functions depend completely on seed name codes, which we should not treat as a requirement for site and channel.  (In fact, in the revisions to the documentation I emphasize that point and give a reason why it isn't at all generic.)   I can see reasons for both organizations.   The seed model might actually be better as I can see a set of seed specific functions we might want to develop that belong in preprocessing.    A bit depends on what additional functionality we think might be needed for manipulating site and channel during preprocessing.  
3. To get this to work at all I had to do a set of fixes to database.py to work around the mishandling of dead data and data with empty history chains.  Both issues aborted a call to save_data when either condition was present.  I think you may be doing a parallel fix so I suggest you look at what I did and make sure you fixed the same problems.  If not you will need to accept my change or add a (perhaps cleaner) solution to the same problem.

The only sure way to test this stuff is to run the raw data tutorial I'm about to also check in on the tutorial branch.  Warning that going through all the steps will consume a couple of hours by the time you read the material, absorb it, and run the steps (some take a while to complete due to the collection size or the data volume (the last step)).

### DIFF
--- a/python/mspasspy/preprocessing/seed/ensembles.py
+++ b/python/mspasspy/preprocessing/seed/ensembles.py
@@ -582,9 +582,8 @@ def load_site_data(db,ens):
                 d['site_elev']=siterec['elevation']
                 d['site_id']=siterec['site_id']
                 if n>1:
-                    message = ('Muliple (%d) matches found for net %s sta %s with reference time %f'
-                               % [n,net,sta,t0])
-                    d.elog.log_error('load_site_data',message,ErrorSeverity.Complain)
+                    message = "Multiple ({n}) matches found for net={net} and sta={sta} with reference time {t0}".format(n=n,net=net,sta=sta,t0=t0)
+                    d.elog.log_error('load_site_data',message,ErrorSeverity.Complaint)
         return ens
     except Exception as err:
         raise MsPASSError('Something threw an unexpected exception',
@@ -635,9 +634,8 @@ def load_channel_data(db,ens):
                 # that is what find_one returns.  We use the count to make
                 # the eror message cleaer
                     chanrec=dbchannel.find_one(query)
-                    message = ('Muliple (%d) matches found for net=%s sta=% chan=%s loc=%s with reference time %s'
-                               % [n,net,sta,chan,loc,t0])
-                    d.elog.log_error('load_site_data',message,ErrorSeverity.Complain)
+                    message = "Multiple ({n}) matches found for net={net} and sta={sta} with reference time {t0}".format(n=n,net=net,sta=sta,t0=t0)
+                    d.elog.log_error('load_site_data',message,ErrorSeverity.Complaint)
                 d['site_lat']=chanrec['latitude']
                 d['site_lon']=chanrec['longitude']
                 d['site_elev']=chanrec['elevation']


pavlis:
I see my changes in database.py broke the test program.  I think my handling of empty history and elog seem to have created a mismatch in the logic of the test for history.   It isn't easy for me to sort that out so this is a case in point that point 3 will require some work on your part to sort out the right solution. 

Since that seems necessary let me suggest one other thing as you revise save_data and the functions it calls.  I think you need to think through some efficiency issues in the flow of that method.   Database updates are expensive in run time and I'm not sure the current flow of that function is the most efficient.   It is complicated enough, however, that I'm far from certain that is true.  I wonder if one strategy to reduce complexity is to make a cleaner separation between an update and an original save.  Having both operations go through a single function adds complexity that doesn't seem essential to me.   That model is my fault since the prototype did that, but maybe that was not the right approach?

wangyinz:
The effort on correcting database methods' behavior duplicated with the pull request #125 that I opened earlier. The dead object and empty history or elog issue should be addressed there already. I have requested your review there.

Yes, I thought about taking apart the save method into an insert and an update. But that requires a major change to the code and the tests, and I feel it not that important at the moment since the save method of the pymongo API is also a mix of insert and replace (although this is a deprecated interface). Let's get the database api working before further polishing it. I still need to add some tests to #125 to make sure the added login makes sense.

pavlis:
So how do you want to handle this.   I thought the database changes I made were duplication of effort but as I said I was being impatient and needed a bandaid to work through other problems.  Can we use the edit features here to accept changes to ensembles.py only and merge that.  

The site_report and channel_report functions are different.  As noted they maybe belong in a different module in mspasspy.preprocessing.seed.   Do you want them there or in database.py as I did here?  You can decide and then just tell me what to do next.  I'm occupied right now on documentation updates since I got the raw data processing workflow prototype completely functional.

wangyinz:
The site_report and channel_report seems to be a print utility only, so we should either turn them into a method of the Database class or move them to a different module. Since they are highly dependent on the schema for a seed channel and are not at all generic, probably moving them to mspasspy.preprocessing.seed is more reasonable.

If the database API's conflict is only in https://github.com/wangyinz/mspass/pull/127/commits/d07be4d877bcc7320f6f630ba64948c0dbd2c556, probably you can just revert this commit and them force push to the origin.

pavlis:
Hmmmm - don't know why we have this conflict.  I did what you suggested with git reset to remove the last commit, but we seem to still have a conflict.   Suggest you accept ensemble.py changes and ignore use the master version of database.py.  There must be some relic conflict in database.py we didn't anticipate.

pavlis:
BTW - I have moved the site_report and channel_report functions to preprocessing.seed.util as we discussed.  That is work in progress that will be pushed later.

wangyinz:
Well, my bad. Turns out https://github.com/wangyinz/mspass/pull/127/commits/32e90964016ab15256c3e38f0b9996406e3be110 also has changes to database.py... Let me revert that too.

PR #137, opened 2021-02-02
Implement check_format for schema

haruming:
* use schema library to create a custom schema template for check

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,10 +4,8 @@ cxx/build/
 docs/build/
 docs/source/doxygen
 docs/source/mspass_schema/*.csv
-build/
 **/_build/
 **/__pycache__/
-**/mspasspy.egg-info
 .idea
-.coverage*
 misc
+.coverage*
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,11 @@ cxx/build/
 docs/build/
 docs/source/doxygen
 docs/source/mspass_schema/*.csv
-**/_build/
+**/build/
 **/__pycache__/
+**/dist/
+**/mspasspy.egg-info
+**/data
 .idea
+.coverage*
 misc
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ docs/source/mspass_schema/*.csv
 **/__pycache__/
 **/dist/
 **/mspasspy.egg-info
-**/data
 .idea
 .coverage*
 misc
--- a/.gitignore
+++ b/.gitignore
@@ -4,10 +4,9 @@ cxx/build/
 docs/build/
 docs/source/doxygen
 docs/source/mspass_schema/*.csv
+build/
 **/_build/
-**/build/
 **/__pycache__/
-**/dist/
 **/mspasspy.egg-info
 .idea
 .coverage*
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ cxx/build/
 docs/build/
 docs/source/doxygen
 docs/source/mspass_schema/*.csv
+**/_build/
 **/build/
 **/__pycache__/
 **/dist/
--- a/python/mspasspy/db/schema.py
+++ b/python/mspasspy/db/schema.py
@@ -4,7 +4,7 @@ Tools to define the schema of Metadata.
 import os
 
 import yaml
-from schema import Schema, And, Or, Optional, SchemaError
+import schema
 import bson.objectid
 
 import mspasspy.ccore.seismic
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -4,7 +4,11 @@ cxx/build/
 docs/build/
 docs/source/doxygen
 docs/source/mspass_schema/*.csv
-**/_build/
```
wangyinz, line 7: Should never remove entries in .gitignore

.gitignore
```diff
@@ -4,7 +4,11 @@ cxx/build/
...
+**/build/
 **/__pycache__/
+**/dist/
+**/mspasspy.egg-info
+**/data
```
wangyinz, line 11: Why adding this? Isn't it gonna ignore the `data` directory that's in our repo?

python/mspasspy/db/schema.py
```diff
@@ -4,7 +4,7 @@
 import os
 
 import yaml
-import schema
+from schema import Schema, Or, Optional, SchemaError
```
wangyinz, line 7: Just a personal preference, but I think `from schema import Schema` here is confusing. It enables such a class: `mspasspy.db.schema.Schema`, which appears to be a class defined by mspasspy but actually isn't.

wangyinz (changes requested):
Please change the `_check_format` function as we discussed offline.

.gitignore
haruming: Got it.

.gitignore
haruming: Because after building of the project, it creates every file in the data repo and git detects this change. Maybe there is something wrong with my build.

python/mspasspy/db/schema.py
haruming: The reason I am doing this is that I don't wanna import the whole library into this program, but what you suggested seems more reasonable to use the previous version.

.gitignore
wangyinz: hmmm... that shouldn't happen. How did you build the package? I usually do ```pip3 install --user ./ -vvv```, which installs the package to `~/.local/`. Can you try that?

python/mspasspy/db/schema.py
wangyinz: I don't think there is a difference in terms of memory usage between the two style of import lines, so this is indeed just a preference thing.

.gitignore
haruming: I ran the command under the python-package.yml. But I tried your command, it raised an error.

pip._internal.exceptions.InstallationSubprocessError: Command errored out with exit status 1: /usr/local/bin/python3.6 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/hw/7p_zv2c16z313fq5jg7lgbmm0000gn/T/pip-req-build-9os00vmv/setup.py'"'"'; __file__='"'"'/private/var/folders/hw/7p_zv2c16z313fq5jg7lgbmm0000gn/T/pip-req-build-9os00vmv/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /private/var/folders/hw/7p_zv2c16z313fq5jg7lgbmm0000gn/T/pip-record-51fwpu0q/install-record.txt --single-version-externally-managed --user --prefix= --compile --install-headers /Users/william/Library/Python/3.6/include/python3.6m/mspasspy Check the logs for full command output.

.gitignore
wangyinz: Actually, let's revert the change to this file since you are able to build the package with pip now.

python/mspasspy/db/schema.py
```diff
@@ -597,5 +597,83 @@ def set_writeable(self, key):
             raise MsPASSError(key + ' is not defined', 'Invalid')
         self._main_dic[key]['readonly'] = False
 
+
+'''
+check if a mspass.yaml file is valid or not
+'''
+def is_basic_type(s):
```
wangyinz, line 604: these additional functions should be considered private, so we should add a `_` prefix to their names. Users are not suppose to call them.

python/mspasspy/db/schema.py
```diff
@@ -597,5 +597,83 @@ def set_writeable(self, key):
...
+    if s in ["ObjectID","int","float","double","complex","bool","string","list","dict","bytes"]:
```
wangyinz, line 605: Check out the `type` method:
https://github.com/wangyinz/mspass/blob/1020dd621b3cb9970858b0522b507767601325d0/python/mspasspy/db/schema.py#L238-L264
You can see that we actually supports more types (except for the "complex"). Note that I also used the `strip()` and `casefold()` there so that extra spaces and cases will also be automatically converted. You might want to do the same here.

python/mspasspy/db/schema.py
```diff
@@ -597,5 +597,83 @@ def set_writeable(self, key):
...
+    db_name_list = get_all_db_name(schema_dic)
+    collection_name_list = []
+    for db_name in db_name_list:
+        collection_name_list += get_all_collection_name_by_db(schema_dic, db_name)
+    
+    type_attribute_schema = schema.Schema({
+        'type':schema.And(str, lambda s: is_basic_type(s)),
+        schema.Optional('reference'):schema.And(str, lambda s: s in collection_name_list),
+        schema.Optional('concept'):str,
+        schema.Optional('aliases'):schema.Or(list, str),
```
wangyinz, line 633: I am not sure that a `str` is supported. Maybe we should automatically convert `str` into a single element `list` here.

python/mspasspy/db/schema.py
```diff
@@ -597,5 +597,83 @@ def set_writeable(self, key):
...
+                                        for attr in dic)))
+    }, ignore_extra_keys=True)
+    
+    metadata_collection_schema = schema.Schema({
+        'schema':schema.And(dict, lambda dic: (metadata_attribute_schema.validate(attribute) for attribute in dic))
+    }, ignore_extra_keys=True)
+    
+    database_schema = schema.Schema({
+        'wf_TimeSeries':schema.And(dict, lambda dic: database_collection_schema.validate(dic)),
+        'wf_Seismogram':schema.And(dict, lambda dic: database_collection_schema.validate(dic)),
```
wangyinz, line 661: wf_TimeSeries and wf_Seismogram are actually not required, instead, what's required really is a collection that is the default `wf`. e.g.:
```
Database:
  wf_something:
    default: wf
```
The rule is that we need one collection that has `default: wf` defined, or if there is no `default: wf`, we will then need a collection that is named as `wf` itself (because `wf` itself will be the default name when not specified).

This is my bad - another implicit rule that I didn't mention anywhere. Actually, `elog` and `history_object` are also required default names, just like the `wf`.

python/mspasspy/db/schema.py
```diff
@@ -597,5 +597,83 @@ def set_writeable(self, key):
...
+        schema.Optional('site'):schema.And(dict, lambda dic: database_collection_schema.validate(dic)),
+        schema.Optional('channel'):schema.And(dict, lambda dic: database_collection_schema.validate(dic)),
+        schema.Optional('source'):schema.And(dict, lambda dic: database_collection_schema.validate(dic)),
```
wangyinz, line 664: Instead of specifying them optional, the rule here really should be "for any entries under `Database`, its schema need to follow `database_collection_schema`"

python/mspasspy/db/schema.py
```diff
@@ -597,5 +597,83 @@ def set_writeable(self, key):
...
+        schema.Optional('history_object'):schema.And(dict, lambda dic: database_collection_schema.validate(dic)),
+        schema.Optional('elog'):schema.And(dict, lambda dic: database_collection_schema.validate(dic)),
+    }, ignore_extra_keys=True)
+    
+    metadata_schema = schema.Schema({
+        'TimeSeries':schema.And(dict, lambda dic: metadata_collection_schema.validate(dic)),
+        'Seismogram':schema.And(dict, lambda dic: metadata_collection_schema.validate(dic)),
```
wangyinz, line 671: These two are OK for now, but it should be similar to the `Database` schema, so that we can support data types other than `TimeSeries` and `Seismogram` in the future.

wangyinz (changes requested):
Very well done! Just had some minor issues (which are mostly my bad for not being explicit)

.gitignore
haruming: Yes! The only things I add here are
**/build/(the folder created in my project after build)
**/mspasspy.egg-info(same above)
.coverage*(files created after running pytest)

Are those path are good to add to .gitignore?

python/mspasspy/db/schema.py
haruming: Got it.

python/mspasspy/db/schema.py
haruming: That's a good pre-checking step, which I didn't notice!

haruming: Also, should we support the complex type here? I guess we should be consistent with the above code.

python/mspasspy/db/schema.py
haruming: because I notice our mspass.yaml template has attributes that their aliases is just a string(see Database:wf_TimeSeries:schema:starttime_shift). I am not sure if it is expected here.

python/mspasspy/db/schema.py
haruming: It's ok, I should have asked you about this first:) so from my understanding, a very basic schema should at least contain a collection called `wf` or a collection that has `default` attribute which equals to `wf`. Then `elog` and `history_object` are collections that related to `wf`.

.gitignore
wangyinz: It is fine to add those, but you need to remove the `*` because it will match any thing with that name under any sub-directories, which is not good. Note that I had things like `**/_build/` mostly because they start with `_`, which is distinctively different from normal files.

Also, if you build with the pip command, `build` and `mspasspy.egg-info` shouldn't be created under current directory.

python/mspasspy/db/schema.py
wangyinz: Complex is not a BSON-supported type, so we cannot have it anyway.

python/mspasspy/db/schema.py
wangyinz: hmmm... so we do support that. Good catch. I think that was a unintentional mistake, but I just checked and found that we do support str.

python/mspasspy/db/schema.py
wangyinz: Exactly. The rule for `wf` also applies to `elog` and `history_object`.

python/mspasspy/db/schema.py
wangyinz: I am taking that back. Turns out some methods support that but some others don't:
```
>>> from mspasspy.db.schema import DatabaseSchema
>>> dbschema = DatabaseSchema()
>>> dbschema.wf_TimeSeries.aliases('starttime_shift')
't0shift'
>>> dbschema.wf_TimeSeries.has_alias('starttime_shift')
True
>>> dbschema.wf_TimeSeries.unique_name('t0shift')
  File "<stdin>", line 1, in <module>
  File "/home/iwang/.local/lib/python3.6/site-packages/mspasspy/db/schema.py", line 285, in unique_name
    raise MsPASSError(aliasname + ' is not defined', 'Invalid')
mspasspy.ccore.utility.MsPASSError: t0shift is not defined
>>> dbschema.wf_TimeSeries.is_alias('t0shift')
False
>>> dbschema.wf_TimeSeries.is_defined('t0shift')
False
```

wangyinz: btw, this is what happened:
```
>>> dbschema.wf_TimeSeries._alias_dic
{'nsamp': 'npts', 'wfdisc.nsamp': 'npts', 'dt': 'delta', 't0': 'starttime', 'time': 'starttime', 't': 'starttime_shift', '0': 'starttime_shift', 's': 'starttime_shift', 'h': 'starttime_shift', 'i': 'starttime_shift', 'f': 'starttime_shift', 'file.dir': 'dir', 'wfdisc.dir': 'dir', 'wfprocess.dir': 'dir', 'file.dfile': 'dfile', 'wfdisc.dfile': 'dfile', 'wfprocess.dfile': 'dfile', 'file.foff': 'foff', 'wfdisc.foff': 'foff', 'wfprocess.foff': 'foff'}
```
The `t0shift` is taken apart into characters.

wangyinz (approved):
Looks good to me. Just a note here: `readonly` is only for metadata schema, but here the `type_attribute_schema` is used to check both metadata and database. It is optional anyway, so it doesn't affect anything. We might clean this up later if time allows.

PR #141, opened 2021-02-06
Dev pavlis

pavlis:
The simplest fixes in this set of changes are the the preprocessing/seed/ensembles.py file.  They should merge clean.  The harder thing to merge is a bug fix that also made the function formerly known as load_seed_channel (now changed to get_seed_channel) more robust.  I previously could mysteriously return nothing.  It still can, but only when the query really fails.  There were cases where the previous version would just not fail when it should have worked because of errors in the algorithm.  

I also changed load_seed_site to get_seed_site.  I thought that was a good word to use because it is fetching the entire document matching a match query.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -1189,7 +1189,7 @@ class Database(pymongo.database.Database):
         print("number of channel records saved=", n_chan_saved)
         return tuple([n_site_saved, n_chan_saved, n_site_processed, n_chan_processed])
 
-    def load_inventory(self, net=None, sta=None, loc=None, time=None):
+    def read_inventory(self, net=None, sta=None, loc=None, time=None):
         """
         Loads an obspy inventory object limited by one or more
         keys.   Default is to load the entire contents of the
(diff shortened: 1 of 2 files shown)


wangyinz:
Seems this is clean to be merged. Do you want to include some tests in this branch before merging?

pavlis:
It would be good to add tests for get_seed_site and get_seed_channel in the test program for database.   Is there a test for save_inventory and save_catalog?  That is needed too if they aren't yet present.

The simplest way to build tests for all of the above is to utilize pieces of the getting_started tutorial I've been working on.   I pushed a working branch last night [here](https://github.com/wangyinz/mspass_tutorial/tree/tutorial_updates).   The tutorial uses obspy's web services to pull catalog and inventory data and then saves them with the db calls.   That would test save_inventory and save_catalog.   The get functions would ned a little more work.  It would particularly be helpful to test all the cases get_seed_channel has to handle.   Testing it might allow you to critically evaluate whether or not the way I designed those functions is ideal.  I made both have required args with options.   Since the objective is always to return a single matching document to match net:sta or net:sta:chan:loc that seemed appropriate.   The get_seed_channel function has to deal with the oddities of how on might handle loc:  loc can be undefined, the literal string "", the literal string  "  " (two space characters), or a defined value like "00".   That was he functionality I added in this version but I absolutely did not test all the conditions.

PR #143, opened 2021-02-08
first version of db refining

haruming:
I wrote a basic version of the normalization methods for the database based on the documentation.

My idea is:
1. get the metadata schema based on the mspass object type
2. get the source_id from the current mspass_object
3. find the unique document associated with this source id in the source collection
4. update the mspass object based on the document by using the direct `put` method in Metadata class.

There may be something in detail that I was unaware of or ignored. Correct me if I am wrong.


python/mspasspy/db/database.py
```diff
@@ -570,6 +571,150 @@ def update_ensemble_metadata(self, ensemble_object, include_undefined=False, exc
...
+
+        if not isinstance(mspass_object, (TimeSeries, Seismogram)):
+            raise MsPASSError("only TimeSeries and Seismogram are supported", ErrorSeverity.Invalid)
+
+        if collection == 'channel' and isinstance(mspass_object, (Seismogram, SeismogramEnsemble)):
+            raise MsPASSError("channel data can not be loaded into Seismogram", ErrorSeverity.Invalid)
+        
+        # 1. get the metadata schema based on the mspass object type
+        wf_collection = 'wf_TimeSeries' if isinstance(mspass_object, TimeSeries) else 'wf_Seismogram'
+        object_type = self.database_schema[wf_collection].data_type()
```
wangyinz, line 609: Object type is already determined by the input argument `mspass_object`. Why getting it from the database schema here? Also, the above line should be removed:
```wf_collection = 'wf_TimeSeries' if isinstance(mspass_object, TimeSeries) else 'wf_Seismogram'```

File: python/mspasspy/db/database.py [hunk 2]
```diff
@@ -570,6 +571,150 @@ def update_ensemble_metadata(self, ensemble_object, include_undefined=False, exc
...
+        
+        :param include_undefined:  when true all data in the matching document are loaded.
+
+        :param collection: requested collection metadata should be loaded
+        :type collection: str 
+
+        :exception:  any detected errors will cause a MsPASSError to be thrown
+        (colleagues:  this may be wrong sphynx syntax for defining an exception)
+        """
+        if collection not in ["source", "site", "channel"]:
```
wangyinz, line 595: This is not necessary as this method should be flexible for any collections. It will check whether `collection + '_id' exists later so, if a collection is not defined in the object, it will throw an error.
`

python/mspasspy/db/database.py
```diff
@@ -570,6 +571,150 @@ def update_ensemble_metadata(self, ensemble_object, include_undefined=False, exc
...
+            raise MsPASSError("collection param should be one of source, site or channel", ErrorSeverity.Invalid)
+
+        if not mspass_object.live:
+            raise MsPASSError("only live mspass object can load metadata", ErrorSeverity.Invalid)
+
+        if not isinstance(mspass_object, (TimeSeries, Seismogram)):
+            raise MsPASSError("only TimeSeries and Seismogram are supported", ErrorSeverity.Invalid)
+
+        if collection == 'channel' and isinstance(mspass_object, (Seismogram, SeismogramEnsemble)):
```
wangyinz, line 604: Likewise, this should also be redundant (but doesn't really hurt).

python/mspasspy/db/database.py
```diff
@@ -570,6 +571,150 @@ def update_ensemble_metadata(self, ensemble_object, include_undefined=False, exc
...
+        # 1. get the metadata schema based on the mspass object type
+        wf_collection = 'wf_TimeSeries' if isinstance(mspass_object, TimeSeries) else 'wf_Seismogram'
+        object_type = self.database_schema[wf_collection].data_type()
+        if object_type not in [TimeSeries, Seismogram]:
+            raise MsPASSError('only TimeSeries and Seismogram are supported, but {} is requested. Please check the data_type of {} collection.'.format(
+                object_type, wf_collection), 'Fatal')
+        wf_collection_metadata_schema = self.metadata_schema[object_type.__name__]
+
+        collection_id = collection + '_id'
+        # 2. get the source_id from the current mspass_object
```
wangyinz, line 616: source_id -> collection_id

python/mspasspy/db/database.py
```diff
@@ -570,6 +571,150 @@ def update_ensemble_metadata(self, ensemble_object, include_undefined=False, exc
...
+        # 3. find the unique document associated with this source id in the source collection
+        object_doc = self[collection].find_one({'_id': object_doc_id})
+        if object_doc == None:
+            raise MsPASSError("no match found in {} collection for source_id = {}".format(collection, object_doc_id), ErrorSeverity.Invalid)
+        
+        # 4. use this document to update the mspass object
+        for k in wf_collection_metadata_schema.keys():
+            col = wf_collection_metadata_schema.collection(k)
+            if col == collection:
+                if include_undefined:
```
wangyinz, line 630: I don't think `include_undefined` is functioning here. The loop is driven by `wf_collection_metadata_schema.keys()` which are all the defined keys. The undefined keys in `object_doc` will never be reached.

wangyinz (changes requested):
Yeah, this is pretty much what we needed. I just found a couple minor bugs in this initial implementation. Also, @pavlis may want to review it, too.

python/mspasspy/db/database.py
haruming: I am a little bit confused about the scope of `include_undefined` and `exclude_keys` parameters. From my understanding, `include_undefined` should include the keys in the `object_doc`, while `exclude_keys` excludes the keys in the metadata. Correct me if I am wrong.

haruming:
I fixed some bugs that appeared above, hopefully, this is what we need in the normalization methods

python/mspasspy/db/database.py
```diff
@@ -570,6 +571,147 @@ def update_ensemble_metadata(self, ensemble_object, include_undefined=False, exc
...
+            raise MsPASSError("only live mspass object can load metadata", ErrorSeverity.Invalid)
+
+        if not isinstance(mspass_object, (TimeSeries, Seismogram)):
+            raise MsPASSError("only TimeSeries and Seismogram are supported", ErrorSeverity.Invalid)
+
+        if collection == 'channel' and isinstance(mspass_object, (Seismogram, SeismogramEnsemble)):
+            raise MsPASSError("channel data can not be loaded into Seismogram", ErrorSeverity.Invalid)
+        
+        # 1. get the metadata schema based on the mspass object type
+        wf_collection_metadata_schema = self.metadata_schema[type(mspass_object).__name__]
```
wangyinz, line 605: Here we 'd better do the following:
```python
        object_type = self.database_schema[wf_collection].data_type()
        if object_type not in [TimeSeries, Seismogram]:
            raise MsPASSError('only TimeSeries and Seismogram are supported, but {} is requested. Please check the data_type of {} collection.'.format(
                object_type, wf_collection), 'Fatal')
        read_metadata_schema = self.metadata_schema[object_type.__name__]
```
which is how `read_data` handled object type.

haruming:
Continue my work when power/internet is back! So far, I have implemented the required changes we discussed before in the meeting. However, there might be some discrepancies between my implementation and the documentation. Correct me and leave a comment if you think I am wrong.

python/mspasspy/db/database.py
```diff
@@ -308,39 +341,93 @@ def save_data(self, mspass_object, storage_mode='gridfs', dfile=None, dir=None,
...
             # FIXME: we could have recorded the full stack here, but need to revise the logger object
             # to make it more powerful for Python logging.
             mspass_object.elog.log_verbose(
                 sys._getframe().f_code.co_name, "Skipped saving dead object")
             self._save_elog(mspass_object)
+        
+        return 1 if save_metadata_fail else 0
+
+    # clean the collection fixing any type errors and removing any aliases using the schema currently defined for self
+    def clean(self, collection, log_id_keys=[], print=False, query=None):
```
wangyinz, line 383: This is not covered by test.

python/mspasspy/db/database.py
```diff
@@ -308,39 +341,93 @@ def save_data(self, mspass_object, storage_mode='gridfs', dfile=None, dir=None,
...
+                    fixes.append("{}errors are fixed successfully".format(log_helper))
+        
+        if print:
+            for fix in fixes:
+                print(fix)
+        
+        return fixes
 
-    def update_metadata(self, mspass_object, include_undefined=False, exclude_keys=[], collection=None):
+    def update_metadata(self, mspass_object, mode='cautious', exclude_keys=[], collection=None, ignore_metadata_changed_test=False):
```
wangyinz, line 430: need to check for mode upfront.

wangyinz (changes requested):
Very good! I have posted some inline comments. There are three generic items need to be done: implement the changes discussed in https://github.com/wangyinz/mspass/discussions/146, make sure any new code is covered by sufficient tests, rearrange the test so we can clearly see how the tests for different modes are done.

wangyinz:
@pavlis You might want to take a look at the API and see if there is anything missing.

data/yaml/mspass_lite.yaml
```diff
@@ -7,85 +7,97 @@ Database:
...
+          constraint: required
         delta:
           type: double
           concept: Data sample interval in seconds
           aliases: [dt]
+          constraint: required
         sampling_rate:
           type: double
           concept: Data sampling frequency in Hz=1/s
-          optional: true
```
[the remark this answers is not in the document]
wangyinz: Actually, I think the methods proposed here are more useful than the `constraint` method that you implemented right now.

python/mspasspy/db/database.py
wangyinz: implement another `clean` method for document and change this one to `clean_collection` and wrap around `clean`

python/mspasspy/db/database.py
```diff
@@ -187,91 +192,153 @@ def read_data(self, object_id, load_history=False, include_undefined=False, excl
...
-        if load_history:
-            history_obj_id_name = self.database_schema.default_name('history_object') + '_id'
-            if history_obj_id_name in object_doc:
-                self._load_history(mspass_object, object_doc[history_obj_id_name])
+        # not continue step 2 & 3 if the mspass object is dead
+        if is_dead:
+            mspass_object.kill()
+            for msg in log_error_msg:
+                mspass_object.elog.log_error('read_data', msg, ErrorSeverity.Invalid)
+            self._save_elog(mspass_object)
```
wangyinz, line 288: We shouldn't save the object to elog's graveyard within a read method. That mechanism is designed for save only. Remove this line should fix it.

wangyinz (commented):
I realized this review request was only for the tests when working half way through it. Anyway, I did notice an issue in the save method, so please go ahead and fix that. Other than that, the tests looks good to me. Thanks!

pavlis:
I plead forgiveness as the old guy and will insert his comment.  I am confused by the review process here because the very long list of comments and changes already in this pull request.  I don't think the comment here was addressed earlier.  Here is the point.

I do not thing this is the right behavior in this section of the clean method of Database:
```
# check if the document has all the rquired fields
        missing_required_attr_list = []
        for k in self.database_schema[collection].keys():
            if self.database_schema[collection].is_required(k) and k not in doc:
                missing_required_attr_list.append(k)
        if missing_required_attr_list:
            # delete this document
            col.delete_one({'_id': doc['_id']})
            error_msg = "required attribute "
            for missing_attr in missing_required_attr_list:
                error_msg += "{} ".format(missing_attr)
            error_msg += "are missing."
            print_messages.append("{}{} the document is deleted.".format(log_helper, error_msg))
            return fixed_cnt, print_messages
```
Deleting documents with required  attributes missing is one thing one might want to do, BUT it should only be an option.  I suggest the clean method have an option `delete_missing_required=False` that does what the verbose name suggests.   i.e. the block above is only executed if that parameter is set true.  Otherwise it should log an error.   

The reason that should require an explicit action by the user and not happen by default is nearly universal in database transactions.  A delete is the most extreme action you can take and can result in the loss of data.   It shouldn't happen unless i is clear to a user it is could.   If this were a gui or an command line program it is the kind of action that would initiate a "are you sure?" message before proceeding.   

This brings up another point we haven't all discussed that perhaps belongs in discussions:  we may want to create a few basic database edit functions to insulate newbies from needing to immediately learn the incantations of pymongo.

pavlis:
Just pushed edits to yaml files.   Did the following:
1.  Changed a few things of the "constraint" values.   
2. I think there was an error in the mspass_lite.yaml file I fixed.   I'm surprised the file was readable.  I don't think it was a misunderstanding on my part of the yaml syntax.
3. Ian you should review the set of additions I made at the very end of mspass.yaml.  I suggested a number of names for common metadata to handle arrival time data.   I am not sure I put them in the right place, but got the information down.

data/yaml/mspass.yaml
```diff
@@ -34,6 +34,10 @@ Database:
         type: ObjectID
         concept: ObjectId used to define a data object
         constraint: required
+      data_tag:
+        type: string
+        concept:  Used to label data at a particular stage of processing. Should be unique for each save operation.
+        constraint: normal
```
wangyinz, line 40: As I mentioned in #158, this is not necessary to be in the schema. We should just implement the data_tag as part of the API.

data/yaml/mspass.yaml
```diff
@@ -330,13 +334,13 @@ Metadata:
         readonly: false
       site_id:
         collection: wf_TimeSeries
-        readonly: false
+        readonly: true
```
wangyinz, line 337: We (with Jinxin) decided to set these `"collection"_id` keys as writable for a reason. These keys correspond to the same attributes in the wf collection, which links a wf record to certain site/channel/source records. They are different from the actual `_id` keys, and we need to be able to change them to correctly associate a newly created wf record. If I remember right, there are codes that rely on this field to do save correctly.

data/yaml/mspass.yaml
```diff
@@ -429,13 +433,10 @@ Metadata:
         readonly: false
       site_id:
         collection: wf_Seismogram
-        readonly: false
-      channel_id:
-        collection: wf_Seismogram
-        readonly: false
```
wangyinz, line 393: The `channel_id` is a list defined in `wf_Seismogram`. It should not be removed here.

data/yaml/mspass.yaml
```diff
@@ -554,3 +555,43 @@ Metadata:
...
+        concept: y-component (+north) of a measured slowness vector (s/km)
+        readonly: false
+      ux_modeled:
+        type: double
+        concept: x-component (+east) of a slowness vector (s/km) computed from an Earth model.
+        readonly: false
+      uy_modeled:
+        type: double
+        concept: y-component (+north) of a slowness vector (s/km) computed from an Earth model.
+        readonly: false
```
wangyinz, line 597: These belongs to a different yaml file. Also, in that file, we need to define corresponding keys in the Database schema first. Then, we can add them to the Metadata schema.

data/yaml/mspass_lite.yaml
```diff
@@ -2,102 +2,105 @@ Database:
...
-          type: ObjectID
-          concept: The _id of data file in the fs.files collection.
-          constraint: optional
-        history_object_id:
-          reference: history_object
-          constraint: xref_key
-        elog_id:
-          type: list
-          reference: elog
-          constraint: xref_key
```
wangyinz: These were not syntax error. We were explicitly excluding `chan` in `&wf_TimeSeries` with previous syntax. I remember we discussed this in the emails before.

wangyinz (changes requested):
I am not sure how this broke the test, but there are quite a few issues in the unnecessary changes. See details in the inline comments.

pavlis:
Well I obviously messed a lot of up there and caused more problems than I fixed.   I'll leave it to you guys to fix my mistakes.  I would also request we spend some time on our call tomorrow educating me and discussing how we see his schema being used.   I think I am not understanding your thinking.

wangyinz:
This is merged by #156.

PR #147, opened 2021-02-15
New documentation pages and edits to existing

pavlis:
I pushed this because the new document on CRUD operations includes a lot of prototype suggests on how the database api should behave.  It is mostly up to date with things on the discussion pages and in the google design doc.   I would hope it provides a new summary of how we currently think it should behave.   Suggest you merge it to master as there shouldn't be any conflicts.  Well, there might be some conflicts with documentation in a previous pull request that hasn't yet been merged with master.  Followed the directive we decided last Thursday to when in doubt add a new branch from the most current master.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -51,7 +51,7 @@ templates_path = ['_templates']
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = []
+exclude_patterns = ['that_style']
 
 
 # -- Options for HTML output -------------------------------------------------
--- a/docs/source/cxx_api/index.rst
+++ b/docs/source/cxx_api/index.rst
@@ -5,9 +5,9 @@ The MsPASS C++ API's key components are the following classes.
 
 .. toctree::
 
-   ../cxx_api/mspass.Metadata
-   ../cxx_api/mspass.TimeSeries
-   ../cxx_api/mspass.Seismogram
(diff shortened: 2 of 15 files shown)


wangyinz:
Looks good to me. Have you tried compiling this document locally? I think you need to add links to `docs/source/user_manual.rst` to make the new pages visible.

pavlis:
I haven't tried to build this locally and a big reason is I have never been able to get the full documentation build system to work correctly.  There is some oddity with the setup on my mac laptop I've never resolved.   I can't get the ubuntu version to work either.  Never considered it worth the trouble to sort out as it has only proven to be a minor annoyance like this.  I didn't do that because I had no way to verify it worked so I "kicked can down the road" to use a bad cliche.  

Note also there is another glaring hole.  I haven't yet constructed a figure the new CRUD document mentions.   I pushed this so you guys could use this as a secondary reference as we clean up the database api.

wangyinz:
Yep... just as expected, it does not compile right. I spent quite some time on fixing the links and formats. It at least builds all the contents correctly now. Probably still need a thorough clean down the road.

pavlis (commented):
Thanks for fixing these.  I see that problems you had to fix in rst file were in sections I did before getting a local previewer working with Atom.  

The other item to note is the adapting_algorithms.rst page is incomplete.   I did that one some time ago.  It may be out of date and is definitely missing some examples we probably need.  I think there are some stubs in the file.  When this gets merged I can try to repair it.

I see no reason you shouldn't merge this with master so we can build on this.

PR #155, opened 2021-03-07
Fix ensemble md

pavlis:
This branch has several significant changes all of which I think will be orthogonal to any changes being done by anyone else in the group. 

1.  As the branch name says this includes a change to the converters to properly handle ensemble metadata.
2. I added lines the relevant test program (forget the name) to test my changes in the converter code.
3. Testing uncovered a gap in the C++ code for the function to extract a single component from a Seismogram object.  The return needed to set hang and vang to preserve any orientation information.
4. I tested that code by extending the C++ test_tcs program.  In the process I realized test_tcs was written before I knew the right structure for a git test program.  I added a long list of asserts to make test_tcs do what it should.  
5. In that process I uncovered a very bad bug in two tiny little functions buried in there to compute degrees to radians and radians to degrees.  The constants were backward and they were producing crazy result since the convertions were inverted.   I don't think we used those functions much, but still was a very serious bug found in computing hang and vang for extract components.


pavlis:
Forgot about the first commit in that group.   Fixed a bug in the SectionPlotter class image plot method. Small but important change that is a very different fix than the rest of this branch.

PR #156, opened 2021-03-08
database API refinement

haruming:
1. merging Gary's command-line tools
2. modify the database API according to meeting and discussion.
3. add more unit testing


haruming:
ready for merge.

python/mspasspy/db/database.py
```diff
@@ -184,86 +192,143 @@ def read_data(self, object_id, load_history=False, include_undefined=False, excl
         object_doc = col.find_one({'_id': oid})
         if not object_doc:
             return None
+        
+        if 'data_tag' in object_doc and object_doc['data_tag'] != data_tag:
```
wangyinz, line 196: I think this should be 
```
if not data_tag and 'data_tag' not in object_doc or object_doc['data_tag'] != data_tag:
    return None
```

python/mspasspy/db/database.py
```diff
@@ -308,53 +375,726 @@ def save_data(self, mspass_object, storage_mode='gridfs', dfile=None, dir=None,
...
+        # collection names.  For now we just freeze the names 
+        # and put them in this one place for maintainability
+        norm_collection_list=['site','channel','source']
+        wf_collection_list=['wf_TimeSeries','wf_Seismogram']
+
+        if not (normalize in norm_collection_list):
+            raise MsPASSError('check_links:  illegal value for normalize arg='+normalize,
+                            'Fatal')
+        if not (wf in wf_collection_list):
+            raise MsPASSError('check_links:  illegal value for wf arg='+wf,
```
wangyinz, line 846: The `wf_collection_list` and its checking is pointless here. Instead, we should only check whether `wf` is in the database or not. This should be done similar to the `read_data` method:
```
        try:
            wf_collection = self.database_schema.default_name(collection)
        except MsPASSError as err:
            raise MsPASSError('collection {} is not defined in database schema'.format(collection), 'Invalid') from err
```
Well, probably this method should be defined as:
```
def check_links(self, normalize='site', collection="wf", query={}, verbose=False, error_limit=1000):
```

The `norm_collection_list` should also be removed. Instead, we can implement a `xref_keys` method in the database schema just like the existing `required_keys` method. We can build the normalized collection list by removing the `'_id'` part of the xref_keys. 

Actually, thinking more about it, I believe this `check_links` method should be very generic that it can check links for any collections with xref_keys defined. By applying the changes above, it should already be able to do that.

python/mspasspy/db/database.py
```diff
@@ -308,53 +375,726 @@ def save_data(self, mspass_object, storage_mode='gridfs', dfile=None, dir=None,
...
+                            col_schema.type(k))
+                        print('This error was thrown and handled:  ')
+                        print(err)
+            
+            if n>0:
+                dbcol.update_one({'_id' : id},{'$set' : up_d})
+                
+        return counts
+
+    def check_links(self, normalize='site', wf="wf_TimeSeries", wfquery={}, verbose=False, error_limit=1000):
```
wangyinz, line 796: Maybe these newly added method should be defined private at least for now? I can see the usage of them are very limited which makes it confusing if we open them to the users.

python/mspasspy/db/database.py
```diff
@@ -308,53 +375,726 @@ def save_data(self, mspass_object, storage_mode='gridfs', dfile=None, dir=None,
...
+                    if self.database_schema[collection].is_defined(key) and self.database_schema[collection].is_xref_key(key):
+                        # only test normalized keys with '_id'
+                        if '_id' in key and key.rsplit('_', 1)[1] == 'id':
+                            normalized_collection_name = key.rsplit('_', 1)[0]
+                            normalized_collection_name = self.database_schema.default_name(normalized_collection_name)
+                            normalized_col = self[normalized_collection_name]
+                            # try to find the referenced docuement
+                            normalized_doc = normalized_col.find_one({'_id': doc[key]})
+                            if not normalized_doc:
+                                problematic_keys[key] = test
```
wangyinz, line 606: It seems we can use these in the `check_links` method instead of what's there right now. It seems the same argument applies to the undefined test and the type test below.

wangyinz (changes requested):
Very well done! I only found some small issues and most of them are actually in the code merged from Gary's branch. Please find them in the inline comments below. I think we should merge as soon as they are fixed.

wangyinz (approved):

PR #161, opened 2021-03-16
Fix incorrectly named keyword

pavlis:
This is a trivial fixed in the c++ code for mistake in the name for souce depth.  Was incorrectly tagged as SEISMICMD_selev and set as source_selev.  Changed selev to sdepth.  

Only complication is watch the history.  I incorrectly forgot to switch the branch name and originally committed these to master.  To allow me to push it to github I did
```
git checkout fix_keyword_bug
git merge master
```
then pushed it here.   Not sure if that will mess up the history chain or not.


wangyinz:
Looks good to me. Just a note here. It won't mess up the history. You probably want to drop the changes to local master branch and sync it with the remote.

PR #171, opened 2021-03-29
Feature Added: Global history Manager

haruming:
1. fixed the issues we talked about at the last meeting
2. added features I missed
3. add docstrings


python/mspasspy/global_history/manager.py
```diff
@@ -0,0 +1,261 @@
...
+            # get the algorithm name
+            alg_name = alg[:alg.index('(')].strip()
+        if not parameters:
+            # get the parameters
+            parameters = alg[alg.index('(')+1:].strip()
+        
+        # get the alg_id if exists, else create a new one
+        alg_id = global_history.get_alg_id(alg_name, parameters)
+        if not alg_id:
+            alg_id = ObjectId()
```
wangyinz, line 43: It seems the `get_alg_id` related code can be put into the `logging` method. The `logging` method can then be defined as `logging(alg_name, parameters)`

.github/workflows/python-package.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - name: Install Python Dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install pymongo numpy obspy pytest pytest-cov codecov "dask[complete]" schema
+        python -m pip install pymongo numpy obspy pytest pytest-cov dill codecov "dask[complete]" schema
```
wangyinz, line 27: Need to rebase to current master and add the `dill` module to the new `requirements.txt`.

wangyinz (changes requested):
Well done! I just have some minor comments here. A more important issue I noticed during the review is that the global history and object-level history are currently disconnected. We need to complete our design doc and then have them linked.

python/mspasspy/global_history/manager.py
```diff
@@ -0,0 +1,261 @@
...
+        self.job_name = job_name
+        # generate an bson UUID for this job, should be unique on the application level
+        self.job_id = ObjectId()
+
+        self.history_db = database_instance
+
+        self.collection = collection
+        if not self.collection:
+            # use the `history` collection defined in database schema
+            self.collection = 'history'
```
wangyinz, line 169: Need to update mspass.yaml and use the `default_name` method to get the default collection name.

python/mspasspy/global_history/manager.py
wangyinz: After revising the design doc, I found that `alg_id` is actually needed here.

wangyinz:
Please go checkout what I just added to our global history design doc and see if there is anything missing.

wangyinz (changes requested):
Very good! I haven't gone through all the changes yet, but have found some minor issues that you might want to address right away. I will take a closer look at the global history manager later.

python/mspasspy/global_history/manager.py
```diff
@@ -0,0 +1,361 @@
...
+import dask.bag as daskbag
+
+from bson.objectid import ObjectId
+from mspasspy.ccore.utility import MsPASSError
+from datetime import datetime
+from dill.source import getsource
+
+import mspasspy.algorithms.signals as signals
+
+def mspass_spark_map(self, func, *args, global_history=None, object_history=False, alg_id=None,
```
wangyinz, line 15: I found that some of the docstrings are out of sync right now. For example, the `alg_id` is missing here. You probably want to add something like: 
```
:param alg_id: a user specified alg_id for the map operation
:type alg_id: :class:`str`/:class:`bson.objectid.ObjectId`
```
Also, the description doesn't say anything about `alg_id`, which should definitely be added.

haruming: Yes, I forgot to sync the docstring before. Now it's updated.

wangyinz (approved):

PR #174, opened 2021-04-08
Fix an invalid example code

### LINKED ISSUE #172: Invalid example in the user manual
In this section (https://wangyinz.github.io/mspass/user_manual/data_object_design_concepts.html#scalar-versus-3c-data), the code shows how to create a simple seismogram, but the code is no longer valid.
```
from mspasspy import Seismogram
d=Seismogram(100)  # Create an empty Seismogram with storage for 100 time steps initialized to all zeros
d.data(0,50)=1.0      # Create a delta function at time t0+dt*50 in component 0
```

seisman:
Closes #172.


wangyinz:
Thanks for the contribution! We will merge it after fixing the bugs in current master branch.

PR #180, opened 2021-04-20
Add libmseed

pavlis:
Here is an initial implementation of the combined C/C++ and python code to build a raw parallel reader for miniseed data.   See the discussion page on this for additonal points of the status.   You should definitely review this critically before merging it.


pavlis:
Hmm - I don't think the failures here are mine.   The version of master I branched from must have had test failures.

pavlis:
I just pushed the revisions that I hope will allow us to read miniseed within the read_data method of Database.   You should review some of my changes to Database code carefully.  I had to resolve some conflicts.   I took only my versions because I knew those passed my tests.   I think you said you would take it from here to assimilate the libmseed code and wrappers into the read_data methods. 

Turned out it was trivial to extract sample rate and not that hard to also compute the number of points in a block of data and the endtime.  One thing I must caution you about, however, as you adapt this to the read data method is to NOT trust the npts and endtime values posted to the wf_miniseed collection.    The reason is I don't have a clue how obspy's miniseed reader handles missing packets.   The computed npts here and endtime will be wrong if the are missing packets.   The C code just adds up the number of samples posted for each packet.  It does not check the actual time stamps.  

I wasn't sure if the wf_miniseed schema yaml file needed any changes.   sample rate, endtime, and npts are already defined there.  One thing you probably should check is that I didn't mix up the keywords.   i.e. I didn't check against the schema to verify the keys in wf_miniseed are consistent with wf_TimeSeries.  I should have - sorry about that but I'll punt that to you given I've already checked this in.

wangyinz:
Thanks for the fix! I saw that the history is sort of broken, so I just manually added the changes you made and force-pushed this branch. I also noticed that the `npts` was not right for the sample mseed file that I downloaded from FSDN, so I decided to drop the `npts` and `endtime` key in the database index. Note that `endtime` was never defined in our schema anyway.

PR #182, opened 2021-04-21
bug fixed for reader

haruming:
fix the reader that converts the attribute to the correct type. It should overwrite the attribute value of the key in the metadata we have built.


wangyinz:
Cool! Thanks for fixing it!

PR #187, opened 2021-04-24
Cpp bug fixes

pavlis:
This set of changes fixes two issues in ccore that were problematic:
1.  Remove the key encoding that messed up C++ interface for operator<<.  At the same time I deleted the relic serialziation and deserialization code superceded by the version using pybind11 directly.  We had discussed this issue in our last group meeting.
2. Fixed a potentially nasty bug in handling const char * values in Metadata.  This bug would not happen in python but could and would be created in C++ code if the one used this kind of construct in C++:  `md.put("foo","bar");`   The reason is that boost any documentation warns that the value saved in a boost any container must be "copy constructable" and explicitly warn about C char arrays ([see the boost docs here](https://theboostcpplibraries.com/boost.any).   Easily fixed by overriding the put method for const char * values.   Unknown if there were bugs hidden before due to this problem, but it is likely there were in the deconvolution code.  I discovered the problem in converting the pwmig code as I had that construct in a library routine and when I tested it it threw an unexpected exception from boost.


pavlis:
Looks like you may have to resolve some version skew issues from when I forked this from master. Sorry about that - I don't remember for sure when this fork was created but it wasn't very long ago.  No doubt a reflection of multiple changes you guys have been doing during the same period I saw fly by in emails from github.

PR #189, opened 2021-04-26
Update start-mspass.sh and add scripts for deploying single/distributed nodes in HPC

haruming:
1.  update start-mspass.sh to support start containers in a distributed manner. 
2.  upload script for deploying a single node in HPC
3. upload script for deploying a distributed nodes in HPC(start worker & shard in worker nodes, start DB manager & scheduler in primary node)


scripts/start-mspass.sh
```diff
@@ -98,5 +157,4 @@ if [ $# -eq 0 ]; then
   fi
 else
   docker-entrypoint.sh $@
-fi
-
+fi
```
wangyinz, line 160: Just a note here, it is preferred to always have a newline at the end of any files.

PR #191, opened 2021-04-30
Improve db docstrings

pavlis:
This branch has major changes to the docstrings to expand the descriptions.  In writing these a long list of points/issues arose that I recorded in a set of notes.   Below are the notes pasted verbatim:
/////Notes from revising docstrings and CRUD user's manual section/////

1.  When storage_mode is file save_data currently required BOTH dir and
dfile be set.  I don't think we should be that dogmatic.  I recommend the
default dir should be "." because by definition it exists, although it may
not be writable.  The current code properly traps that condition.   dfile is
more subject to discussion.  We could require it, as now, if storage_mode is
files, but I would suggest an alternative is to use the object_id string
of the document saved to define the file name.  That would seriously
complicate the logic of the save_data method, however, because now he
waveform data are always saved before the metadata.  Hence, I don't think
this is worth implementing unless we think it would be important to some users.
2.  I think there are some subtleties in how you guys implemented the clean
methods wrt aliases my docstrigs are not capturing.  We probably need to
do a mind meld on this topic.
3.  In clean_collection I was very confused at first by the "check_xref"
argument.  I finally realized it was a name confusion.  The name is more
appropriat for a verify method, but after I looked carefully at this i
realized it was defining what xref attributes were linked to be
enforced when delete_missing_xref was set true.   I am changing the symbol
check_xref in clean and clean_collection to "required_xref_list" to prevent
user's from having the same confusion.
4.  Our clean operation for deletes has a serious hole we either need to fix
or document.  That is, deletions only delete the document.  We probably should
at least include a gridfs manager to delete gridfs data linked to such documents
when they are present.   I think the problem is easily solved by calling
delete_data instead of the current algorithm that just calls the MongoDB
delete_one method.  
5.  I have two issues with the verify method.  First, it has an ambiguous return
for error conditions.  It returns an empty dictionary if there are not problems
but returns the same thing if the id passed is not found.  The later is an
error condition and should be flagged.   I would recommend it throw an
exception because a failed find should never happen for that operation and
should generate an exception.  In fact, I'm changing the code to do that. The
alernative is to return something different like None, but that would just
generate a myserious exception if not handled.  Better to use the exception
mechanism.  Second it should probably also return a list of test names as the
value and not just a single string. I think all the current tests are orthogonal
and this isn't necessary, but for extensibility it would be better to return
a list.  I did not make that change though as it should be discussed collectively.

6.  You guys need to read my docstring for update_metadata carefully.  I
may misunderstand all the subtleties of this complicated function.

7. Correct me if I'm wrong but I think we've lost a functionality we had
in the design for handling metadata.  That is, I don't think the save_data
writer the capability of only updating metadata and not saving the waveform
data.  I think update_metdata was supposed to do that but it evolved and
save_data evolved in a way that we lost that capability.  Correct me if I'm
misreading the code.

8.  I modified read_ensemble_data to have an optional ensemble_metadata
argument.  The data in ensemble_metadata are loaded into the ensemele's
metadata container  My mods are totally untested because I am doing this
when I'm off the grid.

9.  I modified save_ensmble_data to always call sync_metadata.  I would
strongly argue for that behavior so I didn't consider it debatable.
It could be made an option that defaults to true, but I would not recommend it.

10.  The user's manual describes a feature that seems to have not been
implemented.   That is, when I wrote that I was under an impression the
reader would accept either a cursor object or a list of ids.  The cursor
option does no currently exist.  I actually think now that is more than ok
BUT it means we should have a convenience function I'll call cursor2idlist.
It is trivial to implement.  I'm going to add it as a procedural function
the database.py file.

11. I think the default for clear_elog in delete_data should be false.  I'm more than ok with
clear_history true by default, but dropping error logs strikes me as asking
for mysterious behavior that cannot be traced.  I changed database.py in
my revisions.

12. The CRUD_ooerations.rst User's manual file has a paragraph on something
I developed called the "staged_file_manager" in that document.   That
feature has not yet been put into the repository.   Read the CRUD_operations
section revision for the reasons I think this should exist.  If you think
otherwise, we will need to revise that section.  If you concur it would be
a useful addition, I can work to integrate it a new branch.  I think the
current form needs some revisions,however, in a number of ways. I think what
that needs to evolve to is a command line tool to efficiently clear data with
a particular "tag".

13.  Both of you should read the CRUD_operations document and check for
inconsistencies.  I tried, but I'm not at all confident all the things noted
in that document have been implemented or were changed between the time I wrote
that and we finished the database api.  At this point we probably should aim
to document what and implement any new things that document suggests unless the
changes are relativly simple (like my examples above of changing defaults or
something that requires a minimal change)

14.  Case in point about a trivial change.   I changed the dead data entry key
from "gravestone" to "tombstone".   Both are valid english words, but in the
dialect I speak, at least, tombstone is by far the most common name used.
That is probably not universally true, but I made this change anyway.
I only changed database.py and the schema yaml file- I don't know if the name
appears elsewhere.

15.  The CRUD_operations document currently has this quote that I am not sure is
accurate:

"The set of which collections are to be loaded are controlled by optional
parameters in each reader.  An important constraint is that for all
normalized collections defined as required, if the cross-referencing
key is not defined a reader will ignore that datum.  :code:`read_data` silently
signals that condition by returning a None.  :code:`read_ensemble_data` and
:code:`read_distributed_data` normally silently skip such data.   That model
is intentional because it allows initial loading of a large data set with
unresolvable anomalies that prevent one or more of the cross-referencing
ids from being defined.""

Can one or both of you confirm or deny this is correct?

16. The final section in the CRUD_operations.rst is incomplete because we
haven't fully closed on the issue of how to do handle a pure metadata update.
Maybe update_metadata does this, but I need confirmation of that before I can
finalize that revision.   I may also need to update the docstring for
update_metadata for the same reason.


pavlis:
I see checks section above exits rather quickly on a configuration error that I'm clueless about. I wonder if this is a historical relic preserved from the fork I made from master?  I'm hoping someone else can solve this as I'm clueless.

wangyinz:
Let's move the discussion offline.

The config error seems to be due to a recent update to pip (see https://github.com/pypa/pip/issues/9880). We need to either disable the caching in github actions or fix the version of pip in the workflow.

wangyinz:
The changes are already in #198, and will be merged there.

PR #192, opened 2021-05-02
Add mutes

pavlis:
This branch adds a new TopMute algorithm.   The need for this came up from converting pwmig where a top mute is an essential algorithm.   I implemented this as a simple variant of the Taper class in C++ created a while ago for deconvolution.   This particular algorithm is a good example of how to connect a C++ class with an apply method to mspass.   I made the pybind11 make _TopMute the name for the C++ class and then build a python class that added history and elog functionality.   

I put the python code in a rational place under mspasspy.algorithms.window.  The C++ pybind11 code puts _TopMute in a weird place for convenience.  The reason is that I made TopMute a subclass of Taper, which happens to be in the seismic section of ccore.   I think that a minor blemish because _TopMute should not be used by user's anyway so a nonintuitive import line is not a bit issue.  If you disagree, it can be fixed but it would be a lot of pain to get all the includes right.

The test program I wrote for this is not appropriate for automated testing, but demonstrates the basic code does work.  the reason it isn't appropriate for an automated test is the validation is graphically.   Since I still don't have a working test platform it is hard for me to design a useful automated test of this algorithm.  Here it is if you want to try it:
```
from mspasspy.ccore.seismic import (TimeSeries,Seismogram)
from mspasspy.graphics import SeismicPlotter
from mspasspy.algorithms.window import TopMute
import matplotlib.pyplot as plt
ts=TimeSeries(100)
seis=Seismogram(100)
#Fill data arrays with all ones to show form of mute
for i in range(100):
    ts.data[i]=1.0
    for k in range(3):
        seis.data[k,i]=1.0
ts.dt=0.1
ts.t0=0.0
seis.dt=0.1
seis.t0=0.0
ts.live=True
seis.live=True
ts2plot=TimeSeries(ts)
seis2plot=Seismogram(seis)
lmute=TopMute(2.0,4.0,'linear')
cmute=TopMute(2.0,4.0,'cosine')

plt.plot(ts.data)
plt.show()
#fig,ax=plt.subplot(3,1)
plotter=SeismicPlotter()
plotter.change_style('wt')
lmute.apply(ts2plot)
plotter.plot(ts2plot)
plt.show()
lmute.apply(seis2plot)
plotter.plot(seis2plot)
plotter.plot(seis2plot)
plt.show()

ts2plot=TimeSeries(ts)
seis2plot=Seismogram(seis)
cmute.apply(ts2plot)
cmute.apply(seis2plot)
plotter.plot(ts2plot)
plt.show()
plotter.plot(seis2plot)
```


pavlis:
It seems this branch hasn't been assimilated into master yet.   It is more than 3 week old.  I think it is stalled for a test suite.   My work with pwstack is stuck until  that gets fixed as TopMutes are required for that algorithm.   I can throw a test function together if you don't have time, although I may not either with my schedule right now.

wangyinz:
Sorry about the delay. I just quickly finished the refactoring that we have discussed. I should be able to throw in a test for the TopMutes here.

PR #198, opened 2021-05-11
fix issues in the database.py

haruming:
This is a new branch checkout from Gary's pull request.

1. Fixed the issues in the database.py based on Gary's pull request and add relevant tests.
2. The last point(16) in the documentation looks confusing to me, it seems like I don't need to revise anything?
3. Looks like there is still an issue with the documentation build in the GitHub push workflow.


haruming:
I add the feature we talked about at the last meeting that we need to log the global history for the read_data and save_data in a workflow. However, there was an issue when I was working on it.

When we call a spark/dask map for a list of mspass_object to save_data. However, after the map operation, the original mspass_objects did not change, which means there's no _id field added in the original mspass objects. I suspects it is because normally we call a map operation, it should create another new object and those in-place operations on the sparkRDD or DaskBag would not reflect the original data.

Thus, to workaround, I also return the mspass object after the save_data operation when we call a map.

haruming:
I finished the refinement of global history and also the object history based on our discussion, but due to the confusion I have in the middle, I think you guys should take a careful look at the changes I made and make sure that follows the design of the Processing History.

pavlis:
I stepped in to the review section, but am stopped by two things.  First, there are so many changes here with background I only caught pieces of (i.e. conversations between you and Ian) I am not sure what I should even focus on.   There are large parts of this with functionality I can't understand without spending a huge amount of time and testing.   Second, I'm a novice at doing github review feedback anyway and I don't know how to even make a strategic comment at the place I have a question.  Consequence, I guess, of being an old fart without a strong vested interesting in learning a new thing like this.   We have a conference call today, maybe we can iron out more exact things you need my help answering.

PR #208, opened 2021-06-02
Improve ensembles

pavlis:
This is a major change to the API for TimeSeriesEnsemble and SeismogramEnsemble.   Changes include:

- C++ templates add an ErrorLogger (symbol elog as in TimeSeries and Seismogram) to the ensemble template.  
- C++ templates add functionality parallel to live() methods in BasicTimeSeries.  That is, an entire ensemble can be killed or tested for validity with live() and dead() methods.  
- A validate() method was added that scans the data vector for any live members.  I all members are marked dead the ensemble will be marked dead. 
- The new features extend the Ensemble template class.  The C++ name of the template class is LoggingEnsemble.  IMPORTANT though is that LoggingEnsemble gets mapped in the pybind11 code to define a TimeSeriesEnsemble as a LoggingEnsemble<TimeSeries> and a SeismogramEnsemble is now a LoggingEnsemble<Seismogram>.   The extensions should not effect our existing code base because parallel constructors make one of these constructed by any of Ensemble api the same but with an empty error log (a totally rational extension).  
- EQUALLY IMPORTANT is that this implementation adds pickle support for all Ensembles.  I'm concerned the implementation is hideously inefficient and absolutely will be a memory pig, but tests show it seems to work. 
- I added test code to test_ccore.py to test the new functionality.   You guys should look at my test code critically and consider adding additional tests.  I may have missed something.


pavlis:
My bad on this guys.  I thought this passed all the test and it did not.   I'll try to push an fix shortly.

pavlis:
Well, maybe not so quickly.   Now it seg faults, which will be harder to track down.  Sorry for the premature pull request.+

python/tests/test_ccore.py
```diff
@@ -685,6 +686,42 @@ def test_Ensemble(Ensemble):
...
+    assert es.live()
+    es.kill()
+    assert es.dead()
+    # resurrect es
+    es.set_live()
+    assert es.live()
+    # validate checks for for any live members - this tests that feature
+    assert es.validate()
+    # need this temporary copy for the next test_
+    escopy=deepcopy(es)
```
wangyinz, line 707: Well, there is nothing to sorry for. I just checked the error log, it seems the only error is at this line where it complains `deepcopy` is not defined. I guess it is not imported above? Also, if you are testing the serialization here, why not call pickle instead?

pavlis: It was MUCH worse than that.   Just pushed a complete fix with some bare bones tests in the test_ccore.py file.   I am not at all thrilled with this implementation, but to again quote that basic programming axiom "make it work before you make it fast".   In this case I'm very worried about both speed and memory use.   The deserialization section is a horrible memory pig and will create potential problems in some uses.   I think we ought to review this code in our conference call later today. 

The checks are still running as I'm writing this, but it passed my tests with pytest.   I have pytest only partly working, however, as I think spark is broken on this machine.

cxx/python/seismic/seismic_py.cc
```diff
@@ -593,6 +593,174 @@ PYBIND11_MODULE(seismic, m) {
...
+          pybind11::object sbuf;
+          sbuf=serialize_metadata_py(self);
+          stringstream sselog;
+          boost::archive::text_oarchive arelog(sselog);
+          arelog << self.elog;
+          pybind11::module pickle = pybind11::module::import("pickle");
+          pybind11::object dumps = pickle.attr("dumps");
+          size_t nmembers=self.member.size();
+          py::list dlist;
+          for(auto i=0;i<nmembers;++i)
```
wangyinz, line 621: Loop here is not necessary. As mentioned in #207, `pickle.dumps(list(self.member))` here should be supported.

pavlis:
Hmmm - this explains why my revisions to the scale function didn't work as I thought they should with ensembles.  Why was this branch not merged?   Does it need more testing?

wangyinz:
No... I was sort of waiting for your approval, but then forgot. I will merge it now

PR #211, opened 2021-06-03
bugfix for minor issues in start-mspass.sh script

haruming:
These are some changes I made in start-mspass.sh script, but there might be more changes on them.

### DIFF
--- a/scripts/start-mspass.sh
+++ b/scripts/start-mspass.sh
@@ -3,8 +3,10 @@
 # If running with docker use /home, else use pwd to store all data and logs
 if grep docker /proc/1/cgroup -qa; then
   MSPASS_WORKDIR=/home
+elif [[ -z ${MSPASS_WORK_DIR} ]]; then
+  MSPASS_WORKDIR=$MSPASS_WORK_DIR
 else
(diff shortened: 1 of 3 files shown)


scripts/start-mspass.sh
```diff
@@ -224,14 +260,18 @@ if [ $# -eq 0 ]; then
...
+    else
+      clean_up_single_node
+    fi
   else # if [ "$MSPASS_ROLE" = "all" ]
     MSPASS_DB_ADDRESS=$HOSTNAME
     MSPASS_SCHEDULER_ADDRESS=$HOSTNAME
     eval $MSPASS_SCHEDULER_CMD
     eval $MSPASS_WORKER_CMD
-    [[ -d $MONGO_DATA ]] || mkdir -p $MONGO_DATA
-    mongod --port $MONGODB_PORT --dbpath $MONGO_DATA --logpath $MONGO_LOG --bind_ip_all &
```
wangyinz, line 233: @haruming Gary ran into a problem that the db was not started when using the container in all-in-one mode. I checked and found that the problem is these lines' removal. Could you please double check and fix? Thank you!

PR #213, opened 2021-06-15
code formatting for python files

haruming:
1, Formatting the code according to PEP8.
2, When looking into the Trace2TimeSeries problem, I am not sure if it is a bug in the Trace2TimeSeries function in util.converter.py from line 247 - 249, but I changed it to what I understand.
```
    # we don't require endtime in TimeSeries so ignore if it is not set
    if 'endtime' in trace.stats:
        t = trace_stats_dict['endtime']
        trace_stats_dict['endtime'] = t.timestamp
```

### DIFF
--- a/python/mspasspy/algorithms/RFdeconProcessor.py
+++ b/python/mspasspy/algorithms/RFdeconProcessor.py
@@ -91,13 +91,17 @@ class RFdeconProcessor:
 
         :param d: input data (contents expected depend upon 
         value of dtype parameter).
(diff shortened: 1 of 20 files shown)


seisman:
Have you considered formatting the codes using [black](https://black.readthedocs.io/en/stable/). I think it's widely used for Python projects.

wangyinz:
yes, I might integrate that into our CI here, but this branch is not only for formatting issues that can be handled by a formatter. We need to manually clean up the code and then apply the formatter.

PR #214, opened 2021-06-20
bugfix for update metadata

haruming:
When trying to save/update a mspass_object, if we could not find the wf record in the database according to the _id field associated with the mspass_object. We should either continue to save the metadata or raise an error. The rules are:
1. If we only update the mspass_object, we could continue to update only if it is under promiscuous mode
2. If we gonna save the mspass_object, we would treat it as insertion and remove its _id field. Semantically, when we want to save mspass_object with a meaningless _id field, we focus on the metadata associated with the object, thus we could treat it as an insertion.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -578,7 +578,7 @@ class Database(pymongo.database.Database):
         update_res_code = -1
         if mspass_object.live:
             # 1. save metadata, with update mode
-            update_res_code = self.update_metadata(mspass_object, mode, exclude_keys, collection, True, data_tag, alg_name, alg_id)
+            update_res_code = self.update_metadata(mspass_object, mode, exclude_keys, collection, True, data_tag, alg_name, alg_id, True)
 
             if mspass_object.live:
                 # 2. save actual data in file/gridfs mode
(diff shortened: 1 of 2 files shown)


python/mspasspy/db/database.py
```diff
@@ -1519,7 +1519,7 @@ def _check_required(self, collection='site', keys=['lat','lon','elev','starttime
         return tuple([wrong_types,undef])
 
     def update_metadata(self, mspass_object, mode='promiscuous', exclude_keys=None, collection=None, ignore_metadata_changed_test=False,
-                        data_tag=None, alg_name='update_data', alg_id='0'):
+                        data_tag=None, alg_name='update_data', alg_id='0', is_save_data=False):
```
wangyinz, line 1522: Well... because this method is public to the users, I don't think it is a good idea to include options that are for internal use mostly. Well, since we are at this now, I guess it may not be a bad idea to change the structure a little bit and create a private `_update_metadata` method that is called by the `update_metadata` and `save_data` methods. This way, the code should look cleaner, too. What do you think?

haruming: Yeah, I had doubts when I implemented this method, now it seems like what you suggest is a good idea.

python/mspasspy/db/database.py
```diff
@@ -1518,9 +1518,34 @@ def _check_required(self, collection='site', keys=['lat','lon','elev','starttime
                 break
         return tuple([wrong_types,undef])
 
-    def update_metadata(self, mspass_object, mode='promiscuous', exclude_keys=None, collection=None, ignore_metadata_changed_test=False,
+    def update_data(self, mspass_object, mode='promiscuous', exclude_keys=None, collection=None, ignore_metadata_changed_test=False,
```
wangyinz, line 1521: maybe we should still call it update_metadata?

PR #217, opened 2021-07-01
Remove the sentence about DockerHub account

seisman:
As I understand it, DockerHub account is not required to use Docker and MsPASS. The following sentence may give users wrong impression that a docker hub account is necessary. I think it should be removed from the "Get Started" section to lower the barrier for new users.

> You may want to create an account and sign in your Docker following this instruction.


wangyinz:
Thanks for the suggestion. I think that make sense.

PR #220, opened 2021-07-03
Bugfix db ensemble data

haruming:
1. fix bugs for not saving tmatrix and two boolean variables properly in metadata before actually saving to database
2. For the constructor in TimeSeires, I remove the for loop and change it for memcpy for better performance
3. convert the cursor to a list to fix bugs for getting the length of the records correctly when passing MongoDB cursor in read_ensemble_data
4. update the format of yaml files to support the new features.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -1793,19 +1793,18 @@ class Database(pymongo.database.Database):
                 object_type, wf_collection), 'Fatal')
 
         # if objectid_list is a cursor, convert the cursor to a list
-        member_list = []
-        for i in objectid_list:
-            member_list.append(i)
+        if type(objectid_list) is pymongo.cursor.Cursor:
+            objectid_list = list(objectid_list)
 
         if object_type is TimeSeries:
-            ensemble = TimeSeriesEnsemble(len(member_list))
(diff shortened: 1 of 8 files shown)


python/mspasspy/db/database.py
```diff
@@ -1793,9 +1793,9 @@ def read_ensemble_data(self, objectid_list, ensemble_metadata={},
                 object_type, wf_collection), 'Fatal')
 
         if object_type is TimeSeries:
-            ensemble = TimeSeriesEnsemble(len(objectid_list))
+            ensemble = TimeSeriesEnsemble(len(list(objectid_list)))
         else:
-            ensemble = SeismogramEnsemble(len(objectid_list))
+            ensemble = SeismogramEnsemble(len(list(objectid_list)))
```
wangyinz, line 1798: why adding the conversion to list here? @haruming  I guess it is an attempt to handle cursor, but the conversion here will consume the whole cursor, making it invalid to be referenced from in the next loop. Also, it seems this is not tested with a cursor, otherwise it should reveal the problem.

wangyinz (changes requested):
Thanks for the quick fix! I only found one issue and it should be straightforward to fix.

PR #222, opened 2021-07-05
Add configurations to show the 'Edit on GitHub' button

seisman:
Currently, each page on the documentation shows a "View page source" link, which links to the plain text source file:
image

It's more useful to link directly to the source files in the GitHub repository, because the source files are rendered well on GitHub and are easier to read. It's also much easier for users to submit quick fixes for typos.

image

The `html_context` is documented in https://github.com/readthedocs/sphinx_rtd_theme/issues/465.

It works well in my other projects, so I expect it should work well for mspass too. However, I can't test it because for me, it's difficult to build mspass from source codes locally.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -21,7 +21,7 @@ import sphinx_rtd_theme
 # -- Project information -----------------------------------------------------
 
 project = 'MsPASS'
-copyright = '2020, Ian Wang'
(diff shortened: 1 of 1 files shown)


wangyinz:
Great! This is what I have thought about but did not have the chance to try it out. Thanks for contributing!

PR #227, opened 2021-07-14
Fix readonly problem

pavlis:
This branch has a complete rewrite of the save_data method and update_metadata method.  It removes the private _update_metadata method completely.   save_data now is the union of the old _update_metadata and the old save data.   The new update_metadata method behaves very differently from the old one.   See doctrings for details. 

This first push is known to be broken.   I am getting weird results locally with pytest that i am pretty sure are related to the fact my development ubuntu machine has an incomplete test environment.   This push should not be taken seriously until I can view the test run results and try to fix what I can.   @haruming  I will need your help resolving some of the test conflicts and revising the test script to mesh with the changes to save_data and update_metadata.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -7,20 +7,17 @@ import copy
 import pathlib
 import pickle
 import struct
-import sys
 import urllib.request
 from array import array
 
 import dask.bag as daskbag
 import gridfs
 import pymongo
-from bson.objectid import ObjectId
 import numpy as np
 import obspy
 from obspy import Inventory
 from obspy import UTCDateTime
 
-import mspasspy.util.converter
 from mspasspy.ccore.io import _mseed_file_indexer
 
 from mspasspy.ccore.seismic import (TimeSeries,
@@ -89,7 +86,7 @@ def read_distributed_data(db, cursor, mode='promiscuous', normalize=None, load_h
       manages schduling.  See online Spark documentation for details on
       this concept.
     :type spark_context: :class:`pyspark.SparkContext`
-    :param npartitions: The number of desired partitions for Dask or the number 
+    :param npartitions: The number of desired partitions for Dask or the number
       of slices for Spark. By default Dask will use 100 and Spark will determine
       it automatically based on the cluster.
     :type npartitions: :class:`int`
@@ -173,6 +170,9 @@ class Database(pymongo.database.Database):
         return ret
 
     def __setstate__(self, data):
+        # somewhat weird that this import is requiired here but it won't
+        # work without it.  Not sure how the symbol MongoClient is required
+        # here but it is - ignore if a lint like ide says MongoClient is not used
         from pymongo import MongoClient
         data['_Database__client'] = eval(data['_Database__client'])
         self.__dict__.update(data)
@@ -395,7 +395,9 @@ class Database(pymongo.database.Database):
                 md[k] = b'\x00'
             else:
                 md[k] = None
-
+        # This is needed or readonly attributes will always get flagged on
+        # a subsequent save
+        md.clear_modified()
         if object_type is TimeSeries:
             # FIXME: This is awkward. Need to revisit when we have proper constructors.
             mspass_object = TimeSeries({k: md[k] for k in md}, np.ndarray([0], dtype=np.float64))
@@ -438,8 +440,10 @@ class Database(pymongo.database.Database):
 
         return mspass_object
 
-    def save_data(self, mspass_object, mode="promiscuous", storage_mode='gridfs', dir=None, dfile=None, format=None,
-                exclude_keys=None, collection=None, data_tag=None, alg_name='save_data', alg_id='0'):
+    def save_data(self, mspass_object, mode="promiscuous",
+            storage_mode='gridfs', dir=None, dfile=None, format=None, overwrite=False,
+                exclude_keys=None, collection=None, data_tag=None,
+                  alg_name='save_data', alg_id='0'):
         """
         Use this method to save an atomic data object (TimeSeries or Seismogram)
         to be managed with MongoDB.  The Metadata are stored as documents in
@@ -449,7 +453,8 @@ class Database(pymongo.database.Database):
         performance of file system io versus gridfs, but our working hypothesis
         is the answer of which is faster will be configuration dependent. In
         either case the goal of this function is to make a save operation as
-        simple as possible.
+        simple as possible by abstracting the complications involved in
+        the actual save.
 
         Any errors messages held in the object being saved are always
         written to documents in MongoDB is a special collection defined in
@@ -507,8 +512,18 @@ class Database(pymongo.database.Database):
           assumes means to store the data in its raw binary form.  The default
           should normally be used for efficiency.  Alternate formats are
           primarily a simple export mechanism.  See the User's manual for
-          more details on data export.
+          more details on data export.  Used only for "file" storage mode.
         :type format: :class:`str`
+        :param overwrite:  If true gridfs data linked to the original
+          waveform will be replaced by the sample data from this save.
+          Default is false, and should be the normal use.  This option
+          should never be used after a reduce operator as the parents
+          are not tracked and the space advantage is likely minimal for
+          the confusion it would cause.   This is most useful for light, stable
+          preprocessing with a set of map operators to regularize a data
+          set before more extensive processing.  It can only be used when
+          storage_mode is set to gridfs.
+        :type overwrite:  boolean
         :param exclude_keys: Metadata can often become contaminated with
           attributes that are no longer needed or a mismatch with the data.
           A type example is the bundle algorithm takes three TimeSeries
@@ -532,6 +547,18 @@ class Database(pymongo.database.Database):
         :param data_tag: a user specified "data_tag" key.  See above and
           User's manual for guidance on how the use of this option.
         :type data_tag: :class:`str`
+        :return: list of 3 ObjectIDs.  0=waveform id, 1=history collection id,
+          2 = error log (elog) collection id.   Any or all of the ids may
+          be returned as None (python null).  For a valid save of live data
+          0 will be the id for the wf collection.  It is None only if there
+          was no '_id' value set in the data object received this method.
+          The history id (list component 1) is always null unless the
+          history container is found to not be empty.  Similarly, the
+          error log id (component 2) will be None unless there were errors
+          posted to the waveform just save.  The elog entry should then have
+          a valid link back to the parent.  Note dead data will have a
+          "tombstone" entry in elog with a dump of the metadata they contained
+          linked with the error log data.
         """
         if not isinstance(mspass_object, (TimeSeries, Seismogram)):
             raise TypeError("only TimeSeries and Seismogram are supported")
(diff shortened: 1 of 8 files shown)


pavlis:
The first round with this confirmed the weird behavior I was seeing was from my incomplete test environment.  Fixed a few obvious problems from the github error log and did a major edit to the docstring of save_data and update_metadata.  @haruming (you too @wangyinz when you, literally, land and get settled form your trip) should have a look at the docstring and provide a critical review of the behavior I coded into these routines.

pavlis:
I misread the output from the first run.   There is a lot messed up here with the tests. We are going to need to discuss how to correct this in our weekly meeting tomorrow.   I'm going to try to see how this works independently for now with notebooks I'm working on

python/mspasspy/db/database.py
```diff
@@ -574,45 +645,182 @@ def save_data(self, mspass_object, mode="promiscuous", storage_mode='gridfs', di
...
+                for k in copied_metadata:
+                    if save_schema.is_defined(k):
+                        if not isinstance(copied_metadata[k], save_schema.type(k)):
+                            insertion_dict[k]=copied_metadata[k]
+                        else:
+                            if mode == 'pedantic':
+                                mspass_object.kill()
+                                message ='pedantic mode error:  key='+k
+                                if not save_schema.is_defined(k):
+                                    message += ' not defined in schema.'
```
haruming, line 711: This condition checking is redundant because, at line 703, it's always guaranteed that the key is defined.

python/mspasspy/db/database.py
```diff
@@ -574,45 +645,182 @@ def save_data(self, mspass_object, mode="promiscuous", storage_mode='gridfs', di
...
+                insertion_dict = dict(copied_metadata)
+            else:
+                # Other modes have to test every key and type of value
+                # before continuing.  pedantic kills data with any problems
+                # Cautious tries to fix the problem first
+                # Note many errors can be posted - one for each problem key-value pair
+                for k in copied_metadata:
+                    if save_schema.is_defined(k):
+                        if not isinstance(copied_metadata[k], save_schema.type(k)):
+                            insertion_dict[k]=copied_metadata[k]
```
haruming, line 705: This should be the reverse, right? If it matches the type in the metadata schema, we should ignore the checking under different modes.

python/mspasspy/db/database.py
```diff
@@ -574,45 +645,182 @@ def save_data(self, mspass_object, mode="promiscuous", storage_mode='gridfs', di
...
+                                mspass_object.elog.log_error('Database.save_data',
+                                    'message',ErrorSeverity.Invalid)
+                            else:
+                                # Careful if another mode is added here.  else means cautious in this logic
+                                 try:
+                                     # The following convert the actual value in a dict to a required type.
+                                     # This is because the return of type() is the class reference.
+                                     insertion_dict[k] = save_schema.type(k)(copied_metadata[k])
+                                 except Exception as err:
+                                     mspass_object.kill()
```
haruming, line 724: In our previous implementation, under cautious mode, we would only kill objects when we are unable to convert the mismatch required attribute. For those normal attributes, we only post a complaint elog entry to the mspass_object. Just to double confirm that the current change is what we expect right?

python/mspasspy/db/database.py
```diff
@@ -574,45 +645,182 @@ def save_data(self, mspass_object, mode="promiscuous", storage_mode='gridfs', di
...
+                # this case or a the old tag will be saved with this datum
+                if 'data_tag' in insertion_dict:
+                    insertion_dict.erase('data_tag')
+            # finally ready to insert the wf doc - keep the id as we'll need
+            # it for tagging any elog entries
+            wfid = wf_collection.insert_one(insertion_dict).inserted_id
+            # Put wfid into the object's meta as the new definition of 
+            # the parent of this waveform
+            # TODO - I think we need to reset the history container here 
+            # and mark this new save as an ORIGIN leaving till later
```
haruming, line 783: This is done by the _save_history method in database.py.

python/mspasspy/db/database.py
```diff
@@ -574,45 +645,182 @@ def save_data(self, mspass_object, mode="promiscuous", storage_mode='gridfs', di
...
+            mspass_object['_id']=wfid
+
+            # Empty error logs are skipped.  When nonzero tag them with tid
+            # just returned
+            if mspass_object.elog.size() > 0:
+                elog_id_name = self.database_schema.default_name('elog') + '_id'
```
haruming, line 789: This line is redundant.

python/mspasspy/db/database.py
```diff
@@ -574,45 +645,182 @@ def save_data(self, mspass_object, mode="promiscuous", storage_mode='gridfs', di
...
+            if elog_id_name in mspass_object:
+                old_elog_id=mspass_object[elog_id_name]
+            else:
+                old_elog_id=None
+            elog_id = self._save_elog(mspass_object,elog_id=old_elog_id)
+            if '_id' in mspass_object:
+                dead_wf_id=mspass_object['_id']
+            else:
+                dead_wf_id=None
+            return [dead_wf_id,None,elog_id]
```
haruming, line 823: I think save_data is much cleaner than before and it would work well when we save a mspass_object for the first time. However,  I could think of a scenario that a user may call save_data for a mspass_object for many times. In that case, some of the logic in the code is not applicable because we assume when the user call save_data, it's always a new wf document. Also, there would be a lot of wf documents associated with this mspass_object in the MongoDB database. I am not sure if these would cause some problems during our processing workflow and I think I would let you guys for discussions.

pavlis:
Thanks for the review @haruming .   I was doing some parallel things I checked in just now and hadn't noticed your comments.   I'll go back and work them when I get a chance.  

I'm running into a strange thing that is, I think, also causing problems in pytest runs.   I have examples where stepping through the code I can step through a call to insert_one, insert_one returns a normal looking object id, but find_one on that same id after fails.  it is like insert_one is doing nothing.   Only happens inside Database class tests. i.e. I can't recreate the problem with simple externally done ipython lines.   I'm a bit mystified.  Maybe things will be clearer in the morning

python/mspasspy/db/database.py
```diff
@@ -1518,229 +1724,188 @@ def _check_required(self, collection='site', keys=['lat','lon','elev','starttime
...
+            if k in copied_metadata:
+                copied_metadata.erase(k)
+        # Now remove any readonly data
+        for k in copied_metadata.keys():
+            if save_schema.is_defined(k):
+                if save_schema.readonly(k):
+                    if k in changed_key_list:
+                        newkey='READONLYERROR_'+k
+                        copied_metadata.change_key(k,newkey)
+                        mspass_object.elog.log_error('Database.save_data',
```
haruming, line 1851: minor issue here, the elog message should be Database.update_metadata

python/mspasspy/db/database.py
```diff
@@ -1518,229 +1724,188 @@ def _check_required(self, collection='site', keys=['lat','lon','elev','starttime
...
+                                    message,ErrorSeverity.Complaint)
+                            # Note we land here for both pedantic and cautious but not promiscuous
                             try:
                                 # The following convert the actual value in a dict to a required type.
                                 # This is because the return of type() is the class reference.
-                                insert_dict[k] = update_metadata_def.type(k)(copied_metadata[k])
+                                oldvalue=copied_metadata[k]
+                                insertion_dict[k] = save_schema.type(k)(copied_metadata[k])
+                                newvalue=copied_metadata[k]
+                                message='Had to convert type of data with key='+k+' from '+str(type(oldvalue))+' to '+str(type(newvalue))
```
haruming, line 1894: newvalue here is the same as oldvalue because, in Python, the type cast would not change the original variable.

python/mspasspy/db/database.py
```diff
@@ -574,45 +645,182 @@ def save_data(self, mspass_object, mode="promiscuous", storage_mode='gridfs', di
...
+                                     # The following convert the actual value in a dict to a required type.
+                                     # This is because the return of type() is the class reference.
+                                     insertion_dict[k] = save_schema.type(k)(copied_metadata[k])
+                                 except Exception as err:
+                                     mspass_object.kill()
+                                     message = 'cautious mdoe error:  key=' + k
+                                     if save_schema.is_required(k):
+                                        message += ' Required key value could not be converted to required type='+str(save_schema.type(k))+' actual type='+str(type(copied_metadata[k]))
+                                     else:
+                                        message += ' Value has type='+str(type(copied_metadata[k])+' which cannot be converted to required type='+str(save_schema.type(k)))
```
haruming, line 736: minor issue here in the message, in the else condition, it should remove the word 'required'.

python/mspasspy/db/database.py
```diff
@@ -1518,229 +1724,188 @@ def _check_required(self, collection='site', keys=['lat','lon','elev','starttime
...
+                                else:
+                                    message += ' Value stored has type='+str(type(copied_metadata[k])+' which cannot be converted to required type='+str(save_schema.type(k)))+'\n'
+                                message += 'Data for this key will not be changed or set in the database'
+                                message += '\nPython error exception message caught:\n'
+                                message += str(err)
+                                mspass_object.elog.log_error(alg_name,
+                                           message,ErrorSeverity.Complaint)
+                                # The None arg2 cause pop to not throw
+                                # an exception if k isn't defined - a bit ambigous in the try block
+                                insertion_dict.pop(k,None)
```
haruming, line 1909: We probably don't need to include this line here though it doesn't hurt because when a type conversion is unsuccessful, the key-value pair would not be added into insertion_dict.

python/mspasspy/db/database.py
haruming: I found the behavior in update_metadata is different from here. I think we need to agree on what's the correct behaviors under pedantic and cautious mode in tomorrow's meeting.

python/mspasspy/db/database.py
pavlis: Good catch.  Absolutely right!

python/mspasspy/db/database.py
pavlis: That is a good point and indeed one worth discussing.   For our discussion I confirm my intention was:
1.  save_data behaves as before in cautious mode.   That is, it attempts to fix any type mismatch errors and only considers and unrecoverable type mismatch as very bad. Acton on 'very bad' is different here though (I think).   Now the result is a kill of mspass_error instead of the older approach of throwing an exception.   
2. The change I made for update_metadata was intentional.   An "update", by definition is different from "create" in the standard CRUD idea of database transactions.  In an update, of course, the record has to already exist and you are adding to it or replacing one or more name-value pairs.  In fact, writing this led me to make a new change I'll describe in a moment.  My decision, which I probably should have indeed discussed when I was implementing this, was to treat it differently because an update is a different thing.   In an update you don't want to destroy the original record by posting something demonstrably wrong.    In pedantic mode the logic is supposed to refuse to update any data with a schema mismatch.  Cautious and pedantic are handling type mismatches the same - they try to correct type mismatches, alway post an elog message about success of failure, and remove the key-value pair from the update record if the type cannot be converted.   The only difference between 'cautious' and 'pedantic' is in handling keys not defined in the schema.  'pedantic' drops them from the update while 'cautious' adds that data anyway.  both cases cause a message to be posted to elog but with different messages for the different actions. 

In responding I changed the method signature to this:
```
def update_metadata(self, mspass_object, collection=None, mode='cautious',
                exclude_keys=None, force_keys=None, alg_name='Database.update_metadata'):
```
with this explanation of the new force_keys argument in the docstring:
```
       :param force_keys: a list of metadata attributes you want to force
         to be updated.   Normally this method will only update attributes
         that have been marked as changed since creation of the parent data
         object.  If data with these keys is found in the mspass_object they
         will be added to the update record.
```
Definitely this who issue needs to be on our agenda today

python/mspasspy/db/database.py
pavlis: Ok, I removed that comment then.  Did not see that - thanks

python/mspasspy/db/database.py
pavlis: ps:  I had the same inverted logic in update_metadata because the code was nearly identical

python/mspasspy/db/database.py
pavlis: I don't see that?   Are you talking about the test for elog.size() or setting name?   I don't see that either is redundant?

More important is something we should discuss today.   I added these lines immediately above this from testing yesterday:
```
            # We don't want an elog_id in the insertion at this point.
            # A option to consider is if we need an updata after _save_elog
            # section below to post elog_id back.
            insertion_dict.pop('elog_id',None)
```
There are two issues that should be discussed:
1.   how does _save_elog handle saving error logs that are were saved previously?   To me every save should have an independent dump of the error log, but there are comments in the code that suggest you guys were trying to avoid duplicates.   I think that is a problem for any data passed through a reduce operator.  
2. As the comment says, does the waveform record need a reference to the object_id of errors linked to it?   This is a database design issue that relates to something we really need to develop soon.   How does the user get at the error log?  Whether we need this cross link to be saved immediately is debatable.   Efficiency depends upon what our error log report generator should do?  We have not discussed this issue and it is fundamental since the model we are using is kill/log error an no exceptions
3.

python/mspasspy/db/database.py
pavlis: Definitely worth a discussion. To complex to answer coherently here at this time.

python/mspasspy/db/database.py
pavlis: Good catch - cut and paste error

python/mspasspy/db/database.py
pavlis: I need your help on that one.   I'm going to check these changes in when I am finished going through your comments.  Could you correct this and teach me what you did?

python/mspasspy/db/database.py
pavlis: Agreed, done

python/mspasspy/db/database.py
pavlis: My thinking was it was better to be safe. As the comment says any exception in the try block would land there so I thought it better to cover the possibility it had been set somehow before.  That may be being overly paranoid.

pavlis:
This wasn't the original problem that got me here, but I found an issue that needs a solution.   Modifying our notebook for the short course I was trying to create box for reading input data form miniseed files and saving them as TimeSeries data.  Here is the simple script to do that:
```
cursor=db.wf_miniseed.find({})
for doc in cursor:
    d=db.read_data(doc,collection='wf_miniseed')
    retids=db.save_data(d)
```
Problem is that in a debugger when I look inside save_data for one of these object I see this as the contents of the "insertion_dict" variable just before it enters the section to run insert_one on wf_TimeSeries and save the waveform data
```
{'time_standard': 'UTC', 'format': 'mseed', 'dir': '/home/pavlis/tutorials/mspass_tutorial/sage-gage_2021', 'dfile': './wf/shortcourse_1107.mseed', 'sampling_rate': 40.0, 'nbytes': 49152, 'delta': 0.025, 'starttime': 1302187287.1249979, 'npts': 57971, 'last_packet_time': 1302188736.399998, 'utc_convertible': True, 'foff': 0, 'storage_mode': 'file'}
```
HOWEVER, the default for storage is "gridfs" and the code dutifully saves the data as gridfs, BUT leaves the stuff like dir, dfile, etc. in the document created in a call to insert_one(insertion_dict).  At best that is confusing, but more likely it will break things downstream.   We can hard code a fix, but is there a more general solution?

python/mspasspy/db/database.py
haruming: Sorry not being specific, I mean the variable `elog_id_name` because it's declared but not used.

python/mspasspy/db/database.py
haruming: To fix it, we just need to assign `insertion_dict[k]` to `newvalue` because `insertion_dict[k]` is now converted. but `copied_metadata[k]` is still the `oldvalue` one.
That is `newvalue=insertion_dict[k]`.
BTW, we should probably rename the `newvalue` to `new_value`, but that's only my coding style:)

python/mspasspy/db/database.py
haruming: Yeah, sure it would be safe if we do that. I am fine with that.

haruming:
> Thanks for the review @haruming . I was doing some parallel things I checked in just now and hadn't noticed your comments. I'll go back and work them when I get a chance.
> 
> I'm running into a strange thing that is, I think, also causing problems in pytest runs. I have examples where stepping through the code I can step through a call to insert_one, insert_one returns a normal looking object id, but find_one on that same id after fails. it is like insert_one is doing nothing. Only happens inside Database class tests. i.e. I can't recreate the problem with simple externally done ipython lines. I'm a bit mystified. Maybe things will be clearer in the morning

Unfortunately, I also ran into this problem when I ran locally just now. It's really something mysterious that happened in the code. Before returning in the save_data, it does insert a document in the database, but when I call find_one in the test file, it's empty. I had difficulty debugging into it, it seems all correct to me, but I will try my best...

pavlis:
That is an example of the old say "good news and bad news".  Good news you could recreate the, bad news that we have the problem.   Let's both keep chipping away at it and maybe one of us will have a "dohh" moment

python/mspasspy/db/database.py
```diff
@@ -574,45 +642,198 @@ def save_data(self, mspass_object, mode="promiscuous", storage_mode='gridfs', di
...
+                                        message += ' Value has type='+str(type(copied_metadata[k])+' which cannot be converted to type='+str(save_schema.type(k)))
+                                     message += '\nPython error exception message caught:\n'
+                                     message += str(err)
+                                     mspass_object.elog.log_error('Database.save',
+                                           message,ErrorSeverity.Invalid)
+
+        # Note we jump here immediately if mspass_object was marked dead
+        # on entry.  Data can, however, be killed in metadata section
+        # above so we need repeat the test for live
+        wf_collection=self.db[wf_collection_name]
```
haruming, line 745: After comparing the code in the master branch, I found the issue is here. Here we should not assign `self.db[wf_collection_name]` to the `wf_collection `. Instead, we should use `self[wf_collection_name]` because self itself is a Database object already! That's why we should directly access the collection name.

haruming: I found update_metadata also has this problem. Anyway, now it works. But I am a bit confused why we could insert a document into `self.db[wf_collection_name]` because we didn't declare what self.db is.

pavlis: Bravo!   As I predicted it was a stupid little syntax error that was hard to spot.   I concur that I do not understand why the interpreter didn't flag that as an error, or why it didn't generate a runtime error.   So glad you found that.   

Can you push your changes?   Then I'll try to implement some of the changes we discussed yesterday in database.py.   Perhaps you can focus on modifying the test code to deal with the changes.  The biggest issue will be that most of the tests for save_data check the return.  If we convert that to returning the mspass_object all of them will have to change.

pavlis: PS.   Before you push your changes you might quickly change save_data and update_metadata to return the mspass_object.  That way you can be running the tests that use that change while I do the internal changes we discussed yesterday.   Changing the return should be pretty trivial.

haruming: Sure, I will do it today.

haruming:
Sorry, it took more time to fix those test cases and bugs than expected. There are a few obvious bugs in the save_data and update_metadata and those are easy to fix. However, there are some changes or concerns that need to be discussed here.

1.  In the update_metadata, what if we could not find the corresponding document with the _id field in the mspass_object. In pymongo, the default behavior for update_one, in this case, would ignore this update unless we set the `upsert` parameter as True.
2. In the update_metadata and under the promiscuous mode, should we update the read-only attributes. Because if you take a close look into the code, suppose there is a read-only attribute k and we change it to READONLYERROR_k in the `copied_metadata`. Then if the mode is promiscuous, the READONLYERROR_k would never be appended to the `insertion_dict`
3. In the _sync_metadata_before_update, we always set `utc_convertible` in the metadata, however, this attribute is not in the schema. Therefore, we will always post an elog entry complaining about this attribute.
4. I add some logic to remove any attributes with a value that only contains whitespace. In this case, we should erase this key as we did previously.
5.  I add some code dealing with elog in save_data before any discussion. But I do think we need this otherwise test will fail. That is we need to save the elog and get the elog_id before inserting the `insertion_dict` into wf_collection, so that we could associate the elog_id with the wf record and also post the wf_id to the elog entry for cross-reference.

For more detail, you could pull the current changes from this branch and I think things will be clearer when you look at the code.

pavlis:
Very good.  You did a lot more than I thought you were going to so no reason to claim it took a while.  You did a lot.  Let me respond to each of your points.  I'll number my responses the same as your points:
1.  I think what you did is the right approach - throw and exception in that case.   It is a usage error to try to do an update without the id field - at least in my opinion.  The kill mechanism seems inappropriate for an update.
2.  I think the logic handles the situation you note correctly.    This section precedes the test for mode and I am pretty sure it handles the situation you note:
```
        for k in copied_metadata.keys():
            if k == '_id':
                continue
            if save_schema.is_defined(k):
                if save_schema.readonly(k):
                    if k in changed_key_list:
                        newkey='READONLYERROR_'+k
                        copied_metadata.change_key(k,newkey)
                        mspass_object.elog.log_error('Database.update_metadata',
                                'readonly attribute with key='+k+' was improperly modified.  Saved changed value with key='+newkey,
                                ErrorSeverity.Complaint)
                    else:
                        copied_metadata.erase(k)
```
3.  That is an error in the schema, not the code right?  However, I see it defined in the schema.   Did you fix it already?  If not, maybe on your development copy you forgot to install the latest schema.  I've gotten bit on that multiple times before, including this morning.
4. Good
5. I was going to do that but seems you did that already.  We actually discussed that on Thursday and agreed it was needed.  Thanks for taking care of that.

I see you also took care of changing the code for both methods under discussion to return the mspass_object.  The docstrings need an update too, but I'll take care of that. 

Next iteration will have at least two additional this:
1.  I'm adding update_data as we discussed Thursday.
2. I just found and fixed a bug in the bundle code and a related bug in db.save_inventory.   You will see these when I do my next push.   They will precede the commit for adding update_data

haruming:
Sounds good for the next iteration.

Sorry, I didn't make it clear about some points(number 1 and number 3) and it's weird that I am not able to reply to your comment so I leave a new comment here.

1. You are right if we try to update a mspass_object with an `_id` field, we would throw an exception. However, if the mspass_object does have an `_id` field, but we can't find the document in the wf_collection by `find_one` method in pymongo. Should we throw an exception or just ignore this update?
3. Yes the error is in the schema and you are right the `utc_convertiable` attribute is defined in the Database schema, to be specific, in `wf_TimesSeries` and `wf_Seismogram`. But we are using the schema in the Metadata section, in Metadata schema, this attribute is not defined. I checked out the latest version and don't see it is defined in the Metadata.TimeSeries. Since we are using metadata schema handling the logic, we would always post a complaint elog entry about this attribute.

pavlis:
Ah, I see what you mean now.  For these:
1.  I put in a handler for this case.  @wangyinz and @haruming should both confirm the decision I made for this case is the right one;  I posted an elog message and called kill on the datum before returning it (did not save the elog which might be a mistake too).  The issue is do we do that or throw and Exception?  This cannot be silently ignored.  I guess an related issue is if a kill is too brutal?
2. I changed the mspass.yaml file to add this attribute.  @wangyinz should check that I did that right.   

A few residual items here:
1.  I did not extend the test script for Database.  Since I don't have a working test environment I have to punt that to someone else.  I wold be surprised if the update_data method worked without any changes.  github tests are in progress but they shouldn't fail because none of them currently test this new method.
2. I think we might consider creating a private method for handling data killed during a save or update.  We have repetitious code for handling that for which maintenance would be easier if we had a common handler.   I didn't do so because I wasn't sure if it was prudent.  I'm going to leave it to you guys to decide how to proceed on that issue.
3. I had a note from Thursday about a different logic for handling "required" attributes in update_metadata.  I'm afraid my notes are too cryptic to recreate what was being discussed. It was you @wangyinz who suggested we needed a different handler for errors in that situation so perhaps you could look at this version and suggest a fix?   

I would like to pass the ball for finalizing this set of changes to you guys.   I need to focus on finalizing the material for the short course that is less then 3 week from now.

haruming:
Thanks for your efforts Gary, I would take care of the rest and discuss those points with Ian.

PR #228, opened 2021-07-25
Add basic algorithm

wangyinz:
This PR implements the wrapper discussed in #223


wangyinz:
@pavlis I think this is ready to merge, but please review it and let me know if there is anything missing.

pavlis:
I suggest three changes, both of which should take only a few minutes:
1.  It would seem prudent to do a type check that the data type is Seismogram for rotate_to_standard, the rotate methods, and the  transform method.  I think the handler should try to do a kill on the data received but wrap the kill in an exception handler in case what was received is something completely wrong created by some upstream error.  In dask/spark that would seem to be an easy thing to happen in complex error situations.  A robustness thing that seems appropriate here.
2. The methods themselves need to be inside a try: except: block if they can possibly throw an exception. ator does not seem to ever thrown an exception,  rtoa does, rotate_to_standard  and free_surface_transformation do, but the plain rotate methods seem to never throw.  (i.e. rtoa, rotate_to_standard, and free_surface_transformation need an error handler to post the returned error to elog and kill data before returning.)
3. If the input is marked dead return immediately - no reason to depend on the C code to handle that condition

pavlis:
PS to previous:  Maybe the decorator you are using already has an error handler if the function throws an exception?  I vaguely remember it might.  In any case a test for error handling in the test suite seems appropriate.

wangyinz:
Yep, you are exactly right. All those things should be handled in the `mspass_func_wrapper`.  

I guess the type checker is tricky. It right now does not distinguish the four types it's checking (Seismogram, TimeSeries, SeismogramEnsemble, TimeSeriesEnsemble). We might want to revise `mspass_func_wrapper` to make it possible checking a specific subset of them. However, I think this is not that important right now, as a type error will be captured in the try block later.

Well, I just found that the `is_input_dead` checking is missing in the `mspass_func_wrapper` - it is called in the other wrappers that converts obspy objects. That should be easy to fix anyway.

wangyinz:
Yep, you are exactly right. All those things should be handled in the `mspass_func_wrapper`.  

I guess the type checker is tricky. It right now does not distinguish the four types it's checking (Seismogram, TimeSeries, SeismogramEnsemble, TimeSeriesEnsemble). We might want to revise `mspass_func_wrapper` to make it possible checking a specific subset of them. However, I think this is not that important right now, as a type error will be captured in the try block later.

Well, I just found that the `is_input_dead` checking is missing in the `mspass_func_wrapper` - it is called in the other wrappers that converts obspy objects. That should be easy to fix anyway.

PR #230, opened 2021-07-27
Add missing information for data_tag

pavlis:
This is a change only to mspass.yaml to fix an unambiguous mistake in mspass schema.  I found this by running clean_collection in the short course tutorial and found it cleared data_tag I had set in the save.  I wasn't sure if that attribute should be tagged "normal" or "optional".   I set it "normal", but am a bit fuzzy on the distinction

### DIFF
--- a/data/yaml/mspass.yaml
+++ b/data/yaml/mspass.yaml
@@ -117,6 +117,11 @@ Database:
         elog_id:
           reference: elog
           constraint: xref_key
+        data_tag:
+           type: string
+           concept:  Used as a tag for data at a particular stage of processing. Used in save_data.
+           constraint: normal
+
 
   wf_Seismogram:
     data_type: Seismogram
--- a/data/yaml/mspass.yaml
+++ b/data/yaml/mspass.yaml
@@ -118,9 +118,9 @@ Database:
           reference: elog
           constraint: xref_key
         data_tag:
-           type: string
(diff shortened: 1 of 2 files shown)


haruming:
I think the distinction between normal and optional is, normal is the attribute that doesn't hurt the workflow if it doesn't exist in the object, but it would be great if the object contains this attribute. And optional is the attribute that is the lowest priority in the object. I think for data_tag, it's fine for either normal or optional. By the way, we may probably add data_tag in the mspass_lite.yaml, shall we?

wangyinz:
Yes, I agree, and I think data tag should be optional according to the definition. I think the distinction is (at least when we first implement it) that when a normal attribute is missing, we kill the object but does not abort the workflow. 

We definitely need it in the `mspass_lite.yaml`, too, if we are adding it to the `mspass.yaml`

wangyinz:
Yes, I agree, and I think data tag should be optional according to the definition. I think the distinction is (at least when we first implement it) that when a normal attribute is missing, we kill the object but does not abort the workflow. 

We definitely need it in the `mspass_lite.yaml`, too, if we are adding it to the `mspass.yaml`

haruming:
I changed the constraint to optional and also added the attribute in the mspass_lite.yaml. Hopefully, it's ready to merge.

PR #231, opened 2021-07-28
Change the structure of the RFdeconProcessor to make it serializable

wangyinz:
We run into the error that RFdeconProcessor was not able to be serialized by pickle automatically when using it in a map call. This is because a lot of the member variables of this python object are C++ bonded objects. Therefore, we are will not be able to use the object at all unless we serialize all the related C++ objects with pybind11. This includes the different decon operators as well as other underlying objects such as `ComplexArray` and `ShapingWavelet`. 

Therefore, instead of implementing the pickle method of all those objects, I decided to change the `RFdeconProcessor` python object such that it does not store copies of the actual decon processor. Instead, it creates one on the fly with all the necessary parameters stored in it that are native to Python. The overhead of this approach shouldn't be too much, and I think this change may actually be faster than serializing the decon operators in a parallel workflow.

 The other issue I revealed when testing it is that the decon operator depends on the `PFPATH` env that is not defined in either our test workflow or the docker container. I am not sure how you were able to use it in the container previously. Maybe we should change the default `PFPATH` to be a relative path to the already defined `MSPASS_HOME`.

### DIFF
--- a/cxx/python/seismic/seismic_py.cc
+++ b/cxx/python/seismic/seismic_py.cc
@@ -292,14 +292,7 @@ PYBIND11_MODULE(seismic, m) {
     .def_property("tmatrix",
       [](const CoreSeismogram &self){
         dmatrix tm = self.get_transformation_matrix();
-        auto v = static_cast<Publicdmatrix&>(tm).ary;
-        std::vector<double>* c = new std::vector<double>(std::move(v));
-        auto capsule = py::capsule(c, [](void *x) { delete reinterpret_cast<std::vector<double>*>(x); });
-        std::vector<ssize_t> size(2,3);
-        std::vector<ssize_t> stride(2);
-        stride[0] = sizeof(double);
-        stride[1] = sizeof(double) * 3;
-        return py::array(py::dtype(py::format_descriptor<double>::format()), size, stride, c->data(), capsule);
+        return pybind11::module::import("numpy").attr("array")(tm);
       },
       [](CoreSeismogram &self, py::object tm) {
         self.set_transformation_matrix(tm);
--- a/cxx/src/lib/utility/AntelopePf.cc
+++ b/cxx/src/lib/utility/AntelopePf.cc
@@ -7,6 +7,7 @@
 #include <fstream>
 #include <sstream>
 #include "mspass/utility/MsPASSError.h"
+#include "mspass/utility/utility.h"
 #include "mspass/utility/AntelopePf.h"
 namespace mspass::utility {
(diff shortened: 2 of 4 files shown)


wangyinz:
For the record here, after the change, dask still throws this error:
```
distributed.protocol.pickle - INFO - Failed to serialize (<function reify at 0x2b54acabf9d8>, (<function map_chunk at 0x2b54acabfd90>, <function RFdecon at 0x2b54abd19400>, [(<function map_chunk at 0x2b54acabfd90>, <function window_seis at 0x2b54ac3ece18>, [(<function map_chunk at 0x2b54acabfd90>, <function filter at 0x2b54a8f406a8>, [(<function map_chunk at 0x2b54acabfd90>, <function detrend at 0x2b54a8f409d8>, [(<function map_chunk at 0x2b54acabfd90>, <function read_distributed_data.<locals>.<lambda> at 0x2b54ae773840>, ["('from_sequence-f92947ce5c6b439d1f643091613b0284', 8)"], None, {}), (<class 'itertools.repeat'>, 'demean')], None, {}), (<class 'itertools.repeat'>, 'bandpass')], None, {'freqmin': 0.01, 'freqmax': 2.0})], None, {}), (<class 'itertools.repeat'>, <__main__.DeconProcessor object at 0x2b54ae3f64a8>)], None, {})). Exception: can't pickle PyCapsule objects
```
Not sure why and how, but I do found two places in our code base that uses `PyCapsule`: 
https://github.com/mspass-team/mspass/blob/897de8f6badc9c47f24b27ac3dfe7f9766714fe5/cxx/python/seismic/seismic_py.cc#L297
https://github.com/mspass-team/mspass/blob/99a8ac30de462a1270ee6dfaa528f44357188383/cxx/python/utility/utility_py.cc#L205

The one in `SphericalCoordinate` should be irrelevant, but maybe the problem is in the `Seismogram`. So, I have removed the use of `PyCapsule` there and we will see whether the problem goes away.

wangyinz:
@haruming Please check whether the PyCapsule issue goes away. If not, I guess we need to look into the other dependency libraries instead of mspass.

wangyinz:
@haruming Please check whether the PyCapsule issue goes away. If not, I guess we need to look into the other dependency libraries instead of mspass.

wangyinz:
So that did get the problem go away (according to @haruming's test). Not sure why, but I guess it has something to do with how Dask decides what needs to be pickled when a function is called. Although `tmatrix` is not explicitly used here, Dask somehow decides to serialize it.

PR #234, opened 2021-07-29
Add taper functions

wangyinz:
While working out the parallel workflow for the tutorial, we found that the taper functions also needs to be serializable. This branch includes the following changes:

- added wrapper function for tapers
- added basic algorithmic methods for the `DoubleVector` object and it now behaves more close to a numpy array
- made SphericalCoordinate and SlownessVector serializable as they can be used as arguments in a map call to certain algorithms


pavlis:
Good you fixed this.   These examples are also useful for how serialization is done in pybind11.

PR #235, opened 2021-07-29
Fix bundle

pavlis:
This should have been an easy merge, but I botched something when I was doing the git commits and got tangled up with recent changes made by you @wangyinz .   I think I resolved these changes right, but let's see how the tests go.  Review the changes carefully before you merge this with master.

The original purpose of this branch was to fix a bug in bundle that cause it to return an incorrect type (CoreSeismogramEnsemble instead of a SeismogramEnsemble).  The problem was missed changes to LoggingEnsemble from Ensemble in several places.   I also cleaned up what was likely a bug in the old operator= code for LoggingEnsemble.


pavlis:
@wangyinz  I am sorry I complicated this pull request.  I also see I should have run the C++ test_bundle program.   I'll look at that in the morning (tonight for you).

pavlis:
Tests are running but what I just pushed correct the mistake I made in the earlier version.  Simple fix once I found it, but pretty subtle.

wangyinz:
hmmm... I am not sure what the problem was, but the test failed all the time because the new `test_data_object_operators.py` file is not a valid test. I make an attempt to convert that into a valid test, but it seem the code relies on a lot of print statements. It seems most of them can be converted to `assert` lines. However, it appears that these tests are irrelevant to the edits you made in this branch. Is that some relic?

wangyinz:
hmmm... I am not sure what the problem was, but the test failed all the time because the new `test_data_object_operators.py` file is not a valid test. I make an attempt to convert that into a valid test, but it seem the code relies on a lot of print statements. It seems most of them can be converted to `assert` lines. However, it appears that these tests are irrelevant to the edits you made in this branch. Is that some relic?

wangyinz:
Oh, is that because you accidentally reused an old branch?

wangyinz:
Oh, is that because you accidentally reused an old branch?

pavlis:
No, I don't think I used an old branch.  The history was I forked from origin a couple of days ago.  I assume I would updated with a pull, but it is not inconceivable I messed up that way too, but don't really think so as I am pretty sure I wanted to be sure I had the latest version.   I did the fixes on that branch, but the mistake I made was when I was trying to push the new branch.   I'm not exactly sure what I did wrong, but I go into git hell with warnings about detached HEAD, needing to pull, and other errors I don't remember that I fear got me confused.   What is here was finally created by a `git pull origin master` when I was on this branch.   That created some conflicts I resolved with an editor.   I was then able to push this branch and issue the pull request.   This morning's fix was a simple push to this branch.  That fixed the failing of the C++ test program, but now this failure is happening in a python function I don't recognize.  I don't know for sure how to interpret this output to understand what to chase?

wangyinz:
well... I guess I should just remove that test file then. Not sure how you got the history mixed up, but you can see there are a bunch of commits that are from Dec 2020.

PR #238, opened 2021-08-06
SAGE 2021 job script update

haruming:
upload job scripts

### DIFF
--- a/scripts/start-mspass.sh
+++ b/scripts/start-mspass.sh
@@ -85,7 +85,7 @@ if [ $# -eq 0 ]; then
         scp -r -o StrictHostKeyChecking=no ${i} ${MSPASS_LOG_DIR}
       done
     fi
-    sleep sleep ${MSPASS_SLEEP_TIME}
+    sleep ${MSPASS_SLEEP_TIME}
   }
 
   function start_db_scratch {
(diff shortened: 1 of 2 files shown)


scripts/start-mspass.sh
```diff
@@ -85,7 +85,7 @@ if [ $# -eq 0 ]; then
         scp -r -o StrictHostKeyChecking=no ${i} ${MSPASS_LOG_DIR}
       done
     fi
-    sleep 30
+    sleep sleep ${MSPASS_SLEEP_TIME}
```
wangyinz, line 88: duplicated sleep here

scripts/tacc_examples/SAGE_2021_script.sh
```diff
@@ -0,0 +1,72 @@
...
+#SBATCH -p normal          # Queue (partition) name
+#SBATCH -N 2               # Total # of nodes (must be 1 for serial)
+#SBATCH -n 2               # Total # of mpi tasks (should be 1 for serial)
+#SBATCH -t 02:00:00        # Run time (hh:mm:ss)
+#SBATCH -A MsPASS          # Allocation name (req'd if you have more than 1)
+
+# working directory
+WORK_DIR=$SCRATCH/SAGE_2021
+# directory where contains docker image
+MSPASS_CONTAINER=$WORK2/mspass/mspass_latest.sif
```
wangyinz, line 13: I think the work filesystem is already migrated. We should just use $WORK instead of $WORK2

PR #243, opened 2021-08-31
python code check plugin

haruming:
use Black in git action to automatically check the python code and reformat it if necessary


wangyinz:
closed as #248 is merged.

PR #263, opened 2021-11-03
Fix bug in failing to handle added live/dead feature

pavlis:
This is a small fix to handle a different bug in read_ensemble than the one discussed [here](https://github.com/mspass-team/mspass/issues/262).  Some time ago we extended Ensembles to add an error log and a feature wherein an entire ensemble can be killed (set dead).   The problem this bug fixes is ensembles read from MongoDB are always left with the state marked dead.   This probably didn't surface before because we didn't have a test workflow that properly tested that condition.  

This fix is pretty simple and only consists of a handful of additional lines of code.  One thing that might be open to discussion is how I did handled the inverse - saving an ensemble.  I added code that if the ensemble is marked dead save_ensemble does nothing.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -2246,6 +2246,12 @@ class Database(pymongo.database.Database):
             if data:
                 ensemble.member.append(data)
 
+        # explicitly mark empty ensembles dead.  Otherwise assume if 
+        # we got this far we can mark it live
+        if len(ensemble.member) > 0:
+            ensemble.set_live()
+        else:
+            ensemble.kill()
         return ensemble
 
     def save_ensemble_data(self, ensemble_object, mode="promiscuous", storage_mode='gridfs', dir_list=None, dfile_list=None,
@@ -2271,6 +2277,10 @@ class Database(pymongo.database.Database):
         atomic saves will not lose their association with a unique ensemble
         indexing scheme.
 
+        A final feature of note is that an ensemble can be marked dead.
+        If the entire ensemble is set dead this function returns 
+        immediately and does nothing.
+
 
         :param ensemble_object: the ensemble you want to save.
         :type ensemble_object: either :class:`mspasspy.ccore.seismic.TimeSeriesEnsemble` or
@@ -2290,6 +2300,8 @@ class Database(pymongo.database.Database):
         :param data_tag: a user specified "data_tag" key to tag the saved wf document.
         :type data_tag: :class:`str`
         """
+        if ensemble_object.dead():
+            return
         if not dfile_list:
             dfile_list = [None for _ in range(len(ensemble_object.member))]
         if not dir_list:
--- a/python/tests/db/test_db.py
+++ b/python/tests/db/test_db.py
@@ -1489,6 +1489,7 @@ class TestDatabase():
         ts_ensemble.member.append(ts1)
         ts_ensemble.member.append(ts2)
         ts_ensemble.member.append(ts3)
+        ts_ensemble.set_live()
         dfile_list = ['test_db_output', 'test_db_output']
         dir_list = ['python/tests/data/', 'python/tests/data/']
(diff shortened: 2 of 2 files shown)


pavlis:
Thought of one additional point that may require some attention.   When I think about it doing nothing when an ensemble is marked dead may be incorrect.  I think the code needs to check handle errors in elog of the ensemble and members and handle history.   That makes this much more complicated.   Consider accepting this branch to fix this major bug in the reader and put doing the save right on the development agenda.  The reader is mostly broken without the fix applied here.

pavlis:
Hmmm - not sure why those tests are failing.   I need to punt this to one of you guys who can run pytest to work this out. 

I found an unrelated error in database.py.   The section of _read_data_from_dfile below has a debug print statement that need to be removed.  Here is the offending section (lines 2838-2861):
```
        if not isinstance(mspass_object, (TimeSeries, Seismogram)):
            raise TypeError("only TimeSeries and Seismogram are supported")
        fname = os.path.join(dir, dfile)
        with open(fname, mode='rb') as fh:
            fh.seek(foff)
            if not format:
                float_array = array('d')
                if isinstance(mspass_object, TimeSeries):
                    if not mspass_object.is_defined('npts'):
                        raise KeyError("npts is not defined")
                    float_array.frombytes(
                        fh.read(mspass_object.get('npts') * 8))
                    mspass_object.data = DoubleVector(float_array)
                elif isinstance(mspass_object, Seismogram):
                    if not mspass_object.is_defined('npts'):
                        raise KeyError("npts is not defined")
                    float_array.frombytes(
                        fh.read(mspass_object.get('npts') * 8 * 3))
                    print(len(float_array))
                    mspass_object.data = dmatrix(3, mspass_object.get('npts'))
                    for i in range(3):
                        for j in range(mspass_object.get('npts')):
                            mspass_object.data[i, j] = float_array[i *
                                                                   mspass_object.get('npts') + j]
```
The print statement on line 2856 needs to be removed.

Yangzhengtang:
Hi Gary, I ran the pytest on my local machine, and the failure output is as follows:

============================================================================= FAILURES ==============================================================================
_______________________________________________________________ TestDatabase.test_save_ensemble_data ________________________________________________________________

self = <test_db.TestDatabase object at 0x17e26faf0>

    def test_save_ensemble_data(self):
        ts1 = copy.deepcopy(self.test_ts)
        ts2 = copy.deepcopy(self.test_ts)
        ts3 = copy.deepcopy(self.test_ts)
        logging_helper.info(ts1, '1', 'deepcopy')
        logging_helper.info(ts2, '1', 'deepcopy')
        logging_helper.info(ts3, '1', 'deepcopy')
        ts_ensemble = TimeSeriesEnsemble()
        ts_ensemble.member.append(ts1)
        ts_ensemble.member.append(ts2)
        ts_ensemble.member.append(ts3)
        dfile_list = ['test_db_output', 'test_db_output']
        dir_list = ['python/tests/data/', 'python/tests/data/']
        self.db.save_ensemble_data(ts_ensemble, mode="promiscuous", storage_mode='file', dfile_list=dfile_list, dir_list=dir_list, exclude_objects=[1])
        self.db.database_schema.set_default('wf_TimeSeries', 'wf')
>       res = self.db.read_data(ts_ensemble.member[0]['_id'], mode='promiscuous', normalize=['site', 'source', 'channel'])
**E       mspasspy.ccore.utility.MsPASSError: Error trying to extract Metadata with key=_id
E       No value associated with this key is set in Metadata object
E       Expected an entry of type=boost::any**
python/tests/db/test_db.py:1496: MsPASSError

I don't know these parts of code very well. But I guess the problem was in the 'ts_ensemble' variable in the test case above, it is not live when passed into save_ensemble_data, so nothing will be done to it. And after calling save_ensemble_data, the test is trying to access ts_ensemble.member[0]['_id'], which is not defined, that would throw an exception.

Issue #261, opened 2021-11-01
Question in mseed_file_indexer when reading mseed file


niyiyu:
Hi MsPASS team,

Attached is the issue I found in a recent test. The MsPASS is grabbed from Docker image mspass/mspass:latest. And the mseed files are randomly chosen from thousands of files downloaded with ROVER. Here're two major issue:
- Sampling rate is good, but npts is somehow wrong, which probably leads to a wrong endtime? The ROVER downloads the data on a daily basis, so ending time like `2019-01-01T23:59:59.990000Z` is hopefully correct for this file.
- As you see, there're gaps in `UW.OSD..ENZ` channel. In MsPASS however, these gaps are omitted and traces are merged into one single trace.

You could get the file here: [UW.OSD.2019.001](https://drive.google.com/file/d/1YuF7jpLSgXZz47uxlW0eFhJ-RBDraEuL/view?usp=sharing) and [SIFT.CC.2020.005](https://drive.google.com/file/d/1N5D722KieHoKHV0R-HmC39sG5jLgnNbP/view?usp=sharing).  Hope this helps. 

Thanks!

```
In [1]: from mspasspy.ccore.io import _mseed_file_indexer
   ...: import obspy

In [2]: for i in _mseed_file_indexer("OSD.UW.2019.001"):
   ...:     print(i.chan,
   ...:           i.samprate,
   ...:           i.npts,
   ...:           obspy.core.utcdatetime.UTCDateTime(i.starttime),
   ...:           obspy.core.utcdatetime.UTCDateTime(i.endtime))
   ...: 
EHZ 100.0 8640000 2019-01-01T00:00:00.000000Z 2019-01-02T00:00:00.000000Z
ENE 100.0 8640677 2019-01-01T00:00:00.000000Z 2019-01-02T00:00:06.770000Z
ENN 100.0 8640007 2019-01-01T00:00:00.000000Z 2019-01-02T00:00:00.070000Z
ENZ 100.0 8325129 2019-01-01T00:00:00.000000Z 2019-01-01T23:07:31.290000Z

In [3]: st = obspy.read("OSD.UW.2019.001")
   ...: st
Out[3]: 
9 Trace(s) in Stream:
UW.OSD..EHZ | 2019-01-01T00:00:00.000000Z - 2019-01-01T23:59:59.990000Z | 100.0 Hz, 8640000 samples
[...]

PR #267, opened 2021-11-05
bugfix for the mseed gaps

haruming:
This is the bugfix for issue #261 
Basically, we use the obspy built-in merge function to fix this.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -216,9 +216,9 @@ class Database(pymongo.database.Database):
         alg_id="0",
         define_as_raw=False,
         retrieve_history_record=False,
-        method=0,
-        fill_value=None,
-        interpolation_samples=0
+        merge_method=0,
+        merge_fill_value=None,
+        merge_interpolation_samples=0
     ):
         """
         This is the core MsPASS reader for constructing Seismogram or TimeSeries
@@ -459,9 +459,9 @@ class Database(pymongo.database.Database):
                         object_doc["foff"],
                         nbytes=object_doc["nbytes"],
                         format=object_doc["format"],
-                        method=method,
-                        fill_value=fill_value,
-                        interpolation_samples=interpolation_samples
+                        merge_method=merge_method,
+                        merge_fill_value=merge_fill_value,
+                        merge_interpolation_samples=merge_interpolation_samples
                     )
                 else:
                     self._read_data_from_dfile(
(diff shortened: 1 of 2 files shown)


python/mspasspy/db/database.py
```diff
@@ -203,8 +203,21 @@ def set_database_schema(self, schema):
...
+        normalize=None,
+        load_history=False,
+        exclude_keys=None,
+        collection="wf",
+        data_tag=None,
+        alg_name="read_data",
+        alg_id="0",
+        define_as_raw=False,
+        retrieve_history_record=False,
+        fill_value=None,
```
wangyinz, line 219: We should probably add all three parameters from obspy here: `Stream.merge(method=0, fill_value=None, interpolation_samples=0)`. We can point user to obspy's detailed documentation here, which includes [this](https://docs.obspy.org/packages/autogen/obspy.core.trace.Trace.__add__.html#handling-overlaps) and [that](https://docs.obspy.org/packages/autogen/obspy.core.stream.Stream._cleanup.html#obspy.core.stream.Stream._cleanup).

python/mspasspy/db/database.py
```diff
@@ -269,6 +282,16 @@ def read_data(self, object_id, mode='promiscuous', normalize=None, load_history=
...
         :type data_tag: :class:`str`
+        :param alg_name: alg_name is the name the func we are gonna save while preserving the history.
+        :type alg_name: :class:`str`
+        :param alg_id: alg_id is a unique id to record the usage of func while preserving the history.
+        :type alg_id: :class:`bson.objectid.ObjectId`
+        :param define_as_raw: a boolean control whether we would like to set_as_origin when loading processing history
+        :type define_as_raw: :class:`bool`
+        :param retrieve_history_record: a boolean control whether we would like to load processing history
+        :type retrieve_history_record: :class:`bool`
+        :param fill_value: Fill value for gaps. Defaults to None. Traces will be converted to NumPy masked arrays if no value is given and gaps are present.
```
wangyinz, line 293: "Traces will be converted to NumPy masked arrays if no value is given and gaps are present" I don't think this is true. This is what obspy does, but in our data object we don't have such feature.

wangyinz (changes requested):
Very good. There is only some small issues need to be addressed.

python/mspasspy/db/database.py
haruming: Agreed.

python/mspasspy/db/database.py
```diff
@@ -216,7 +216,9 @@ def read_data(
         alg_id="0",
         define_as_raw=False,
         retrieve_history_record=False,
+        method=0,
```
wangyinz, line 219: Sorry that I should've thought about this: we should probably name it "merge_method" instead. Probably should also add "merge_" to the next two parameters. The names are confusing as this is the read_data method.

wangyinz (changes requested):

python/mspasspy/db/database.py
haruming: Yeah I agree that it's a bit confusing and change to `merge_` it makes sense.

wangyinz:
hmmm.... why did it failed? Could you please look into it @haruming ? Thank you!

PR #268, opened 2021-11-07
Add taper serialization

pavlis:
This branch adds missing serialization code for the set of objects defined in taper.h.   That includes true two-sided tapers as used, for example, in spectral analysis and the TopMute class (mapped in pybind11 code to _TopMute).

I didn't add test pickle code for TopMute because I still don't have pytest fully implemented.  I notice that I don't think there is any test code for Taper classes that I added pickle to either:  LinearTaper, CosineTaper, and VectorTaper.  

There are some minor changes to the C++ taper test program that do test the boost serialization used in the C code.  The pybind11 serialization with pickle, however, is completely different for LinearTaper, CosineTaper, and TopMute.  VectorTaper directly applies the boost serialization. The reason I did that is that the first 3 are defined by 3 to 4 numbers that are easily serialized as a tuple in pybind11.   Should be must faster than preserving all the baggage of serializing the full object interface.  Note I tested all 4 pickle instances with the same one we use in all the other tests:
```
#data_object is the object we want to test with pickle
dcopy=pickle.loads(pickle.dumps(data_object))
# followed by assert to verify copy and original are equal
```
I can add the test code for TopMute easily of you want.  The taper objects, I think, are harder because we don't have any existing tests that I found.


Yangzhengtang:
Sure! I will work on it soon.

PR #269, opened 2021-11-11
Add table import

Yangzhengtang:
Break down the original import_table function into two seperate steps: 
Textfile2Dataframe and db.save_dataframe

Modify Textfile2Dataframe, make the pd and dask dataframe API consistent.

Remove datascope.py and textfile.py.
(Functions in textfile.py are now in converter.py and database.py)


python/mspasspy/db/database.py
```diff
@@ -3804,8 +4421,112 @@ def index_mseed_file(self, dfile, dir=None, collection='wf_miniseed'):
...
-            doc['format'] = 'mseed'
-            doc['dir'] = odir
-            doc['dfile'] = dfile
+            doc["storage_mode"] = "file"
+            doc["format"] = "mseed"
+            doc["dir"] = odir
+            doc["dfile"] = dfile
             dbh.insert_one(doc)
+
+    def save_dataframe(self, collection, df, parallel, null_values=None, one_to_one=True):
```
wangyinz, line 4430: To be consistent with the API of `save_data` as well as `save_textfile` below, I think we should probably change the order of parameters to `save_dataframe(self, df, collection, null_values=None, one_to_one=True, parallel=False)`

python/mspasspy/db/database.py
```diff
@@ -3804,8 +4421,112 @@ def index_mseed_file(self, dfile, dir=None, collection='wf_miniseed'):
...
+        if parallel:
+            df = df.compute()
+            
+        df = df.apply(pd.Series.dropna, axis=1)
+        doc_list = df.to_dict(orient="records")
+        if len(doc_list):
+            dbcol.insert_many(doc_list)
+        return len(doc_list)
+
+    def save_textfile(
```
wangyinz, line 4497: Although it appears redundant, we do want to add all the parameter definitions in the docstrings here.

python/mspasspy/util/converter.py
```diff
@@ -480,6 +482,185 @@ def TimeSeriesEnsemble2Stream(tse):
...
+        for k in names:
+            nulls[k] = nullvalues[i]
+            i += 1
+
+    return tuple([names, dtypes, nulls])
+
+
+def Textfile2Dataframe(
+    filename,
+    separator="\s+",
```
wangyinz, line 551: May want to state clearly that the default separator is a regular expression of "one or more spaces". We should also point the user to the corresponding documentations of the underlying read_csv calls (i.e. [this](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html))

wangyinz (changes requested):
I am not sure whether this is ready for review, but I assumed so. Please request for review next time when you have something ready so that I can act on it faster. 

This looks pretty good, and I only have a couple minor suggestions.

wangyinz (approved):

PR #270, opened 2021-11-15
Improve mseed

pavlis:
This branch implements time tear processing in the C++ interface function with libmseed.  The api for the C++ function has changed dramatically with arguments added to the indexing function AND it now has a return that is a bound with pybind11 to a tuple with the index and a new ErrorLogger object.  I modified the index_mseed_file method of database to handle this new error logging return and save elog entries to MongoDB automatically.   That required a minor change to the schema for elog.  

This version should be considered a prototype for comment.  It has two serious issues that need to be dealt with before this branch should be merged with master.  The issues are in an email to Chad Trabant at IRIS who is the main author of libmseed, but I summarize them here for the record:
1.  We need to change the read function from ms3_readmsr to ms3_readmsr_r.  The latter support threading cleanly.  The only reason that didn't bite us before is I don't think any of us ever tried parallel reads with dask or spark driven by a list of file names.  
2. We are not using libmseed's internal error logging mechanism properly.   I could not figure out how that is supposed to be done.  A change is needed to extract messages from the library and translate them to elog messages in the C++ code whenever libseed posts an error message.  I think by default any such errors are either dropped on the floor or send to stderr.   We need to find out from Trabant before we can finalize this. 

For now have a critical look, especially, at the api changes I made.  I made everything backward compatible and have tested to verify it works on a clean miniseed file the same way it did before.  I could not easily find a file with errors that we should use to test the new functionality.


pavlis:
TODO to finalize this pull request:
1.  Resolve two pending questions to Chad Trabant about libmseed issues.
2. Implement a windowed indexer that would limit the index to specified time window.
3. Add a gap threshold parameter to discard data with excessive gaps.

pavlis:
I just changed the code to use the thread safe reader.   I have no easy way to test it, however, that would test if it truly is thread safe.   For now that functionality is purely theoretical.   I am not sure if we have a test python function to use this reader. If not, we need one to verify this really works at all before merging this branch.

I have not yet cracked the error logging issue.  I urge you, @wangyinz , to validate that this reader at least works single threaded and merge it with master.  If you know how to design a multithread test please do - I do not know how to do that without some serious digging.  See parallel email to Chad Trabant I will copy @wangyinz  about the error logging issue.

pavlis:
I don't think the errors that are causing the test to fail are mine.  I don't quite understand that as this branch was created from a direct fork of master a few weeks ago.   

I see from the TODO list I also have some more work to do on this also.  I forgot about the need for a windowing option and a gap threshold option. Ignore this pull request until I have time to work those out.

pavlis:
I started work on finalizing these changes to the mseed reader.   There are three points worth discussing as the proceeds.
1.  First, turns out this implementation has the automatic gap segmentation option.  
2. I am finding the error log implementation ever more tricky than that i thought it was.   I don't want to get too deep in the weeds, but this gets into some nasty issues about two things:  (a) huge complexities in C++ about what "static" means in different contexts, and (2) mismatches in what the idea of a "function pointer" means in C and C++.   I'm working on a solution but what I'm most struggling with is how I'm going to even test this code?   libmseed does indeed have an internal logging mechanism we really do need to tap into to get error messages that are meaningful.   There are A LOT of places that messages can get logged in libmseed, but I think they can be reduce to two forms:  (a) bad errors that it (by default) things of as stderr and (b) less serious messages it things of as messages you might write to stdout (informational or warning messages).   At least  have to produce two interface functions to catch log messages emited to those two error streams and put them into our ErrorLogger structure.   I think I know how to do that now, but I need to find a way to read a file that has errors that test both handlers.   I really don't know how to do that unless I write a C++ test program that writes a file and reads it back.   yuck.   Any better ideas?
3. I am becoming resistant to the idea of have an option on the mseed_file_indexer function to index a restricted time window.   On reflection I think that is a really bad idea for data stored in files.   The idea emerged from reading from the cloud and on reflection I think that needs to be treated as a different concept.   A file is a file so let's not mess up the concept of what an index to that file means.   It seems to me if a person needs to segment data the right model is to use our existing WindowData function and if the user needs to manage storage it is there local problem to know how to deal with that.   Another reason this strikes me as dangerous is there are many cases where someone will want a segment of data that spans two files.   Then the very concept of that indexing breaks. 

Point 3 reminds me of a standard function we need to design.  We have [this function in obspy](https://docs.obspy.org/packages/autogen/obspy.core.stream.Stream.merge.html) but I think we could do something much cleaner with MongoDB.  Let me move this to a discussion topic - it should appear shortly.

pavlis:
I have not implemented the code to handle the libmseed error log translation to our ErrorLogger.  It is totally untested, however, and I don't know how to actually test it short of releasing this on a lot of data.   We need to figure out how to do that before we merge this.  One possibility is I deploy a dev version from this pull request and run it on a year of usarray data I have staged.  there would undoubtedly be some bad data in that much data.

wangyinz:
For the question in 2, I think we may use obspy to create some problematic mseed files. I guess the question is what kind of erroneous files are needed. The most obvious one is an invalid file, which can just be a random file that is not a mseed file. Then, what else?

I also agree that the segmentation option invalidates the idea of indexing. I guess the very option should really be in the reader. It seems that is already in [obspy reader](https://docs.obspy.org/packages/autogen/obspy.core.stream.read.html) that we are using right now. We might just add those parameters to our API.

wangyinz:
For testing, I think it will be easier to use the dev container, but we need to merge in the master first. Let me try to resolve the conflict.

pavlis:
In response to the issue about what we test, that is not so easy to answer.   I don't have a copy of libmseed on this laptop where I'm composing this, but grep for ms_log in the source code and you will it is called A LOT in many contexts.   As I understand it any call to ms_log causes the logging function defined ms_loginit to be invoked.   This version implements that in a very simple way that may or may not work.   If I understand the subtle scope rules of the new (to C++11  "thread_local" declaration) i should be ok and be thread safe when dask runs with the thread scheduler (I presume Spark has a similar construct when it is initialized).   I know this works on a simple "hello world" type test program and I have very reason to think it will do what it is supposed to do in this function.   What I'm most concerned is that  it will (a) create poorly formatted messages and/or (b) sometimes have garbled messages because it isn't really thread safe (or worse  crash).   

Anyway, designing a test for that is not trivial.  Not at all sure how to test it in a threaded environment other than releasing it on a bunch of data.   I guess this is an example, however, of an sidebar test to verify an algorithm is working correctly compared to the automated regression tests you use for a pull request.   We need both.     

I fear this comment just muddies this issue but it illustrates that this problem is very much at the fringe of my expertise and likely will be for you too.  In many ways libmseed shows the ugly side of plain C.   It definitely is one more bit of evidence for the long term joke in our field that "SEED is a four letter word"

pavlis:
In the call today there was concern that the elog entries created by mseed_file_index were not being saved properly.  I checked and there definitely are code changes in database.py related to that that issue.  I think it is done right, but that is again something that we need tests to validate.  I'm going to try to run this on quakes on a year of usarray data and see what happens.

pavlis:
I knew ms_log was a common function call in libmseed.  I did a grep of the source code and found 709 individual calls to that function in the library.  That shows the problem in trying to know what we should test. 

I did notice this block in the reader section of libmseed that suggests a simple approach worth trying:
```
    if (msio_feof (&msfp->input) && MSFPBUFLEN (msfp) < MINRECLEN)
    {
      if (msfp->recordcount == 0)
      {
        ms_log (2, "%s: No data records read, not SEED?\n", msfp->path);
        retcode = MS_NOTSEED;
      }
      else
      {
        retcode = MS_ENDOFFILE;
      }

      break;
    }
```
I think if you just give the reader a file with some random bytes this error might get thrown. That might work for an automated test.   How could we implement that in pytest script?

pavlis:
Unless you see a problem I'm missing, I think you can merge this branch to master.   I just ran this branch on a 2012 extended usarray data set.  I worked for with 3.5 million records so I think it is working.  I warn, however, that there are not formalized tests for the new features.  Further, my test wasn't definitive either by any means as I did not run it in parallel with the dask thread scheduler - that is the only situation I can see where using the multithreaded reader matters.   Nonethelss, I advise we merge this and deal with more complete testing later.   We can be 100% sure it will work with process schedulers.

PR #275, opened 2021-12-10
Add snr functions

pavlis:
This branch adds a new suite of algorithms for computing signal-to-noise ratio.   There is a basic "snr" function that can be used to compute one of four possible time-domain based snr metrics.  There is also a much more elaborate hierarchy of functions that build up to a special function that computes a set of metrics in the frequency domain and then optionally computes a set of time-domain metrics in the passband computed dynamically by a C++ function called EstimateBandwidth.  That function used to be a file-scope member of the CNR3CDecon.cc file, but I moved it out and constructed wrappers so it can be called from python.  

This branch is NOT ready at this point to merge to master.   I need to test the database capability but more importantly there is a bug that has me totally mystified.   I could use help from any of  you, but especially @wangyinz who knows the oddities of pybind11.   The bug is that I cannot get pickle to work correctly with the binding code I created for PowerSpectrum in the seismic module.  From the dev container you should be able to construct the version created by this branch.  Then run this small script:
```
import pickle
from mspasspy.ccore.seismic import TimeSeries,PowerSpectrum
from mspasspy.ccore.algorithms.deconvolution import MTPowerSpectrumEngine

d=TimeSeries(100)
d.data[0] = 1.0    # delta function - spectrum will be flat
d.live=True
engine=MTPowerSpectrumEngine(100,5,10)
D = engine.apply(d)
pickled_D = pickle.dumps(D)
```
I presume you will get the same error I do which is this:
```
  File "pickle_test_powerspectrum.py", line 10, in <module>
    pickled_D = pickle.dumps(D)
TypeError: 'mspasspy.ccore.seismic.PowerSpectrum' object is not iterable
```
Running this in a debugger it is easy to verify that the contents of the D symbol are intact and what is expected.   All the member functions give the answers I expect and all the attributes are set to what I think they should be and are accessible from python.  What is "not iterable" it the object is not clear.  

This error is mysterious for two reasons:
1.   I have viewed the pickle section of seismic_py.cc so many times I'm going blind looking at it and cannot for the life of me see anything wrong with the code.  I modeled the pickle section for PowerSpectrum after that for TimeSeries that has a pickle definition in the same file.   They are very similar because both inherit Metadata, both have an ErrorLogger member, and both have an std::vector container to hold the main data.  Compare the pickle lambda for TimeSeries and PowerSpectrum and you will see what I mean.   This may well be some silly little hidden typo, but it is driving me crazy. 
2. I have tried but found few useful sources with a google search of various permutation of: pybind11, TypeError: object is not iterable, and pickle.   Most are either some other situation where this low level message appears or otherwise completely unrelated

One thing I do know is this is clearly coming from a python layer in pickle., BUT I cannot get to it by just stepping into the pickle.dumps function.  I am pretty certain the pickle module is a compiled C so file based on what I read and the way it acts.  If I knew how to get to the source code for pickle.dumps I would search for what condition throws that exception and maybe make some progress, but I don't know how to do that.   On the other hand, google conversations I've read suggest it may be even lower level than that and something thrown by the interpreter when pickle does something that cause a low level exception to be thrown (the TypeError).  

I'm hoping @wangyinz can have an independent look at this and maybe explore this from a direction I haven't yet thought of.  

I would also appreciate comments on the functionality of this algorithm.  It has a ton of options that may be putting to much into one package.


pavlis:
PS:  note there are a bunch of unrelated fixes bundled into this branch.  They are bugs found by working with this new algorithm.

pavlis:
I tried some new things that reduce the universe a lot, but the reasons for this remain a deep mystery to me.  

What I tried this morning was a process of elimination method of problem solving.  The original code for PowerSpectrum created this tuple in the pickle section of the pybind11 code:
```
return py::make_tuple(sbuf,self.df,self.f0,self.spectrum_type,
        ss_elog.str(),darr);
```
By removing one component at a time and testing I found that the source of the "object is not iterable" error is the sbuf argument that above is the first argument to make_tuple.   sbuf is created like this:
```
pybind11::object sbuf;
sbuf=serialize_metadata_py(self);
```
I tried one other trick that failed.  PowerSpectrum inherits Metadata in a much simpler manner than other objects using this construct (i.e. TimeSeries and Seismogram) as follows:
```
class PowerSpectrum : public mspass::utility::Metadata
```
so I tried 
```
pybind11::object sbuf;
sbuf=serialize_metadata_py(dynamic_cast<const Metadata&>(self));
```
but I still get the "is not iterable" error when I put sbuf in the tuple.   

This remains baffling because the serialize_metadata_py function is used in at least four other places in the same file that I put the new PowerSpectrum binding code (seismic_py.cc) - i.e. TimeSeries, Seismogram, SeismogramEnsemble, and TimeSeriesEnsemble all use serialize_metadata_py in exactly the same way.   All of them are known to work so what in the world is different here?

wangyinz:
I just got the problem fixed. Turns out all you need to do is to make pybind11 aware of that the `PowerSpectrum` should be a child of `Metadata`.

pavlis:
That is one of those things where the only relies are one of two common colloquial sayings:  (1) doohh - from the Simpsons, or (2) slap me on the side of the head (origin not known to me).   Thank you so much for figuring that out.  A classic example where you get blind to a solution focusing too much on details and missing something outside of what you are focused on.  I will implement this and start some serious testing of it.

pavlis:
I just pushed a major revision of this branch.   pytest in github is still running as I write this, but this branch is almost ready to merge to master.   Three points to consider:

1. I am sending you a complete test function for PowerSpectrum as part of full file I used outside pytest to test PowerSpectrum.   I did that a crude way because still don't fully have pytest implemented locally to try these things out.   I am sending you that script separately by email.   I think it would only take you a few minutes to integrate it into the test script for ccore. 
2. I tested the snr functions on the first session data set used for the sage-gage short course and it handles all that data in a serial run.   I haven't fully evaluated the numbers, but spot checks look good.  I'll give you some updates if I find anything.
3. You need to review the actual code api carefully.   The main function I think most users will be interested is in highest level function called arrival_snr_QC.   Anyone doing any kind of phase-based processing would likely want to run this function.  With default parameters running serial it seems to run for the order of 1 s per datum, which is not tiny if one has to do that calculation for millions of records (which would be common for many applications).  This code needs a test with dask and spark to verify it works that way as it very much needs that functionality. 

A final major hole is that we need to add the attributes this thing computes to our schema.   The arrival_snr_QC function, you will note, puts the results by default in a collection it calls arrival.  We don't currently even have an arrival collection schema defined.  There are very very good reasons to put these data into a different collection linked by an id.   The link, however, is not from wf to arrival but the reverse.    That is, the core idea here is to use some query based on QC metrics to only select data where the snr exceeds one or more of these computed measures.   Then the data set would be formed by joining the subsetted arrival collection to wf.   We actually probably need a prototype for how to do that with read_distributed_data.  

A lot to chew on really.   Look for my email @wangyinz

pavlis:
@wangyinz do not merge this until I fix something I discovered that is clearly wrong.   This implementation does not handle errors correctly.  It throws exceptions in places where it should post a message to elog.

pavlis:
Ignore my previous comment @wangyinz .   When I went to make that fix I realized that the exceptions were correct as used.  All of this code follows our guidelines.   the low level snr function can throw an exception that would need a handler if used.  The reason is that function is made to return a snr specified estimator as a single number.   In that situation handling elog is problematic if the function was run in parallel.

If you want I can write a thin wrapper that would return the snr estimator and set a Metadata field with the value using a key defined in the arg list.   i.e. we might create something like this:
```
def ComputeSNR(data_object,metadata_key='mad_snr',noise_window=TimeWindow(-130.0,-5.0),
  signal_window=TimeWindow(-5.0,120.0),noise_metric='mad',signal_metric='mad',
  perc=95.0):
```
which is identical to the snr function but adding the argument `metadata_key`.  The only other thing needed would be an error handler.  

Let me know if you want me to implement that and I will make it so.

pavlis:
One other thing - I think the checks are failing because of tangled branches.

PR #289, opened 2022-01-06
Fix bug when t0shift options is used

pavlis:
This fixes a bug in WindowData that did not handle the t0shift options correctly.  It failed every time it was called because the range test used did not apply the shift when that was enabled.  The changes are pretty trivial so I advise you simply merge this branch immediately. 

Longer term action is needed to have the tests check the t0shift option.  This bug was present because that test was missing.   I looked at the test code and decided it was not so trivial to fix because the existing test code is not set up for utc data where this feature would be used.   Advise you accept this pull request and post adding such a test to the fairly short term development agenda.


pavlis:
PS:  corollary to not putting in a test of this revision is  also never tested my fix.  The fix is pretty trivial, however, and doubt it will not work.

PR #292, opened 2022-01-11
Add data editing functionality

pavlis:
This is a complete package of (pure python) code to perform the important, generic capability of data editing (commonly called trace editing in seismic reflection processing).  This branch implements all python comparison operators that can be used to kill data (using the kill method) based on a comparison of the data's metadata value and one or more boundary values.   i.e. there are kill operators for >,>=,<,>=,==,and != as well as a "defined" or "undefined" test. There is also a generic interval tester that can handle and combination of a op x op b where a and b are bounds (the interval), x is the value being tested for a kill, and op is any of the binary python comparison operators (<,>, >=, etc.)   There is also a `FiringSquad` operator that implements a list of executioners that can be run on data in a single pass. 

I wrote a manual page because the way these are implemented is novel and based on a the object oriented concept of inheritance and functional programming that would be confusing to most seismologists. The manual page is essential for that reason.  

I didn't create and index entry for the new documentation page.   I think what we need is a hierarchy of sections we need to expand on "algorithms" with subsection pages on distinct functionality of MsPASS specific algorithms.  We probably should also move the obspy section to a subsection of algorithms.

### DIFF
--- a/docs/source/user_manual/DataEditing.rst
+++ b/docs/source/user_manual/DataEditing.rst
@@ -1,0 +1,349 @@
+Data Editing
+=======================
+Concepts
+------------
+MsPASS uses an idea that has been a part of seismic reflection from the
+earliest days of digital processing of seismic signals commonly called
+trace editing.   Trace editing means removing "bad" data from subsequent
+processing where the meaning of "bad" is based on some computed metric or
+human editing using a graphical user interface.   In all cases, the decision
+of "good" or "bad" is not grey but a black and white (boolean) decision.
+We implement this concept in MsPASS through a boolean `live` attribute
+that is a component of all MsPASS data objects (i.e. both atomic data
+TimeSeries and Seismogram as well as ensembles of same).   Any MsPASS
+data object can be "killed" by callig it's `kill` method.   Algorithms
+can and always should test the attribute `live` to verify the data object
+has data that can be viewed as valid.  No algorithm should ever assume
+data marked dead (live set False) can be used for any further processing.
+
+In MsPASS we abstract the process of defining "trace editing" through
+a standardized API.  In this section of the manual we describe how to
+use the generic metrics that are part of MsPASS.   We also show how
+users can add custom metrics by either computing custom Metadata or
+running an inline computation on the sample data.
+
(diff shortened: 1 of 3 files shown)


pavlis:
I think there is some internal issue causing the 3.8 tests to fail.   Not sure what you want me to do about the formatter complaints that are harmless.  

Also note I implemented a test_edit.py function in algorithms but I'm not sure I enabled it in this testing framework correctly.  It ran on my local copy fine.

pavlis:
The latest  push (1/26/2022) implements a class method decorator to handle error logging and history as with the function decorator.   

This branch isn't really ready to merge yet.  I want to add the inline metadata editor to this some edit.py module.  I pushed this to make sure the test_edit.py worked on all the tests as it did locally.

pavlis:
I just pushed a final update to this branch that should make it ready for assimilation into master.  I'm confused, however, because I don't see this pull request being updated.   Not sure what is going on.  

The latest change implements the capability to chain a set of operators together to do more complex calculations from Metadata attributes.  I also added a user manual page on the new features.

pavlis:
Maybe the conflicts have caused this request to not be update.  I don't know how to handle that

wangyinz:
I got the problem fixed by merging with the master. Also, the `test_edit.py` didn't work because it is not written to be recognized by pytest. I got that fixed too.

wangyinz:
@Yangzhengtang Could you please help Gary add some more test to this branch? For example, the later added wrapper code for object history is not tested at all. Also, when you go through the code, please also resolve some apparent issues such as:
https://github.com/mspass-team/mspass/blob/6f1505d1b83122c9ba6626cf4fc88e3746559efa/python/mspasspy/algorithms/edit.py#L26-L29
where the if branches are redundant. There could be a few minor issues like this in the branch. Just correct them when you see it. Thank you!

wangyinz:
btw, I think we should add a `__call__` method to the `Executioner` class so that instead of doing 
```
int_tester.kill_if_true(enscpy, apply_to_members=True)
```
we can just call this instead:
```
int_tester(enscpy, apply_to_members=True)
```

PR #299, opened 2022-01-19
Minor bug fix

pavlis:
I would suggest you do a quick review for anything obvious and merge this branch with master.  It contains only two minor things:
1. Fixed a bug in the CoreSeismogram constructor from a vector of CoreTimeSeries objects.  The previous version did not call set_live() on success so all data passed through it were marked dead.
2. I cleaned up a minor irregularity in handling dead dead for operator+= and unncessary relic code dealing with a const issue that is no longer relevant.


wangyinz:
Looks good to me

PR #300, opened 2022-01-24
Aws benchmark

Yangzhengtang:
Add new read_data and indexing methods that read/index data from AWS using S3 and Lambda API.
Add AwsLambdaClient module and examples to allow users to define, upload and use their own lambda functions.


data/yaml/mspass.yaml
```diff
@@ -196,6 +196,78 @@ Database:
...
+        reference: channel
+        constraint: normal
+      chan:
+        reference: channel
+        constraint: normal
+      channel_id:
+        reference: channel
+        constraint: xref_key
+
   site:
```
wangyinz, line 271: Instead of inserting this new content to our master yaml schema, let's create a new copy, maybe called `mspass_s3.yaml` and `mspass_fdsn.yaml`, and then replace the content of `wf_miniseed` section with corresponding lines.

python/mspasspy/db/database.py
```diff
@@ -4752,3 +5026,333 @@ def save_textfile(
...
+            LogType="Tail",
+            Payload=json.dumps(event),
+        )
+
+        response_payload = json.loads(response["Payload"].read())
+        ret_type = response_payload["ret_type"]
+
+        if (
+            ret_type == "key"
+        ):  #   The output file is saved to another bucket (s3_output_bucket). We don't support this workflow now.
```
wangyinz, line 5109: I thought we have tested with large return object before, and why is this not supported here? Is there a newer version of the implementation somewhere?

wangyinz (changes requested):
Looks good to me! I only have two minor comments in the code. By the way, before we can merge this branch, could you please do a rebase to current master branch? We also need to make sure all the new code passes the black formatter.

PR #304, opened 2022-02-25
Improve fileio

pavlis:
This branch implements several important changes that should dramatically improve file-based io performance.  The fundamental strategy was to isolate all io to C++ code on so the readers and writers never had to deal with the "DoubleVector" python construct created on the python side by pybind11.  The new C++ code is in the  fileio.h and fileio.cc in the "io" sections of include and lib respectively.   The readers all use raw fread to the data array buffers and the writers all use fwrite from the same buffers.   I've tested the code with a local python script I wrote, but I think the existing tests should certify the changes work correctly.  The changes I made to Database to use the new C++ functions should not effect the tests.  If not, we'll need to deal with that when we find out if my claim is true. 

There are a couple of issues that came up in this development mspass-team members may want to look at before we merge these changes.  
1.  I first started to implement old-school file locking for the writers since there was a strong possibility in a parallel setting two threads/process could try to write to the same file simultaneously.  However, when I read the unix man pages for functions related to file locking they pretty much said they were archaic.   What I read says the posix standard now requires an intrinsic lock on any file when it is opened in write or append mode.   If my reading of this is wrong, we could have an issue here.  
2. I fixed a piece of Database written by @haruming that I didn't test, but which I think our test suite will verify.   The issue is that I think the original had an unnecessary inefficiency.  If no an inefficiency the change here is a least clearer as to what it does.  Here is the block of code where I left the original commented out in case this breaks:
```
                elif isinstance(mspass_object, Seismogram):
                    # This was previous form.   The toSeismogram run as a
                    # method is an unnecessary confusion and I don't think
                    # settign npts or data are necessary given the code of
                    # Stream2Seismogram - st.toSeismogram is an alias for that
                    #sm = st.toSeismogram(cardinal=True)
                    #mspass_object.npts = sm.data.columns()
                    #mspass_object.data = sm.data
                    mspass_object = Stream2Seismogram(st,cardinal=True)
```
If the tests show this works we should strip those commented out lines before we merge this with master.

### DIFF
--- a/cxx/include/mspass/io/fileio.h
+++ b/cxx/include/mspass/io/fileio.h
@@ -1,0 +1,138 @@
+#ifndef _MSEED_INDEX_H_
+#define _MSEED_INDEX_H_
+#include <string>
+#include "mspass/seismic/TimeSeries.h"
+#include "mspass/seismic/Seismogram.h"
+namespace mspass::io
+{
+/*! \brief Fast file writer for native TimeSeries save to a file.
+
+When saving data to a file system there is no standard way to do so we
+know of that is faster than the low level C fwrite function.  This function
+uses fwrite to write ONLY the sample data in the input TimeSeries object to
+a file specified by a directory (dir) and leaf file name (dfile).  It can
+do so because the std::vector container is required by the standard to define
+a contiguous block of memory. This function is expected to be used in MsPASS
+only under the hood of the python database writer for native saves.
+
+The function constructs a unix path file name as dir+"/"+dfile.  If that
+file does not exist it will be created.  If it does exist the write will
+append to the existing file.   The attributes dir, dfile, and foff are
+always saved in the data's Metadata container so they can be saved to
+the MonogoDB database in MsPASS AFTER calling this writer.  For that reason
+d is not const because it simply isn't.
+
+\param d data to be saved (altered on return as noted above)
+\param dir directory name (assumed to not contain a trailing /)
+\param dfile leaf file name for save
+
+\return file position in output file of the first byte written.
+the same number is saved in the "foff" metadata attribute.
+
+\expection This function will throw a MsPASS error for one of several
+common io related issues. Caller should always include the call to this
+function in a try block.
+
+*/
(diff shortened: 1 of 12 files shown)


pavlis:
I was most assuredly wrong that these changes would pass the test suite.  There is a cascade here showing a lot of errors I cannot reproduce.   I'm hoping one of the other team members can make sense of the errors this is showing.

pavlis:
Well, looking closer a hypothesis - there is some kind of disconnect here when the C++ buffers are handled native they aren't being recast as a "DoubleVector".  Understanding that is way beyond my expertise and something I hope someone else can fix.

pavlis:
I think I figured out what went wrong here.   It was a very subtle issue related to virtual class inheritance - a pretty advanced feature of OOP that enters the picture with multiple inheritance.   An example of the dangers of blindly following a suggestions from the web based on a compilation error.   

There seems to be a residual problem that is more in line with with what I might have expected.  I'm hoping one of you can try to sort the residual problem out since I still don't have pytest running properly on any local system.

I also don't know how to properly fix the minor blemishes the syntax checker finds now.  Do  I just push the "commit suggestion" box for each of the items flagged?

wangyinz:
Now the error is only in the database where we compare the data saved with the original, and it doesn't match now. So it may still relate to the new IO implementation. But for now, I suggest we wait till @Yangzhengtang pushes his branch and integrate the changes here into that branch.

For the syntax checker, the "commit suggestion" is only useful when you have only a few formatting issues. When there are too many, it can not mark all of them at a time, so you will be getting new issues every time you commit the suggested changes. So, it is better to run the formatter locally in this case.

pavlis:
I think the problems have been resolved that were first causing this to break. The problem was not in the new code but some changes I made based on misunderstanding of the older code handling of formatted data.  Don't really understand the reasons the code has extra lines for handling the data buffer but this makes it work.   If any of you understand the reasons there should be comments in the code explaining why the code lines manipulating data for TimeSeries and Seismogram are need and why the are different.

wangyinz:
This is included in #305 already

Issue #308, opened 2022-03-28
Increasing peak memory consumption in CoreTimeSeries::set_npts


Yangzhengtang:
This issue was found when testing aws s3 lambda. I set the npts to an incorrect value (8640000, the npts before being timewindowed) when indexing the s3 lambda miniseeds. Then read_data set the npts of the TimeSeries object in the code below:
https://github.com/mspass-team/mspass/blob/a4a5417cf317adc0cb29527beed6f99a492aa46f/python/mspasspy/db/database.py#L529-L537
And after the windowed mseed file was downloaded from s3 and parsed, the correct npts number (30001) was set here:
https://github.com/mspass-team/mspass/blob/a4a5417cf317adc0cb29527beed6f99a492aa46f/python/mspasspy/db/database.py#L3702-L3703
As I ran the test over 60GB miniseeds, the memory consumption kept increasing and finally I got the std:bad_alloc error, indicating that there was something wrong with the memory, probably a memory leak.

It turns out that the `CoreTimeSeries::set_npts` function is problematic: it allocates the memory for the DoubleVector in the TimeSeries in the code below: 
https://github.com/mspass-team/mspass/blob/2c44d9624db5ed41f9d72a3f3e5b03107d7d4bb7/cxx/src/lib/seismic/CoreTimeSeries.cc#L244-L246
Here, [clear()](https://www.cplusplus.com/reference/vector/vector/clear/)and [reserve()](https://www.cplusplus.com/reference/vector/vector/reserve/) don't change the capacity of the vector, so the memory allocation remains unchanged after these two calls. So when we try to set npts to a smaller number and store a smaller vector, the redundant memory won't be freed, and this would cause a problem in memory utilization. 
_Note: This issue has nothing to do with pybind11's memoryview, because the memory is always managed by the C++ parts._

For now, we plan to fix this by replacing clear() with `vector<T>().swap(x);` to force a reallocation. We might need to check its impact on performance and decide if we need to find another solution later.

Yangzhengtang:
Sorry for the delay... my pytest environment was broken. I didn't push the change until I finally get it working. Anyway, the fix for this issue is [here](https://github.com/mspass-team/mspass/commit/e6f6f00e56f71b97d30841c8366848ee769d430d).
Now whenever set_npts is called, the memory will be reallocated.

Here is a simple performance benchmark:
```python
def memory_test():
    ts = TimeSeries()
    npts = 2
    t0 = time()
    time_and_mem = []
[...]

PR #309, opened 2022-04-04
Fix the memory issue in TimeSeries

Yangzhengtang:


wangyinz (approved):
This fix is discussed in #308. Well done!

PR #310, opened 2022-04-12
Extend normalization options

pavlis:
This branch contains the prototype normalization classes we have been discussing for the past few weeks and for which there are a set of test results found in the discussion pages [here](https://github.com/mspass-team/mspass/discussions/307).

As we discussed this needs some cleanup and some feedback on the design before we put it into the production code base.  What you find in this branch should be viewed as a working prototype that needs the rough edges removed.   

I'm going to past the pieces of the test code I used for the discussion plots below.  The data set I ran this on was the one created by the getting_started tutorial from the 2021 SAGE-GAGE workshop.   As noted in our conference call you could be able to recreate that modest data set in less than an hour by running the jupyter notebook using the web service downloading box. 

Here is a file I used as a working module that contains the code used for the bulk update tests that are a side issue for the development we have prioritized (i.e. finalizing this set of normalization classes):
```
from mspasspy.ccore.utility import MsPASSError,ErrorSeverity
import pandas as pd
import time
from mspasspy.db.normalize import mseed_channel_matcher,mseed_site_matcher

def normalize_mseed(db,
                    wfquery={},
                    blocksize=1000,
                    normalize_channel=True,
                    normalize_site=False,
                    verbose=False):
    """
    In MsPASS the standard support for station information is stored in 
    two collections called "channel" and "site".   When normalized 
    with channel collection data a miniseed record can be associated with 
    station metadata downloaded by FDSN web services and stored previously 
    with MsPASS database methods.   The default behavior tries to associate
    each wf_miniseed document with an entry in "site".  In MsPASS site is a 
    smaller collection intended for use only with data already assembled 
    into three component bundles we call Seismogram objects.  


    For both channel and site the association algorithm used assumes
    the SEED convention wherein the strings stored with the keys 
    "net","sta","chan", and (optionally) "loc" define a unique channel 
    of data registered globally through the FDSN.   The algorithm then 
    need only query for a match of these keys and a time interval 
    match with the start time of the waveform defined by each wf_miniseed 
    document.   The only distinction in the algorithm between site and 
    channel is that "chan" is not used in site since by definition site 
    data refer to common attributes of one seismic observatory (commonly 
    also called a "station").   
    
    :param db: should be a MsPASS database handle containing at least 
    wf_miniseed and the collections defined by the norm_collection list.
    :param blockssize:   To speed up updates this function uses the 
    bulk writer/updater methods of MongoDB that can be orders of 
    magnitude faster than one-at-a-time updates for setting 
    channel_id and site_id.  A user should not normally need to alter this 
    parameter.
    :param wfquery: is a query to apply to wf_miniseed.  The output of this 
    query defines the list of documents that the algorithm will attempt 
    to normalize as described above.  The default will process the entire 
    wf_miniseed collection (query set to an emtpy dict).
    :param normalize_channel:  boolean for handling channel collection. 
    When True (default) matches will be attempted with the channel collection
    and when matches are found the associated channel document id will be 
    set in the associated wf_miniseed document as channel_id.
    :param normalize_site:  boolean for handling site collection. 
    When True (default) matches will be attempted with the site collection
    and when matches are found the associated site document id will 
    be set wf_miniseed document as site_id.  
    :param verbose: When set true the database methods for matching the 
    net:sta:chan:loc:time keys will be run in verbose mode.  Those database 
    methods will print a diagnostic for all ambiguous matches.  Because 
    this function is expected to be run on potentially large raw data sets of 
    miniseed inputs the default is False to reduce the overhead of potentially 
    large log messages created by the all to common duplicate metadata problem.
    Users are encouraged to verify the channel and site collections have 
    no serious problems with ambiguous net:sta:loc(chan) that are truly 
    inconsistent (i.e. have different attributes for the same keys)
    
    :return: tuple with two integers.  0 is the number of documents processed in 
    wf_miniseed (output of query) and 1 is the number with ids set.   Note this 
    is the number where any id was set.  The algorithm does not count site and 
    channel individually.
    """
    # this is a prototype - use all defaults for initial test
    matcher = mseed_channel_matcher(db)
    sitematcher = mseed_site_matcher(db)
    ndocs = db.wf_miniseed.count_documents(wfquery)
    if ndocs == 0:
        # check me - may be wrong number of args
        raise MsPASSError("normalize_mseed","query of wf_miniseed yielded 0 documents\nNothing to process",ErrorSeverity.Fatal)
    # An immortal cursor should not be necssary for this algorithm
    cursor = db.wf_miniseed.find(wfquery)
    counter = 0
    number_set = 0
    dt1=0.0
    dt2=0.0
    dt3=0.0
    #bulk = db.wf_miniseed.initialize_unordered_bulk_op()
    bulk = db.wf_miniseed.initialize_ordered_bulk_op()
    for doc in cursor:
        wfid = doc["_id"]
        stime = doc["starttime"]
        t=time.time()
        if normalize_channel:
            chandoc = matcher.get_document(doc,time=stime)
        if normalize_site:
            sitedoc = sitematcher.get_document(doc,time=stime)
        dt1 += time.time() - t
        # signal of no match is returning None
        update_dict = dict()
        if normalize_channel:
            if chandoc != None:
                update_dict["channel_id"] = chandoc["_id"]
        if normalize_site:
            if sitedoc != None:
                update_dict["site_id"] = sitedoc["_id"]
        if len(update_dict) > 0:
            t = time.time()
            bulk.find({"_id" : wfid}).update({"$set" : update_dict})
            dt2 += time.time() - t
            counter += 1
        if (counter % blocksize == 0):
            t = time.time()
            bulk.execute()
            dt3 += time.time() - t
            #bulk = db.wf_miniseed.initialize_unordered_bulk_op()
            bulk = db.wf_miniseed.initialize_ordered_bulk_op()
            number_set += blocksize
            
    if (counter % blocksize != 0):
        t = time.time()
        bulk.execute()
        dt3 += time.time() - t
    number_set += counter%blocksize
    
    print("Elapsed time fetching normalization channel and site documents=",dt1)
    print("Elapsed time runnin bulk.find function=",dt2)
    print("Elapsed time executing bulk updates=",dt3)
    return [ndocs,number_set]

def normalize_single(db,
                     locksize=1000,
                    normalize_channel=True,
                    normalize_site=True,
                    verbose=False):
    """
    temporary function for running single update comparison
    """
    matcher = mseed_channel_matcher(db)
    sitematcher = mseed_site_matcher(db)
    cursor = db.wf_miniseed.find({})
    counter = 0
    for doc in cursor:
        wfid = doc["_id"]
        stime = doc["starttime"]
        if normalize_channel:
            chandoc = matcher.get_document(doc,time=stime)
        if normalize_site:
            sitedoc = sitematcher.get_document(doc,time=stime)
       
        # signal of no match is returning None
        update_dict = dict()
        if normalize_channel:
            if chandoc != None:
                update_dict["channel_id"] = chandoc["_id"]
        if normalize_site:
            if sitedoc != None:
                update_dict["site_id"] = sitedoc["_id"]
        if len(update_dict) > 0:
            db.wf_miniseed.update_one({"_id" : wfid},{"$set" : update_dict})
            counter += 1
    return counter
        
def df_load_collection(dbcol,keys_to_load=["net","sta","chan","starttime","endtime","_id"],
                       load_if_defined=["loc"]):
    """
    this is a test prototype to load an entire collection into a pandas dataframe.
    This is use to test hypothesis that the performance of bulk updates above 
    was limited by normalization queries to channel and site. This function 
    will be used below to load the entire site and channel collections into 
    a dataframe.
    """
    # operator + for a list appends contents of second to first
    column_keys = keys_to_load + load_if_defined

    # Not sure if this is the fastest way to load up a dataframe but 
    # is a common approach in web examples
    data = dict()
    for key in column_keys:
        data[key] = []
        
    cursor = dbcol.find({})
    for doc in cursor:
        # this prototype does not testing - a careful implementation needs
        # sanity checks here
        for key in keys_to_load:
            val = doc[key]
            data[key].append(val)
        # this is not a generic solution -works for loc null= ""
        for key in load_if_defined:
            if key in doc:
                val = doc[key]
                data[key].append(val)
            else:
                data[key].append("")
    
    df = pd.DataFrame(data,columns=column_keys)
    return df
    
def df_find_channel(df,net,sta,chan,time,loc=None):
    if loc == None:
        result = df[ (df["net"] == net) & (df["sta"] == sta) & (df["chan"] == chan) & (df["starttime"] <= time) & (df["endtime"] >= time)]
    else:
        result = df[ (df["net"] == net) & (df["sta"] == sta) & (df["chan"] == chan) & (df["loc"] == loc)
                    & (df["starttime"] <= time) & (df["endtime"] >= time)]
    return result

def df_normalize_mseed(db,

                    wfquery={},
                    blocksize=1000,
                    normalize_channel=True,
                    normalize_site=True,
                    verbose=False):
    """
    Like above but loads a panda dataframe instead of using db query for
    channel.  Ignore site for now.
    """
    ndocs = db.wf_miniseed.count_documents(wfquery)
    if ndocs == 0:
        # check me - may be wrong number of args
        raise MsPASSError("normalize_mseed","query of wf_miniseed yielded 0 documents\nNothing to process",ErrorSeverity.Fatal)
    df = df_load_collection(db.channel)
   
    # An immortal cursor should not be necssary for this algorithm
    cursor = db.wf_miniseed.find(wfquery)
    counter = 0
    number_set = 0
    #bulk = db.wf_miniseed.initialize_unordered_bulk_op()
    bulk = db.wf_miniseed.initialize_ordered_bulk_op()
    for doc in cursor:
        wfid = doc["_id"]
        net = doc["net"]
        sta = doc["sta"]
        chan = doc["chan"]
        stime = doc["starttime"]
        if "loc" in doc:
            loc = doc["loc"]
            dfmatch = df_find_channel(df,net,sta,chan,time=stime,loc=loc)
            if len(dfmatch)==0:
                print("Warning:  no match for ",net,sta,chan,loc,stime)
        else:
            dfmatch = df_find_channel(df,net,sta,chan,time=stime)
            if len(dfmatch)==0:
                print("Warning:  no match for ",net,sta,chan,stime)
        # signal of no match is returning None
        update_dict = dict()
        if normalize_channel:
            if len(dfmatch) > 0:
                # there is likely a better way to do this - I am hacking
                i = dfmatch.index[0]
                update_dict["channel_id"] = df.at[i,'_id']

        if len(update_dict) > 0:
            bulk.find({"_id" : wfid}).update({"$set" : update_dict})
            counter += 1
        if (counter % blocksize == 0):
            bulk.execute()
            #bulk = db.wf_miniseed.initialize_unordered_bulk_op()
            bulk = db.wf_miniseed.initialize_ordered_bulk_op()
            number_set += blocksize
            
    if (counter % blocksize != 0):
        bulk.execute()
    number_set += counter%blocksize
    return [ndocs,number_set]
```
Note there is a bunch of debris in there that are relics like all the dataframe stuff.  I include it to reduce the chances of accidentally deleting something you need to duplicate what I did. 

Here is the crude script I used to drive the tests in the discussion page:
```
from mspasspy.db.client import DBClient

dbclient = DBClient()
db=Database(dbclient,"SG2021")
# Build necessary indices for channel and site.  
# these lines can be commented out after being run once
db.channel.create_index(
  [
    ('net',1),
    ('sta',1),
    ('chan',1)
  ]
)

db.site.create_index(
  [
    ('net',1),
    ('sta',1)
  ]
)


from normalize_using_matchers import (normalize_mseed,
                         normalize_single,
                         df_load_collection,
                         df_find_channel,
                         df_normalize_mseed)
import time

# temporary for debug
#from obspy import UTCDateTime
#testtime = UTCDateTime("2004-08-25T23:16:40.000000Z")
#dbcol = db.channel
#df = df_load_collection(dbcol)
#testtime = UTCDateTime("2004-08-25T23:16:40.000000Z")
#dftest = df_find_channel(df,"AZ","BZN","HHE",testtime)

timing_results=[]
t0=time.time()
n = normalize_single(db)
t=time.time()
print("update_one number of updates=",n)
print("elapsed time=",t-t0)
print("average update time=",(t-t0)/n)
timing_results.append([1,t-t0,n,(t-t0)/n])
bsize = [10, 100, 1000, 10000]
for blocksize in bsize:
    t0=time.time()
    result = normalize_mseed(db,blocksize=blocksize)
    #result = df_normalize_mseed(db,blocksize=blocksize)
    t=time.time()
    print(result)
    print("elapsed time=",t-t0)
    print("Average update time=",(t-t0)/result[1])
    n = result[1]
    timing_results.append([blocksize,t-t0,n,(t-t0)/n])

bv=[]
dtv=[]
dtper=[]

for i in range(len(timing_results)):
    b = timing_results[i][0]
    dt = timing_results[i][1]
    dtp = timing_results[i][3]
    bv.append(b)
    dtv.append(dt)
    dtper.append(dtp)
    
import matplotlib.pyplot as plt
plt.semilogx(bv,dtper)
plt.show()
```

As usual some hacking might be needed to make that work, but hopefully it will allow you to recreate what I did.


pavlis:
I had a thought about the design of this set of normalization function objects.   In the same way we implemented a "FiringSquad" in edit.py and "MetadataOperatorChain" to the set of metadata editors (in the same edit.py module) we should have a chain class that would run a chain of these normalization operators.    I realized that need when I started to think about the design for a command line tool for which the only objective was to find and set the mspass standard way to normalize up till now - that is setting attributes like "channel_id", "site_id", and "source_id".   With caching options and the established performance advantages of bulk updates compared to mongodb update_one, it would make sense to assemble all the ids one wanted to set first, then do updates of collections like wf_miniseed, wf_TimeSeries, or wf_Seismogram.  Similarly, a dask/spark map operation with a chain operator would likely be more efficient than multiple map calls, although that is a pure guess/hypothesis.  Both would be easily implemented with an operator chain like that we built for trace editing and metadata math/edits.

pavlis:
I am not sure of the status of this pull request.   I thing Zhengtang has been cleaning this up and producing standard tests.  I didn't want to collide with his changes so I'm not doing this by pushing a revision of this branch.  The thing I did that I don't think he was planning to do was write a short command line tool to normalize the wf_miniseed collection with our standard use of channel_id.   I went ahead an implemented the code below and ran it on my test "SG2021" database - output of the getting started notebook from sage-gage 2021.  It ran and did what it was supposed to do.   

Incidentally, this has a small change from the code I used for doing bulk write update tests.   I'm not sure why, but I suddenly got warnings about deprication of the  `initialize_bulk_op` lines.   Documentation confirmed that was needed.  The new api is 
much simpler.  I left the old code in below but commented it out.

We can hopefully finalize this in our meeting today, but recommend Zhengtang be tasked to merge this with this branch.   Note this file should probably be broken apart.  I recommend the `normalize_mseed` function be put into the normalize.py module  (mspasspy.db.normalize).   The main (command line tool) would belong in mspaspy/db/scripts.   I called in normalize_mseed as the main is just a command line interface to the python function - pretty much our standard way of doing that kind of thing.
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

"""
This file contains a python script that defines a command line
tool we call normalize_mseed.   It is intended to be a used as a 
preprocessing tool to "normalize" the contents of the wf_miniseed 
collection.   The idea of the tool is that miniseed data, by
definition (the mini of miniseed means part of the see format) is 
incomplete and lacks station metadata.   The only thing miniseed 
data have are the four seed station code strings (net,sta, chan, and loc)
and time tags for each data packet.   The wf_miniseed collection is 
used to produce an index into a set of miniseed files that are 
presumed to define a set of waveform time segments.  This tool 
will try to create the entries channel_id and (optionally) site_id 
for each record in wf_miniseed.   It issues a report to stdout 
on completion that give the number of documents set and the number 
of documents for which no match could be found in the channel (or site)
collections.

"""
import argparse
import sys
from pymongo import UpdateOne
from mspasspy.db.client import DBClient
from mspasspy.db.normalize import mseed_channel_matcher,mseed_site_matcher
from mspasspy.ccore.utility import MsPASSError,ErrorSeverity

def normalize_mseed(db,
                    wfquery={},
                    blocksize=1000,
                    normalize_channel=True,
                    normalize_site=False,
                    verbose=False):
    """
    In MsPASS the standard support for station information is stored in 
    two collections called "channel" and "site".   When normalized 
    with channel collection data a miniseed record can be associated with 
    station metadata downloaded by FDSN web services and stored previously 
    with MsPASS database methods.   The default behavior tries to associate
    each wf_miniseed document with an entry in "site".  In MsPASS site is a 
    smaller collection intended for use only with data already assembled 
    into three component bundles we call Seismogram objects.  


    For both channel and site the association algorithm used assumes
    the SEED convention wherein the strings stored with the keys 
    "net","sta","chan", and (optionally) "loc" define a unique channel 
    of data registered globally through the FDSN.   The algorithm then 
    need only query for a match of these keys and a time interval 
    match with the start time of the waveform defined by each wf_miniseed 
    document.   The only distinction in the algorithm between site and 
    channel is that "chan" is not used in site since by definition site 
    data refer to common attributes of one seismic observatory (commonly 
    also called a "station").   
    
    :param db: should be a MsPASS database handle containing at least 
    wf_miniseed and the collections defined by the norm_collection list.
    :param blockssize:   To speed up updates this function uses the 
    bulk writer/updater methods of MongoDB that can be orders of 
    magnitude faster than one-at-a-time updates for setting 
    channel_id and site_id.  A user should not normally need to alter this 
    parameter.
    :param wfquery: is a query to apply to wf_miniseed.  The output of this 
    query defines the list of documents that the algorithm will attempt 
    to normalize as described above.  The default will process the entire 
    wf_miniseed collection (query set to an emtpy dict).
    :param normalize_channel:  boolean for handling channel collection. 
    When True (default) matches will be attempted with the channel collection
    and when matches are found the associated channel document id will be 
    set in the associated wf_miniseed document as channel_id.
    :param normalize_site:  boolean for handling site collection. 
    When True (default) matches will be attempted with the site collection
    and when matches are found the associated site document id will 
    be set wf_miniseed document as site_id.  
    :param verbose: When set true the database methods for matching the 
    net:sta:chan:loc:time keys will be run in verbose mode.  Those database 
    methods will print a diagnostic for all ambiguous matches.  Because 
    this function is expected to be run on potentially large raw data sets of 
    miniseed inputs the default is False to reduce the overhead of potentially 
    large log messages created by the all to common duplicate metadata problem.
    Users are encouraged to verify the channel and site collections have 
    no serious problems with ambiguous net:sta:loc(chan) that are truly 
    inconsistent (i.e. have different attributes for the same keys)
    
    :return: tuple with three integers.  0 is the number of documents processed in 
    wf_miniseed (output of query), 1 is the number with channel ids set, 
    and 2 contains the number of site documents set.  1 or 2 should 
    contain 0 if normalization for that collection was set false.
    """
    # this is a prototype - use all defaults for initial test
    matcher = mseed_channel_matcher(db,attributes_to_load=["_id","net","sta","chan","starttime","endtime"])
    if normalize_site:
        sitematcher = mseed_site_matcher(db,attributes_to_load=["_id","net","sta","starttime","endtime"])
    ndocs = db.wf_miniseed.count_documents(wfquery)
    if ndocs == 0:
        # check me - may be wrong number of args
        raise MsPASSError("normalize_mseed","query of wf_miniseed yielded 0 documents\nNothing to process",ErrorSeverity.Fatal)
    # An immortal cursor should not be necssary for this algorithm
    cursor = db.wf_miniseed.find(wfquery)
    counter = 0
    number_channel_set = 0
    number_site_set = 0

    # this form was used in older versions of MongoDB but has been 
    # depricated in favor of the simpler bulk_write  
    # commented out lines using the symbol bulk are the old form
    # revision sets bulk to a simple list of instructions ent to bulk_write
    #bulk = db.wf_miniseed.initialize_unordered_bulk_op()
    #bulk = db.wf_miniseed.initialize_ordered_bulk_op()
    bulk = []
    for doc in cursor:
        wfid = doc["_id"]
        stime = doc["starttime"]
        if normalize_channel:
            chandoc = matcher.get_document(doc,time=stime)
        if normalize_site:
            sitedoc = sitematcher.get_document(doc,time=stime)
        # signal if no match is returning None so don't update in that situation
        update_dict = dict()
        if normalize_channel:
            if chandoc != None:
                update_dict["channel_id"] = chandoc["_id"]
                number_channel_set += 1
        if normalize_site:
            if sitedoc != None:
                update_dict["site_id"] = sitedoc["_id"]
                number_site_set += 1
        # this conditional is needed in case neither channel or site have a match
        if len(update_dict) > 0:
            #bulk.find({"_id" : wfid}).update({"$set" : update_dict})
            bulk.append(UpdateOne({"_id" : wfid},{"$set" : update_dict}))
            counter += 1
        if (counter % blocksize == 0):
            #bulk.execute()
            #bulk = db.wf_miniseed.initialize_unordered_bulk_op()
            #bulk = db.wf_miniseed.initialize_ordered_bulk_op()
            db.wf_miniseed.bulk_write(bulk)
            
    if (counter % blocksize != 0):
        #bulk.execute()
        db.wf_miniseed.bulk_write(bulk)
    
    return [ndocs,number_channel_set, number_site_set]

def main(args=None):
    """
    Main function for normalize_mseed command line tool.   
    All this function does is parse the command line to build the 
    args sent to the function called "normalize_mseed" defined above. 
    Note that function may need to be moved to the normalize python 
    module.  
    """

    if args is None:
        args = sys.argv[1:]
    parser = argparse.ArgumentParser(
        prog="normalize_mseed",
        usage="%(prog)s dbname [--normalize_site --blocksize n -wfquery querystring]",
        description="MsPASS program to create cross-referencing keys from wf_miniseed to channel",
    )
    parser.add_argument(
        "dbname", metavar="dbname", type=str, help="MongoDB database name to be normalized"
    )
    parser.add_argument(
        "--normalize_site",
        action="store_true",
        help="Normalize site collection as well as channel collection (default is false)",
    )
    parser.add_argument(
        "--blocksize",
        type=int,
        default=1000,
        help="Set number of entries to cache for each bulk update call for MongoDB"
    )
    parser.add_argument(
        "--wfquery",
        type=str,
        default="",
        help="Optional query to apply to wf_miniseed collection"
        
    )
    
    args = parser.parse_args(args)
    dbname = args.dbname
    print("Normalizing database with name=",dbname)

    normalize_site = args.normalize_site
    if normalize_site:
        print("Will normalize all matching wf_miniseed documents site_id and channel_id")
    else:
        print("Will normalize all matching wf_miniseed documents with channel_id")
    blocksize = args.blocksize
    if len(args.wfquery) > 0:
        query = eval(args.query)
    else:
        query = {}
    
    dbclient = DBClient()
    db=Database(dbclient,dbname)
    
    ret_tuple = normalize_mseed(db,
                                wfquery=query,
                                blocksize=blocksize,
                                normalize_site=normalize_site)
    print("Normalization completed for wf_miniseed collection of database = ",dbname)
    if len(args.wfquery) > 0:
        print("Applied with this query to wf_miniseed:  ",args.query)
    print("Number of documents processed=",ret_tuple[0])
    print("Number of channel_ids set=",ret_tuple[1])
    print("Number of site_ids set=",ret_tuple[2])
        
if __name__ == "__main__":
    main()

```

python/mspasspy/db/normalize.py
```diff
@@ -0,0 +1,2052 @@
...
+            # (and usually kill it) and land here. Allows application in a map
+            # operation
+            return d
+        else:
+            raise TypeError(
+                "css30_arrival_interval_matcher.normalize:  received invalid data type"
+            )
+
+
+def bulk_normalize(
```
pavlis, line 1819: You extended this to be much more generic.   Has it been tested on anything but the default?   Docstring needs some clarification on usage.   My prototype was frozen to only handle what here is the default.

pavlis: I looked at this again this morning and realized I misunderstood what this class was doing.   It is not a generic version of normalize_mseed but a method to do apply mulitple normalizers in a single pass.  I retract my comments on this class and say instead the docstring needs to be expanded to better clarify its use.

Yangzhengtang: Hey Gary, I just read through your comments, thanks for pointing them out. I will add the missing docstrings. 

Also, in this week's meeting, Ian and I went through the code and decided to make some revisions. Since some different derived classes of NMF share lots of duplicated codes (for example, mseed_channel_matcher and mseed_site_matcher), It might be better to extract the duplicated parts, and put them in the NMF or another intermediatory class, to avoid the duplication and complexity. I am currently working on it, and I will push the update very soon.

python/mspasspy/db/normalize.py
```diff
@@ -0,0 +1,2052 @@
...
+    It is recommended unless the memory foot print is excessive.
+    That too can usually be avoided by using a query to weed out unnecessary
+    channel documents or by editing the channel document to reduce the
+    debris from extraneous data.
+    """
+
+    def __init__(
+        self,
+        db,
+        attributes_to_load=None,
```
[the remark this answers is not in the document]
Yangzhengtang: We need to set the default value of a list/dict argument to None, and initialize it in the function body. That's because the weired feature of python: https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument. 
And I totally agree that we should clarify it in the docstring, I will do that. Thanks for pointing that out.

python/mspasspy/db/normalize.py
pavlis: Makes a lot of sense.  That is a classic use of a base class.

Yangzhengtang:
I just pushed several commits for this branch. As discussed in the last week's meeting, most of the commits were trying to make the code more concise. They contain the following changes. 

1. Extract duplicated codes in NMF subclasses, and put them in the NMF classes.

2. Added 2 new intermediate classes, as another layer of abstraction. 
The inheritance relationship is now as follows:
NMF -> single_key_matcher -> ID_matcher
NMF -> composite_key_matcher-> mseed_channel_matcher
NMF -> composite_key_matcher-> mseed_site_matcher
NMF -> origin_time_source_matcher
NMF -> css30_arrival_interval_matcher

3. Change the interface of all NMF classes to let get_document always return a Metadata (Unifying the return types of  _db_get_document() and _cached_get_document())

4. Change the xref mechanism in mseed_channel_matcher and mseed_site_matcher (composite_key_matchers). The old way for caching in these two classes was to create a cache (self.cache) for all original documents, and another indirection (self.xref) to map composite keys to the corresponding _id of documents. In the new implementation, we don’t use the xref, instead just map composite keys directly to lists of matched documents. This dictionary is still called self.cache, but we handle it in a different way by overriding related methods.
By using the new mechanism, we can expect 1) the runtime speed will be faster, because we don’t need another direction 2) the code will be simpler and easier to maintain

5. Change the mechanism of mseed_channel_matcher and mseed_site_matcher (composite_key_matchers). They will now cache the keys required for creating composite keys even if these keys are not defined in the attribute_to_load. 
(Consider the situation where a mseed_channel_matcher has attribute_to_load = [“_id”], and the caching is enabled. Since "net", "sta", "chan" are required for creating the composite keys, but they are not present in the attribute_to_load, an error will be thrown in the old implementation.)
The new implementation can cover the situation above. Also, “starttime” and “endtime” are also always cached, for the similar reason.

6. Change the normalize_mseed(), it is now implemented with the bulk_normalize(), its functionality and behavior are not changed. 

7. Minor changes on the code structure, some codes are extracted and saved to some shared methods to avoid duplication, for example NMF._load_doc(), composite_key_matcher._get_test_time()... These changes won’t have an effect on the actual functionality and behavior of classes and other methods, they are just intended to make the codes more compact and simple.

8. Minor changes on the log_error() method, to avoid using redundant arguments.

9. Minor change on composite_key_matcher: when multiple documents are matched, we should add an warning log before returning the first one

10. Add/modify docstrings

pavlis:
I looked through your changes @Yangzhengtang  and see no issues.   Very nice work.

One detail I would point out for your education.  @wangyinz  I think the changes are excellent should be accepted.   The point I want to make for you @Yangzhengtang is that adding a layer of inheritance is best done as a design choice up front rather than added as a mechanism to reduce the code base.   In the design stage one asks what are common concepts that a intermediate level like this can enscapsulate?   In fact, you did just that but didn't articulate that as your motivation.   Maybe that was your thinking, and if so ignore this.  Specifically the common concepts you capture are:  (1) single key matching, and (2) multiple key matching.   I applaud your insight to realize these were the common denominator you could fold into the design.   They provide a useful scaffolding on which other other future matching classes could be created.   Well done!  The only lesson is the path I think that took you there.   I OOP design the books all say (correctly) that before you start coding you should think through the common concepts and how they are best linked in any inheritance structure.    You did that in a way by taking my prototype and reworking it into this much improved structure.  Again, is more the starting point in your thinking about how to improve the prototype.  

Don't misunderstand my lecture.  Brilliant work.  We are blessed to have you on this team.

Yangzhengtang:
Great point, Gary! I totally agree. I guess I might have some of the OOP thinking when I started, but I made those changes mostly out of consideration of the size of code base, you are right. And I really appreciate the lesson you taught me.

PR #313, opened 2022-05-09
Improve window

pavlis:
This branch has only a small change to improve the efficiency of the window functions.  The previous had a built in inefficiency when windowing long signals using a copy constructor from the parent. Changes now only create a workspace of the needed size for the returns.  There is a cost as the code is much more fragile now as it has some dependencies on implementation details.  I left a MAINTENANCE node in comments to help all be aware of that issue in the future


pavlis:
Ouch, the new code seg faults.  Suspect I know why, but will take some debugging to find out if my hypothesis is correct.

pavlis:
I fixed the window code but when I look over this branch I think it got tangled up with some experimental changes in the file mspasspy/db/matcher.py.   I think it may not be too hard to untangle as I am pretty sure all the relevant changes are in slice_and_dice.cc.   Not sure how to proceed.

PR #315, opened 2022-05-28
Normalize bugfix

pavlis:
This branch has small changes from the current master to fix two bugs found in the normalize module:
1.  Previous had a serious mistake in bulk_normalize where the list container of update instructions passed to bulk_write was not being cleared between calls to bulk_write.  update list then grew and grew and created an awful redundant update.  Was not found until this was run on a large dataset where the size could be come overwhelming.
2. The previous version did not handle setting normalizing id keys correctly for anything.  The base class had the error and bulk_normalize had the same mistake.   The net result was normalization keys like "channel_id" where given a double underscore to become "channel__id" (hard to even see).  After fixing that bug I removed a hack fix in normalize_mseed to fix that problem.

I could not fully test my changes.  I am confident they work, however, as I made the same changes inside the container I was running on the large test that found these bugs.   Unless I made a typo this should work so I urge you to simply merge this bugfix into the master - assuming all the pytest things complete without error.   I think that unlikely as I know there are no syntax errors and nothing I did is likely to break the existing tests unless the tests were expecting the double _ bug, which I doubt.  I didn't check.


pavlis:
Hmmm - I clearly need to dig into the tests.  The test for bulk normalize fails.   Let me work on this.

pavlis:
I do not understand why the code-format check failed.   @wangyinz can you check the log?   I don't know why it gives the message it does?

pavlis:
I just realized there are a bunch of suggestions for picky errors that can no doubt just be accepted.  I may have created those by pulling into a branch that was forked from an earlier version of master.   I had some conflicts in database.py I had to resolve manually.  The picky errors may have been autofixed in the pull retaining my old version.   A guess.   I'm going to let @wangyinz resolve those.   Are those the things causing core-format to error?

wangyinz:
hmmm... it seems something has seriously messed up. Why are there changes from March in this pull request?

wangyinz:
Can you point me to the commit that's actually fixing the bugs? There are too many irrelevant changes

pavlis:
https://github.com/mspass-team/mspass/pull/315/commits/c8466124bede5c1cb893759bb8f8ab33092121fb

Incidentally, I did one thing unconnected to the bug that I thought was confusing.   Zhengtang had a src_col variable that I changed to wf_col.   It refers to a collection name, and I thought src_col was confusing.  wf_col implies, correctly, that is normally a “wf” collection.

While I have you a different point.  Struggling with this documentation.   I want to spend some time in Monday’s meeting clarifying a few things.   I want to use a variation of Zhengtang’s distributed_node script for the tutorial section I’ve been struggling with.
gary

On May 28, 2022, at 11:01 AM, Ian Wang ***@***.******@***.***>> wrote:

This message was sent from a non-IU address. Please exercise caution when clicking links or opening attachments from external sources.



Can you point me to the commit that's actually fixing the bugs? There are too many irrelevant changes

wangyinz:
The fix will be merged in #316

PR #318, opened 2022-06-02
Bug fix for miniseed reader problem

pavlis:
This branch should fix a bug created, it turns out, when I restructured some things to improve performance in memory reallocation of C++ objects.   The issue was in Metadata driven constructors for both TimeSeries and Seismogram objects.  Neither handled an undefined npts in metadata correctly.  I believe these changes make the two behave the same and sensibly.  In both cases if the Metadata driven constructor finds the field for npts undefined it defaults it to 0.     In all cases the result will silently marked dead.   I also added test in in the test_ccore.py function to make sure this behavior is standardized.


wangyinz:
will be merged by #319

PR #326, opened 2022-08-15
New normalize api

pavlis:
This is the new revised implementation of the normalize module.   It includes a revision of the tests to match the new api.   @wangyinz note I wasn't able to run pytest completely because of a problem in my local setup that seems to lack some data files needed for the tests I made by just hacking the existing test_normalize.py file.   Point is I am nearly positive this initial pull request will generate a ton of errors when pytest is run.   @wangyinz  we have a meeting today so the contents of this pull request can work a discussion today of any major issues anyone else in the group can see with my implementation.


pavlis:
Hey guys, I was working on the documentation for this module in the User's Manual.   Realized we it might be sensible to implement some generic testers similar to EqualityMatcher that would do three other common arithmetic tests:  gt/ge, lt/le, and intervals.   

On the other hand, there is another simple way anyone could do something like that with extensions to Metadata.   e.g. one could write a little python function  to test if some time stamp was in 
the time span of a wavform and set a boolean something like this (untested just a demonstration):
```
def test_range(d,key_to_test,outkey="example_bool"):
   x = d[key_to_test]
   if x>=d.t0 and x<=d.endtime():
    d[outkey] = True
  else:
    d[outkey] = False
  return d
```

Similar functions could compute some value from one more Metadata values using one of the header math processors and do tests on the composite.   Come to think of it that is an example where 
additional arithmetic tests would be most useful. 

What do you think about this idea?   All would be pretty easy to implement with a dataframe.  Hardest part is a proper design of the API.

wangyinz:
That make sense. I guess we can implement a customizable matcher, which can take a function as the input and the function returns a boolean. Then, we only need to implement those arithmetic tests in functions, and use them to define the corresponding matcher with the customizable matcher. This should be a more flexible design. What do you think?

pavlis:
@wangyinz I'm not sure what you are suggesting about a function returning a boolean and how that would interact with the DataframeCacheMatcher, which is probably what we should aim to utilize.   

Also on the negative side I was thinking about this further and I think my suggestion may be misplaced.   The context here is "normalization" with a (normally small) collection/table.   I can't actually think of an example where using a generic inequality test would be any easier to do than just implementing a new custom class in python.   Extending `DataframeCacheMatcher` requires only implementing a custom `subset` method and range testing is a standard thing you find in almost every tutorial on pandas.   We might consider a minor revision to the branch I just submitted with the new User Manual page on normalization saying that for an example on how to implement a matcher with a range test see the class `MiniseedMatcher` (I think that was the name - didn' verify).   Hence, I now think we should dismiss the idea I posed above until a need surfaces that would show a need for a more generic solution.

PR #330, opened 2022-09-03
Update SlownessVector.h

pavlis:
Add the missing friend declaration.   Always required for boost serialization IF the object is used in an std container.  Found this from mysterious downstream errors when trying to create a pickle interface in a pwmig class.  A very subtle error that was not obvious.


pavlis:
I do not understand why a one line change in a C++ include file that only a declaration can cause the tests suite to fail.  I am pretty certain the reasons for the failure have nothing to do with this one line change.

wangyinz:
I changed all the tests that download waveforms so that they all get mocked to local files. Will merge this branch soon.

Issue #320, opened 2022-07-02
parallel reading of ensembles


pavlis:
I think we have two problems we need to deal with for reading large ensembles in parallel.   This should perhaps be put into two issues pages, but they are closely enough related I elected to put them together.   The two issues are:
1.   There may be an issue with serialization of MongoDB Cursor objects.  
2.   Any handling of large ensembles is prone to a memory error.    

For item 1, the evidence is this.   Running the test we discussed on the huge lower 48 broadband array from 2012 I used this parallel construct:

```
import dask.bag as daskbag
dataset = daskbag.from_sequence(srcidlist)
dataset = dataset.map(srcid_query)
dataset = dataset.map(db.wf_miniseed.find)
#dataset.map(db.read_ensemble_data,collection='wf_miniseed',normalize=normalize_list)
#dataset = dataset.map(window_ensemble)
#dataset = dataset.map(bundle_seed_data)
#dataset = dataset.map(lambda ens : len(ens.member))
x=dataset.compute()
print(x)
```
where I commented out the lines that do real processing to limit the workflow to only the call to MongoDB find.   (A number of things here are defined earlier in the notebook.)   Note I know the syntax is right and the data are intact as a serial test on few ensembles runs.   When I run the above I get this error:
```
---------------------------------------------------------------------------
KilledWorker                              Traceback (most recent call last)
<ipython-input-6-77a914a1235f> in <module>
     12 #dataset = dataset.map(bundle_seed_data)
     13 #dataset = dataset.map(lambda ens : len(ens.member))
---> 14 x=dataset.compute()
     15 print(x)

/usr/local/lib/python3.6/dist-packages/dask/base.py in compute(self, **kwargs)
    281         dask.base.compute
    282         """
--> 283         (result,) = compute(self, traverse=False, **kwargs)
    284         return result
    285 

/usr/local/lib/python3.6/dist-packages/dask/base.py in compute(*args, **kwargs)
    563         postcomputes.append(x.__dask_postcompute__())
    564 
--> 565     results = schedule(dsk, keys, **kwargs)
    566     return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
    567 

/usr/local/lib/python3.6/dist-packages/distributed/client.py in get(self, dsk, keys, workers, allow_other_workers, resources, sync, asynchronous, direct, retries, priority, fifo_timeout, actors, **kwargs)
   2652                     should_rejoin = False
   2653             try:
[...]

PR #331, opened 2022-09-08
implement foldby method to assemble ensembles

wangyinz:
This is the attempt to simplified the process of building what is usually called station gathers or event gathers from a list of timeseries or seismograms. The details of this capability is discussed in #241 and #320.


pavlis:
Not sure where you are on this implementation, but it seems a step forward for ensembles.  I wonder if we should consider implementing (C++ dialect) operator += and operator+ in C++ for our ensemble objects.   That would, I think, simply and probably improve the performance of fold (dask) as I think you could make the combine function little more than a call to operator+.   Operator +, which is trivial to implement i a standard way if you implement opeator+=, I am thinking of would not be a stack, which might be very confusing.  Let me clarify.

Operator + abstracts c=a+b where in this case a, b, and c would all be ensembles.   I think the right abstraction in this case is a merge comparable to what that operation means for a string (e.g. "I am a " + "string" = "I am a string".   For the case of an ensembe the "member" array of c would be the concatenation of the contents of a and b.   

The other issue is what to do with the ensemble Metadata container and the ErrorLogger?   I would suggest that Metadata be merged and the operator should check for matching keys and log an error if there are differences i the values being merged.   ErrorLogger data could just use opeator+= for ErrorLogger. 

The worst thing about this idea is that I think it breaks our object level history mechanism since ensembles were not originally considered atomic.   I wonder if this suggests we should add a ProcessingHistory attribute to LoggingEnsemble?   

I fear this mostly muddied the water on a straightforward set of changes.   Seems a topic for discussion by the group outside this forum.

wangyinz:
The implementation and tests are done. However, this does not handle object level history at all because of exactly the complexity you mentioned. I think it make sense to add processing history to the ensemble class instead of casting those info the each individual members. But I don't think this is critical at the moment because we may have different implementations of ensemble class (a matrix version) soon. So, maybe it is OK to merge this branch and use it as a temporary solution anyway.

For the + operator, it is actually a bit more complex than what you described above.  To make the reduce step in the foldby operation work, we need to define a few variant of the operator such that 
```
Seismogram + Seismogram = Ensemble
Seismogram + Ensemble= Ensemble
Ensemble + Seismogram = Ensemble
Ensemble + Ensemble = Ensemble
Seismogram + self_defined_zero = Ensemble
self_defined_zero + Seismogram = Ensemble
Ensemble + self_defined_zero = Ensemble
self_defined_zero + Ensemble = Ensemble
```
Current implementation in this branch uses a temporary list as the object to hold the result and then convert the list to ensemble in the end. It is a much cleaner implementation, and I don't think the performance will be much slower. It is also easier to expand to a matrix version of the ensemble class later.

pavlis:
Hadn't thought about the need for overloading operator + like that.   I actually don't know how you even do that in python, but apparently this implementation is an example.  I'll want to have a look at that later for my education.

One last tidbit.   I didn't say one thing quite right above.  I think we should consider having the `LoggingEnsemble` template inherit `ProcessingHistory`.   Long ago we made the decision to put an ErrorLogger object as an attribute of ProcessingHistory (ProcessingHistory "has an" ErrorLogger exactly describes our design concept.)   Why that matters is I think object level history will work exactly like it does for Seismogram and TimeSeries if we make that small change.  I don't think it will impact the existing code base at all.  To make it really work, however, would require some nontrivial work:
1.  Ensemble pickle will immediately break.  Well, it might work but the ProcessingHistory data would be lost in a parallel operation.  Easily fixed with some minor changes to the pybind11 code.
2.  The decorators for adding processing history functionality would likely need some changes but I am not sure how difficult that would be.
3. New tests would have to be created.

I think you should merge these changes and put this on our development agenda for the not to distant future.

wangyinz:
yeah, I agree that we should think more carefully about the design and implement that later. For now, I do plan to merge this branch, but I should at least add the docstrings to the new function calls first.

Issue #333, opened 2022-09-15
Bug in file based reader


pavlis:
I have found an umambiguous bug in the Seismogram file reader.  It definitely fails with low level reads when "format" is set None to the function called `_read_data_from_dfile`.   In that situation I can tell from debugging that the following section of code of that function is being executed 
```
else:
                # We can only get here if this is a Seismogram                try:
                    nsamples = 3 * mspass_object.npts
                    count = _fread_from_file(mspass_object, dir, dfile, foff)
                    if count != nsamples:
                        message = "fread count mismatch.  Expected to read {nsamples} doubles but fread returned a count of {count}".format(
                            nsamples=nsamples, count=count
                        )                        mspass_object.elog.log_error(                            "_read_data_from_dfile", message, ErrorSeverity.Complaint
                        )
                except MsPASSError as merr:                    # Errors thrown must always cause a kill
                    mspass_object.kill()
                    mspass_object.elog.log_error(merr)
```

The mystery, however, was that I had an example from reading a raw, binary data file where the error log contains the following entry posted by the C++ code being called in the line with `_fread_from_file`:
```
[LogData({'job_id': 0, 'p_id': 2041, 'algorithm': 'MsPASSError', 'message': 'fread_data_from_file:  Open failed on file simdata/simdata_22', 'badness': ErrorSeverity.Invalid})]
```
Debugging I can also see the code step into the exception handler for a MsPASSError that is the last 3 lines of the above.   What was baffling to me was that output of the Database read_data method for the same example was NOT marked dead but alive. 

The problem, however, is clear.  There is a flaw in the logic of how our reader is handling kills in any part of the read.   The offending section is the following, which is the last section of the read_data method that is executed after the call to _read_data_from_dfile:
```
            # 3.load history
            if load_history:
                history_obj_id_name = (
                    self.database_schema.default_name("history_object") + "_id"
                )
                if history_obj_id_name in object_doc:
                    self._load_history(
                        mspass_object,
                        object_doc[history_obj_id_name],
                        alg_name,
                        alg_id,
                        define_as_raw,
                        retrieve_history_record,
                    )

            mspass_object.live = True
            mspass_object.clear_modified()

            # 4.post complaint elog entries if any
            for msg in log_error_msg:
                mspass_object.elog.log_error("read_data", msg, ErrorSeverity.Complaint)

        return mspass_object
```
There is clearly an error with the line `mspass_object.live = True` as it will always override a kill.  

[...]

PR #334, opened 2022-09-15
Fix bug wrongly setting object alive in read_data

wangyinz:
This PR fix the bug in #333, where a killed object during the file reading is set back to alive later by mistake


wangyinz:
I initially just move the line of code that sets the object alive to earlier section. This will eliminate the original error. However, after thinking it over, I believe it is a bad idea to silently kill the object when an exception occurs in the reader. Such an error should be considered fatal in all cases.

wangyinz:
I implemented the change suggested above. This is also the behavior of the other readers we define, so the behavior is consistent now.

Issue #332, opened 2022-09-15
Problem with ExtractComponent and implications for ensemble handling


pavlis:
I have uncovered a problem with our implementation of the function `ExtractComponent`.   That functions use is to extract single component from three component data.   It is overloaded in C++ and defined the pythonic way in the pybind11 python bindings.  i.e. if you give the function a Seismogram it should return a TimeSeries and if you give it a SeismogramEnsemble it should give you a TimeSeriesEnsemble back.  The version for Seismogram works, although it is lacking in another way I'll get to in a minute.  The ensemble version does not work right as it returns, in C++ syntax, an `Ensemble<CoreTimeSeries>` object.  That, for example, creates the following error if you run the BasicGraphics tutorial when the tutorial tries to show how to plot a SeismogramEnsemble:
``
RuntimeError: ('SeismicPlotter._imageplot (Error):  Received unsupported data type=', <class 'mspasspy.ccore.seismic.CoreTimeSeriesEnsemble'>)
``
The symbol CoreTimeSeriesEnsemble is defined this way in the pybind11 code:
```
py::class_<Ensemble<TimeSeries>,Metadata>(m,"CoreTimeSeriesEnsemble","Gather of scalar time series objects")
    .def(py::init<>())
     ...
```
Noting a `TimeSeriesEnsemble` is defined this way:
```
  py::class_<LoggingEnsemble<TimeSeries>, Ensemble<TimeSeries> >(m,"TimeSeriesEnsemble","Gather of scalar time series objects")
```
where a reminder that that incantation says TimeSeriesEnsemble is a subclass of CoreTimeSeriesEnsemble aka Ensemble<TimeSeries>,   In short, the problem is created by a type mismatch in the return of `ExtractComponent` for ensembles that returns the superclass when mspass workflows normally expect the top of the hiearchy for ensembles - TimeSeriesEnsemble.  

The specific fix to the problem at hand, I believe, is to deal with two related problems;

1. The `ExtractComponent` function needs a python wrapper to: (a) handle type mismatches (including None types) more cleanly in the mspass way, and (b) add the decorators for mspass functions to enable the standard goodies (Right now calls to this function go straight to the C++ code.).   A wrapper provides an immediate but inelegant fix to the known bug illustrated by the graphics tutorial.     
[...]

PR #338, opened 2022-09-28
Add Extract Component Function and Tests

Aristoeu:
For issue #332 Problem with ExtractComponent

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,7 +2,7 @@ sphinx == 4.4.0
 sphinx-rtd-theme
 m2r2
 breathe
-nbconvert==5.6.1
+nbconvert
 ipython_genutils
 nbsphinx
 sphinx-copybutton
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,8 +2,6 @@ sphinx
 sphinx-rtd-theme
 m2r2
 breathe
-nbconvert
 ipython_genutils
-nbsphinx
 sphinx-copybutton
 jinja2
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,8 +2,9 @@ sphinx == 4.4.0
 sphinx-rtd-theme
 m2r2
 breathe
-nbconvert==5.6.1
+nbconvert==7.0.0
 ipython_genutils
 nbsphinx
 sphinx-copybutton
 jinja2==3.0.0
+mistune==0.8.4
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,10 +1,9 @@
-sphinx == 4.4.0
+sphinx
 sphinx-rtd-theme
 m2r2
 breathe
-nbconvert==6.5.4
+nbconvert
 ipython_genutils
 nbsphinx
 sphinx-copybutton
-jinja2==3.0.0
-mistune==0.8.4
+jinja2
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,7 +2,7 @@ sphinx == 4.4.0
 sphinx-rtd-theme
 m2r2
 breathe
-nbconvert==7.0.0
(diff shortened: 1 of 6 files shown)


python/mspasspy/algorithms/basic.py
```diff
@@ -1,5 +1,55 @@
...
+    object_history=False,
+    alg_name=None,
+    alg_id=None,
+    dryrun=False,
+    inplace_return=False,
+    function_return_key=None,
+    **kwargs
+):
+    if isinstance(data, Seismogram):
+        empty = TimeSeries()
```
wangyinz, line 28: It is not a good idea to create an empty data object here. We only need it when the input is dead or exception occurs. Otherwise, this is only wasting CPU cycles and memory.

Also, the returned data object (whether being empty or not) should inherit the elog and object level history of the input data object. Note that the object level history is like a chain. This step converts a Seismogram into TimeSeries (or the corresponding Ensemble type), and should be considered a part of the chain.

python/mspasspy/algorithms/basic.py
```diff
@@ -1,5 +1,55 @@
...
+        empty.kill()
+        if data.dead():
+            return empty
+        try:
+            d = bsc.ExtractComponent(data, component)
+            return d
+        except Exception as err:
+            data.elog.log_error("ExtractComponent", str(err), ErrorSeverity.Invalid)
+            return empty
+    elif isinstance(data, SeismogramEnsemble):
```
wangyinz, line 38: This has the same issue as above. Although we currently don't have a history for the ensemble type, the members of the ensemble may still container history that should be retained.

wangyinz (changes requested):

python/mspasspy/algorithms/basic.py
Aristoeu: I will add load_history() here to copy the ProcessingHistory of input data to the new TimeSeries object

python/mspasspy/algorithms/basic.py
Aristoeu: Here I let each member in the TimeSeriesEnsemble load history of the corresponding member in the input SeismogramEnsemble. When the input is dead or an exception occurs, the return object will have elog inherited.

python/mspasspy/algorithms/basic.py
```diff
@@ -1,5 +1,66 @@
...
+import mspasspy.ccore.algorithms.basic as bsc
+from mspasspy.ccore.seismic import (
+    TimeSeries,
+    TimeSeriesEnsemble,
+    Seismogram,
+    SeismogramEnsemble,
+)
+
+
+@mspass_func_wrapper
```
wangyinz, line 14: Please also add docstrings (as well as necessary comments) to new functions.

python/tests/algorithms/test_basic.py
```diff
@@ -18,6 +24,37 @@
 from helper import get_live_seismogram, get_live_timeseries, get_sin_timeseries
 
 
+def test_ExtractComponent():
```
wangyinz, line 27: The tests here is far from sufficient. First, you need to make sure most (if not all) the new codes are covered, and you should double check the coverage [here](https://app.codecov.io/gh/mspass-team/mspass/pull/338). Then, you need to make sure the tests also cover reasonable amount of corner cases. This is especially important to ensure exceptions are captured as expected.

docs/requirements.txt
```diff
@@ -2,7 +2,7 @@ sphinx == 4.4.0
 sphinx-rtd-theme
 m2r2
 breathe
-nbconvert==6.5.1
+nbconvert==5.6.1
```
wangyinz, line 5: Please also resolve the version issue here.

wangyinz (changes requested):
Please see the detail comments. Thanks!

wangyinz (approved):

PR #340, opened 2022-10-10
Identify if Dask or Spark is installed

Aristoeu:
Added a judgment, if there is Dask, use Dask first, otherwise, detect whether there is a Spark. 
Added mspass_normal_map function using map in Python and mspass_normal_reduce function using reduce in functools


Aristoeu:
Defined global variable _hasDask and _hasSpark in `__init__` file, to show if Dash or Spark is installed. 

In client.py, in case no scheduler is specified:
- If there is a local Dask, use Dask, otherwise if there is a local Spark, use Spark
- If neither is available, self._scheduler = None, start only DBClient

In manager.py, I defined mspass_normal_map and mspass_normal_reduce, using neither Spark nor Dask

python/mspasspy/global_history/manager.py
```diff
@@ -1,21 +1,179 @@
...
+                *args,
+                object_history=object_history,
+                alg_name=alg_name,
+                alg_id=str(alg_id),
+                **kwargs
+            ),
+            data,
+        )
+
+    return functools.reduce(
```
wangyinz, line 172: Is this gonna work? The data here needs to be a iterable, right? This is not mentioned in the docstrings. `args` is not needed for `functools.reduce`, right?

python/mspasspy/global_history/manager.py
```diff
@@ -1,21 +1,179 @@
...
+    alg_name=None,
+    parameters=None,
+    **kwargs
+):
+    """
+     This decorator method performs map function in Python. Instead of performing the normal map function,
+     if user provides global history manager, alg_id(optional), alg_name(optional) and parameters(optional)
+     as input, the global history manager will log down the usage of the algorithm. Also, if user set
+     object_history to be True, then each mspass object in this map function will save the object level history.
+
```
wangyinz, line 39: definition of data is missing. Also, in the map function, it doesn't need `args` and `kwargs`.

python/mspasspy/client.py
```diff
@@ -4,10 +4,23 @@
...
 from mspasspy.db.database import Database
 from mspasspy.global_history.manager import GlobalHistoryManager
-from pyspark import SparkConf, SparkContext
-from pyspark.sql import SparkSession
+
+try:
+    from pyspark import SparkConf, SparkContext
+    from pyspark.sql import SparkSession
+
+    hasSpark = True
```
wangyinz, line 12: Don't use global variables like this. Rename them to something like `__mspasspy_has_pyspark` and `__mspasspy_has_dask_distributed`. Also, for dask, we need to consider the difference between `dask` and `dask.distributed`. Without `dask.distributed`, `dask` could still be used on a single node for parallelism.

python/mspasspy/db/database.py
```diff
@@ -10,8 +10,12 @@
...
 from array import array
 import pandas as pd
-import dask.bag as daskbag
-import dask.dataframe as daskdf
+
+try:
+    import dask.bag as daskbag
+    import dask.dataframe as daskdf
+except:
+    pass
```
wangyinz, line 18: This should decide which should be use as the default `format` argument in the `read_distributed_data`. If neither is available, we should just hide `read_distributed_data`.

python/mspasspy/util/converter.py
```diff
@@ -6,7 +6,11 @@
 import obspy.core
 import collections
 import pandas as pd
-import dask.dataframe as daskdf
+
+try:
+    import dask.dataframe as daskdf
+except:
+    pass
```
wangyinz, line 13: This should change how `Textfile2Dataframe` behaves on the `parallel` argument.

wangyinz (changes requested):
This is good. However, I figured a lot of the new code are not covered by tests. Please go ahead and add the missing tests. Note that you could use mock to fail an import in the test. Please refer to solution 4 of [this article](https://medium.com/python-pandemonium/how-to-test-your-imports-1461c1113be1).

Aristoeu:
All the tests run via Pytest are run in a single process. So it is difficult to simulate both cases with and without spark and dask.

python/mspasspy/global_history/manager.py
```diff
@@ -1,21 +1,173 @@
...
 from mspasspy.util.converter import AntelopePf2dict
 from datetime import datetime
 from dill.source import getsource
+import functools
 
 import mspasspy.algorithms.signals as signals
 from mspasspy.global_history.ParameterGTree import ParameterGTree, parameter_to_GTree
 
 
+def mspass_normal_map(
```
wangyinz, line 23: Please name them as `mspass_map` and `mspass_reduce` for consistency. Note that those are the same name we insert into the Bag and RDD.

python/mspasspy/reduce.py
```diff
@@ -1,6 +1,8 @@
-import pyspark
-import dask.bag as daskbag
-
+try:
+    import dask.bag as daskbag
+    import pyspark
```
wangyinz, line 3: Same as above.

python/mspasspy/reduce.py
```diff
@@ -76,5 +78,8 @@ def mspass_dask_foldby(self, key="site_id"):
     ).map(lambda x: list2Ensemble(x[1]))
 
 
-pyspark.RDD.mspass_foldby = mspass_spark_foldby
-daskbag.Bag.mspass_foldby = mspass_dask_foldby
+try:
+    daskbag.Bag.mspass_foldby = mspass_dask_foldby
+    pyspark.RDD.mspass_foldby = mspass_spark_foldby
```
wangyinz, line 83: same issue

python/mspasspy/util/converter.py
```diff
@@ -674,12 +680,12 @@ def Textfile2Dataframe(
     if (
         header_line is None or header_line < 0
     ):  #   Header_line not given, using attribute_names
-        if parallel:
+        if parallel and __mspasspy_has_dask:
             df = daskdf.read_csv(filename, sep=separator, names=attribute_names)
```
wangyinz, line 684: Let's add a warning here. If parallel is specified but dask doesn't exist, the user need to be aware of the fact that it is running pandas dataframe. 

Same with the rest below.

python/mspasspy/db/database.py
```diff
@@ -54,7 +68,7 @@ def read_distributed_data(
     normalize=None,
     load_history=False,
     exclude_keys=None,
-    format="dask",
+    format=None,
```
wangyinz, line 71: Change the default to none doesn't make sense here. The default should be dask or pyspark if only one of the two exist, and it should be dask if both exist. Then, if neither of them exists, this function shouldn't be imported at all.

.github/workflows/python-package.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       run: |
         export MSPASS_HOME=$(pwd)
         export SPARK_LOCAL_IP=127.0.0.1
-        pytest --cov=mspasspy python/tests/
+        for file in `find python/tests -type f -name "test_*.py"`; do pytest --cov=mspasspy $file; done
```
wangyinz, line 57: @Aristoeu Why do you change this line?

Aristoeu: The tests run via pytest are run in a single process and the import will be a mess, as shown in this link:
https://stackoverflow.com/questions/55229163/pytest-independent-import-of-the-same-module-from-different-test-files

But if using `pytest --forked`, tests will be able to run in different processes and the problem can be partially solved, as shown in the link below. 
https://stackoverflow.com/questions/48234032/run-py-test-test-in-different-process
I tested it locally with `pip install pytest-forked pytest-xdist` and execute `pytest --forked python/tests/`, then all the tests can pass except for `test_manager.py` (global_history/test_manager_spark_dask.py). It will have four functions failed if adding '--forked'. I see the log and think the reason is: the four tests in this file depend on the results of other tests, and if using '--forked' the tests are all run independently, so they don't have the expected results and will fail.

I can come up with a solution as: not using `--forked` parameter for `test_manager.py` and using this parameter for other files. Or I can modify the tests in `test_manager.py` to make them independent and add `--forked` parameter to `pytest` here.

wangyinz: Cool! Thanks for collecting the info! I think using --forked makes sense. Let's do that.

Aristoeu:
pytest hangs here after I added "--forked", but it doesn't stop when I run locally, similar to the flowing links.
https://github.com/pytest-dev/pytest-xdist/discussions/805
https://github.com/sktime/sktime/issues/1941

Aristoeu:
Removed pytest-forked, as it will lower the code coverage. Moved new tests to the folder tests/manual

PR #347, opened 2022-11-26
Improve snr algorithm

pavlis:
This is an important addition that does two important things:
1.  It adds tests for the snr module.
2.  Running the tests revealed a number of bugs in the snr module that have been fixed. 

There was also a minor feature addition that wasn't really a bug.  I added an argument for the low level FD_snr_estimator function that fixes the high frequency corner used to define the signal bandwidth.   That option should probably be the norm for teleseismic P wave processing for receiver functions.   

I am working on a user manual page and tutorial on this topic.   The bug fixes here are important enough, however, that we shouldn't delay merging this branch any longer than necessary to verify it tests out.   

There are two issues with the tests worth pointing out.  
1.  Computing signal to noise ratio requires a mechanism to generate random numbers to define controlled noise.   I chose to do this with numpy's random function.   That creates a potential false failure if we are unlucky and happen to get an outlier in a test run.   I think i made all my bounds for the tests large enough that that is not likely, but it will always remain possible.  I could be avoided by adding a fixed noise sequence but that would require a big revision to the tests. I suggest we use this and it turns out to bite us in the future it can be changed. Perhaps I should have put a stronger warning in the comments of the test code.
2. I know for sure this revision breaks  another test that computes amplitudes.  One of the most serious bugs I found was a mistake in the rms formula that is breaking other tests when corrected.


pavlis:
This seems to now be testing clean.  I would strongly advise you to merge this branch right away.   There are residual issues related to the need it demonstrates for an arrival collection.   I just started a discussion on that topic that has already gotten a bit lengthy and could get worse.   I would not like to see the fixes in this branch be delayed in being implemented.  There were some very serious bugs  this branch resolves.  

There is also a partially written user manual section on the snr module, but that also can't be completed until the arrival collection design issue is resolved.

wangyinz:
Sounds good and merged!

Issue #345, opened 2022-11-19
problem with save_data file writes


pavlis:
Multiple things appear to be broken in the save_data method of Database when writing to files with "format == None".   That triggers a block of C code that writes the raw data with a binary fwrite to  a file.   The offender is this private functions:
```
@staticmethod
    def _save_data_to_dfile(
        mspass_object, dir, dfile, format=None, kill_on_failure=False
    ):
        """
        This is a private method used under the hood to save the sample data
        of atomic MsPASS data objects.  How the save happens is highly
        dependent upon the format argument.  When None, which is the
        default, the data are written to the file specified by dir and dfile
        using low-level C fwrite calls.  The function used always appends
        data to eof if the file already exists to allow accumation of data
        in "gather" files to reduce file name overhead in hpc systems.
        If format is anything else the function attempts to use one of
        the obspy formatted writers.  That means the format string must be
        one of the `supported formats <https://docs.obspy.org/packages/autogen/obspy.core.stream.Stream.write.html#supported-formats>`__ of ObsPy reader.
        Writing to a file can fail for any number of reasons.   write errors
        are trapped internally in this function any errors posted to elog of
        mspass_object.   If the kill_on_failure boolean is set true the
        function will call the kill method of the data function and the
        that datum will be marked dead.   That is not normally a good idea
        if there is any additional work to do on the data so the default
        for that parameter is false.
        Because we use a stream model for file storage be aware
        that insertion and deletion in a file are not possible.  If you
        need lots of editing functionality with files you should use the
        one file to one object model or (better yet) use gridfs storage.
        :param mspass_object: the target object.
        :type mspass_object: either :class:`mspasspy.ccore.seismic.TimeSeries` or :class:`mspasspy.ccore.seismic.Seismogram`
        :param dir: file directory.
        :type dir: :class:`str`
        :param dfile: file name.
        :type dfile: :class:`str`
        :param format: the format of the file. This can be one of the `supported formats <https://docs.obspy.org/packages/autogen/obspy.core.stream.Stream.write.html#supported-formats>`__ of ObsPy reader. By default (``None``), the format will be the binary waveform.
        :type format: :class:`str`
        :param kill_on_failure:  When true if an io error occurs the data object's
          kill method will be invoked.  When false (the default) io errors are
          logged and left set live.  (Note data already marked dead are return
          are ignored by this function. )
        :type kill_on_failure: boolean
        :return: Position of first data sample (foff) and the size of the saved chunk.
[...]

PR #356, opened 2022-12-05
Fix save_data problem and add c++ writer reader

Aristoeu:
Applied the patch in #345, and added a writer and reader for ensembles in C++
For the writer, it will write the whole ensemble in one file.
For the reader, it has a list of file paths and will read from all files to form an ensemble.

I am not sure if the parameters of these functions are correct. I haven't added Python bindings to these functions.


cxx/src/lib/io/fileio.cc
```diff
@@ -83,6 +84,86 @@ long int fwrite_to_file(Seismogram& d, const string dir,const string dfile)
...
+		  /* Null as always in unix means use current directory*/
+			fname=dfile;
+		if((fp=fopen(fname.c_str(),"a")) == NULL)
+		  /* use the name of the overloaded parent instead of the actual function - intentional*/
+			throw MsPASSError("fwrite_to_file:  Open failed on file "+fname,ErrorSeverity::Invalid);
+	  /* Both fseek and ftell can fail in weird circumstances, but I intentionally
+		do not trap that condition as if either have issues I am quite sure
+		the fwrite will fail */
+		fseek(fp,0L,2);
+		foff = ftell(fp);
```
pavlis, line 147: This call to fseek is misplaced.   It belongs inside the loop over members.  This code as is will not work.  Each member's metadata will have the same foff == 0.

cxx/src/lib/io/fileio.cc
```diff
@@ -129,4 +210,34 @@ size_t fread_from_file(TimeSeries& d,const string dir, const string dfile,
 		return ns_read;
 	}catch(...){throw;};
 }
+size_t fread_from_file(Ensemble<Seismogram> &de, vector<string> dirs, vector<string> dfiles,
```
pavlis, line 213: This algorithm has some issues.   A reader has some complexity not shared by a writer.   In fact, there are a some design decisions the group needs to discuss:
1.   A problem in ensemble reader is matching Metadata stored in MongoDB with the sample data stored as a block by the writer.   With the current model having TimeSeries and Seismogram atomic each document in a wf colletion defines one TimeSeries or Seismogram with dir/dfile and foff defing where fread should be positioned to load the sample data.   If we stick with that model the algorithm needs to sort documents properly.   The sort requires 2 keys:  dir and dfile.   The reader then could remain stateless by grouping the data by unique dir:dfile keys.  We open and close the file then only when the dir or dfile change.   
2. The alternative is to allow a new structure for ensembles in a wf collection using subdocuments.   That is, the ensemble's entry in a wf collection would have the ensemble metadata at the top level of the document and the member metadata in a subdocument with some unique key to define the subdocument lile "member_metadata

I think the sort - group model is the better solution.  It is more generic and doesn't potentially break other things.

Aristoeu:
In `database.py`, I added two methods `save_ensemble_data_binary_file` and `read_ensemble_data_group` to use the new reader and writer. Firstly reading or writing the metadata for every object in the ensemble, and calling the C++ reader and writer for reading or writing local binary files.

ready to review @wangyinz

pavlis:
I couldn't figure out how to add comments to the work here.   I had commented on an older version and don't see how to get into a mode to comment directly on the revised code.   Guess I deserved a troglodyte label. 

The new C++ ensemble writers look fine but the reader has aome fundamental flaws.   
1.   The use of the argument `const long int length` is not correct.  It is used later for a call to the `resize` method of `std::vector`.   Not positive but I'm pretty certain resize will invalidate any existing content of the container or at least leave it undefined.  
2. I think the approach of passing in the file offset values with an atd::vector<long int> is not the right model.   The writer writes foff, dir, dfile, and npts with each document it saves - assumed since the C++ code posts those attributes.  I need to look at how you did the python function that calls the C++ functions but I assume you are using the constructors with a `BasicTimeSeries` and `Metadata` arguments.    Both load the Metadata, create the `BasicTimeSeries` skeleton, and initialize the sample data container to all zeros.   This function should then extract foff from Metadata and then call fseek.  It can then, as now, use the `npts` method to tell fread how many bytes should be read.   
3.  We need to consider what exceptions should be handled and how they should be handled?   First, we need to follow the mspass model where if there is a failure use kill since the member vector is assumed already filled with a skeleton.   I think the current things are right - calls to fseek and fread that fail should cause an exception as they are a serious problem that shouldn't happen if the data set is properly loaded.   The one thing to consider is a check of npts stored in the metadata with the size of the data vector defined with the npts method.   The cost is tiny and would make the function more self contained.   
4. The arguments should be `LoggingEnsemble` not `Ensemble`.  
5. I do not see a function prototype for these new functions in ensemble.h.  Maybe I missed them as I only looked on github and didn't know how to run the equivalent of grep.   They are needed and should include doxygen comments to document the usage of the new functions.  The documentation should emphasize the functions are mainly intended as a companion to python functions in mspass and should be used with caution for any other application.

cxx/src/lib/io/fileio.cc
```diff
@@ -83,6 +84,88 @@ long int fwrite_to_file(Seismogram& d, const string dir,const string dfile)
...
+		else
+		  /* Null as always in unix means use current directory*/
+			fname=dfile;
+		if((fp=fopen(fname.c_str(),"a")) == NULL)
+		  /* use the name of the overloaded parent instead of the actual function - intentional*/
+			throw MsPASSError("fwrite_to_file:  Open failed on file "+fname,ErrorSeverity::Invalid);
+	  /* Both fseek and ftell can fail in weird circumstances, but I intentionally
+		do not trap that condition as if either have issues I am quite sure
+		the fwrite will fail */
+		fseek(fp,0L,2);
```
pavlis, line 106: This call to fseek is not necessary.   The file is opened "a" which means it will be position at end of file if it already exists.   The comment should be removed as well.

cxx/src/lib/io/fileio.cc
```diff
@@ -83,6 +84,88 @@ long int fwrite_to_file(Seismogram& d, const string dir,const string dfile)
 
 	return(foff);
 }
+std::vector<long int> fwrite_to_file(mspass::seismic::Ensemble<mspass::seismic::TimeSeries>& d, const std::string dir,const std::string dfile)
```
pavlis, line 87: This should be a `LoggingEnsemble`.   Not essential here for this writer as inheritance might allow the subclass to work, but I would have to test that to be sure.

cxx/src/lib/io/fileio.cc
```diff
@@ -83,6 +84,88 @@ long int fwrite_to_file(Seismogram& d, const string dir,const string dfile)
...
+			t.put_string(SEISMICMD_dir, dir);
+			t.put_string(SEISMICMD_dfile, dfile);
+			t.put_long(SEISMICMD_foff, foff);
+		}	
+		fclose(fp);
+
+		return foffs;
+	}catch(...){throw;};
+}
+std::vector<long int> fwrite_to_file(mspass::seismic::Ensemble<mspass::seismic::Seismogram>& d, const std::string dir,const std::string dfile)
```
pavlis, line 128: As for TimeSeries version - LoggingEnsemble

cxx/src/lib/io/fileio.cc
```diff
@@ -83,6 +84,88 @@ long int fwrite_to_file(Seismogram& d, const string dir,const string dfile)
...
+		else
+		  /* Null as always in unix means use current directory*/
+			fname=dfile;
+		if((fp=fopen(fname.c_str(),"a")) == NULL)
+		  /* use the name of the overloaded parent instead of the actual function - intentional*/
+			throw MsPASSError("fwrite_to_file:  Open failed on file "+fname,ErrorSeverity::Invalid);
+	  /* Both fseek and ftell can fail in weird circumstances, but I intentionally
+		do not trap that condition as if either have issues I am quite sure
+		the fwrite will fail */
+		fseek(fp,0L,2);
```
pavlis, line 147: Same issue with fseek in this similar code.

cxx/src/lib/io/fileio.cc
```diff
@@ -129,4 +212,80 @@ size_t fread_from_file(TimeSeries& d,const string dir, const string dfile,
 		return ns_read;
 	}catch(...){throw;};
 }
+size_t fread_from_files(mspass::seismic::Ensemble<mspass::seismic::Seismogram> &de, const std::string dir,
```
pavlis, line 215: 1.  Name should be singular `fread_from_file` since only one file is referenced.
2. Change the signature to the following:
```
size_t fread_from_file(mspass::seismic::LoggingEnsemble<mspass::seismic::Seismogram>&de,
 const std::string dir, const std::string dfile)
```
See related comment in pull request page for why.  I'll try to show patches needed for this below but may miss some.

cxx/src/lib/io/fileio.cc
```diff
@@ -129,4 +212,80 @@ size_t fread_from_file(TimeSeries& d,const string dir, const string dfile,
...
+		fname = dir + "/" + dfile;
+	else
+	  /* Null as always in unix means use current directory*/
+		fname=dfile;
+	if((fp=fopen(fname.c_str(),"r")) == NULL)
+		throw MsPASSError("fread_data_from_file:  Open failed on file "+fname,ErrorSeverity::Invalid);
+	
+	for (int i = 0; i < n; ++i) {
+		size_t ns_read;
+		long int foff = foffs[i];
```
pavlis, line 235: Change to:
```
if(ds.member[i].is_defined(SEISMICMD_foff))
{
   foff = ds.member[i].get_long(SEISMICMD_foff);
}
else
{
   ds.member[i].kill();
   ds.member[i].elog.log_error( some error message);
   continue;   
}
```

cxx/src/lib/io/fileio.cc
```diff
@@ -129,4 +212,80 @@ size_t fread_from_file(TimeSeries& d,const string dir, const string dfile,
...
+		int index = indexes[i];
```
pavlis, line 236: remove

cxx/src/lib/io/fileio.cc
```diff
@@ -129,4 +212,80 @@ size_t fread_from_file(TimeSeries& d,const string dir, const string dfile,
...
+		try{
+			if(foff>0)
+			{
+				if(fseek(fp,foff,SEEK_SET))
+				{
+					fclose(fp);
+				throw MsPASSError("fread_data_from_file:  fseek failure on file="+fname,ErrorSeverity::Invalid);
+				}
+			}
+			ns_read = fread((void*)de.member[index].u.get_address(0, 0), sizeof(double), 3 * de.member[index].npts(), fp);
```
pavlis, line 246: `index` should just be `i`.   Also you must trap a error condition for fread failures.   The usual test would, I think, be some version of this:
```
if(ns_read != 3*de.member[i].npts()} 
{
   de.member[i].elog.log_error( some error message);
   de.member[i].kill();
}
else
{
  de.member[i].set_live();
}
```
Noting that we also must explicitly set live all successful reads.

cxx/src/lib/io/fileio.cc
```diff
@@ -129,4 +212,80 @@ size_t fread_from_file(TimeSeries& d,const string dir, const string dfile,
...
+	for (int i = 0; i < n; ++i) {
+		size_t ns_read;
+		long int foff = foffs[i];
+		int index = indexes[i];
+		try{
+			if(foff>0)
+			{
+				if(fseek(fp,foff,SEEK_SET))
+				{
+					fclose(fp);
```
pavlis, line 242: You shouldn't raise an exception here.  Use the kill and elog model as illustrated above.

cxx/src/lib/io/fileio.cc
```diff
@@ -129,4 +212,80 @@ size_t fread_from_file(TimeSeries& d,const string dir, const string dfile,
...
+	string fname;
+	if(dir.length()>0)
+	  /* for expected context for use in python we will assume dir does not
+		have a trailing path separator so we always insert / */
+		fname = dir + "/" + dfile;
+	else
+	  /* Null as always in unix means use current directory*/
+		fname=dfile;
+	if((fp=fopen(fname.c_str(),"r")) == NULL)
+		throw MsPASSError("fread_data_from_file:  Open failed on file "+fname,ErrorSeverity::Invalid);
```
pavlis, line 231: This should perhaps not throw an exception but cause the entire ensemble to be killed (`de.kill()`), an error message posted, and then return -1.   In a large dataset we want to handle the (common) issue that some but not all files are missing.

cxx/src/lib/io/fileio.cc
```diff
@@ -129,4 +212,80 @@ size_t fread_from_file(TimeSeries& d,const string dir, const string dfile,
...
+				}
+			}
+			ns_read = fread((void*)de.member[index].u.get_address(0, 0), sizeof(double), 3 * de.member[index].npts(), fp);
+			ns_read_sum += ns_read;
+		}catch(...){throw;};
+	}
+	fclose(fp);
+	return ns_read_sum;
+}
+size_t fread_from_files(mspass::seismic::Ensemble<mspass::seismic::TimeSeries> &de, const std::string dir,
```
pavlis, line 253: Apply the same changes you make to the Seismogram version of this algorithm to this function.

python/mspasspy/db/database.py
```diff
@@ -2761,6 +2769,336 @@ def read_ensemble_data(
...
+
+                # 4.post complaint elog entries if any
+                for msg in log_error_msg:
+                    mspass_object.elog.log_error(
+                        "read_data", msg, ErrorSeverity.Complaint
+                    )
+
+        res = cur_collection.aggregate(
+            [
+                {"$match": {"$expr": {"$in": ["$_id", objectid_list]}}},
```
pavlis, line 3060: For efficiency you might want to sort this group by foff at the end of the chain.  The reason is that when reading from a file sequential reads are almost always much faster than than random reads because the input is always buffered.   When you read through a file sequentially the operating system can keep the read buffer full and not waste resources moving in the same page many times.

python/mspasspy/db/database.py
```diff
@@ -2761,6 +2769,336 @@ def read_ensemble_data(
...
+                    }
+                },
+            ]
+        )
+
+        files = list(res)
+        for f in files:
+            cur_dir = f["_id"]["dir"]
+            cur_dfile = f["_id"]["dfile"]
+            foofs = list(map(int, f["foffs"]))  # string list -> int list
```
pavlis, line 3075: This section needs changes to match what I suggested for the C++ function being used here.  

Also you need to add some comment here about how the python code interacts with the C code and what assumptions both sides make for efficiency.   For your education that is always good practice but especially for a mixed C - python example like this.   We all get lazy and violate that rule way too often.

python/mspasspy/db/database.py
```diff
@@ -2761,6 +2769,336 @@ def read_ensemble_data(
...
+            # get original index
+            indexes = list(map(objectid_list.index, f["ids"]))
+
+            try:
+                cnt = _fread_from_files(
```
pavlis, line 3080: This couldn't have worked before.   You need to first build the skeleton of the ensemble before call the new C++ function `_fread_from_file`.   Suggest this be done as a helper, private method.   The reason is you can probably write a single function to handle building the skeleton for both the Seismogram and TimeSeries versions.  Use the constructors for both objects that take a `Metadata` and `BasicTimeSeries` as inputs.  You create the Metadata container directly from mongodb document retrieved in a cursor loop, construct the BasicTimeSeries object, and then call that constructor.  That can be done in the cursor loop pushing each skeleton to the member container.     If that isn't clear (highly likely) reply and I'll try to expand.

cxx/src/lib/io/fileio.cc
```diff
@@ -129,4 +212,80 @@ size_t fread_from_file(TimeSeries& d,const string dir, const string dfile,
...
+					fclose(fp);
+				throw MsPASSError("fread_data_from_file:  fseek failure on file="+fname,ErrorSeverity::Invalid);
+				}
+			}
+			ns_read = fread((void*)de.member[index].u.get_address(0, 0), sizeof(double), 3 * de.member[index].npts(), fp);
+			ns_read_sum += ns_read;
+		}catch(...){throw;};
+	}
+	fclose(fp);
+	return ns_read_sum;
```
pavlis, line 251: Before returning set the ensemble live:  `de.set_live();`

Aristoeu:
Thanks for the comments and explanation!

cxx/src/lib/io/fileio.cc
Aristoeu: I think there may be a case that the elements in the ensemble are stored in different files. And `fread_from_file` function reads only a part of the ensemble (several TimeSeries or Seismogram objects) from this single file. I think we need a parameter `indexes` here to specify which part to read from this file, and also specify the order, as we need the sequential reads.

python/mspasspy/db/database.py
Aristoeu: I don't quite understand here. I think I have built the ensemble skeleton in the above code (from line 2855), which is copied from the `read_data` function. Firstly build the `Metadata` and then use metadata to build the `TimeSeries` or `Seismogram` object, next add the object to the ensemble. And when this line `_fread_from_files` is executed, there are already objects in the ensemble. I will add more comments in this function to explain the confusing part.

PR #357, opened 2022-12-07
Extend snr

pavlis:
This should largely finish the snr module for now.   Includes a bunch of bug fixes found when applying this to thousands of real data from our large usarray test data set that is not feasible to put in github due to the size.  In the process I also reorganized the top level function to split what previously was a single function into two.   That is necessary to allow the user to run the same function either inline with a map call followed by filter to remove junk or in a database mode where the computed metrics are posted to an arrival collection.

This also now has a complete test suite for the functions in snr.py.   I'm sure coverage will show some missing but I think the tests cover the main features. 

Before this branch is merged we need to decide if we need to add a formal "arrival" collection to the default schema.   This implementation has "arrival" hard wired as the default save target for the database save function.

### DIFF
--- a/cxx/src/lib/algorithms/snr.cc
+++ b/cxx/src/lib/algorithms/snr.cc
@@ -7,10 +7,13 @@ namespace mspass::algorithms::amplitudes{
 using mspass::seismic::PowerSpectrum;
 using mspass::utility::Metadata;
 using mspass::utility::VectorStatistics;
+/* These are used to flag 0/0 and x/0 respectively */
+const double INDETERMINATE(-1.0),NOISE_FREE(9999999.9);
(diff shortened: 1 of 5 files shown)


pavlis:
I don't understand why the 3.8 test failed.  The same code passed pytest on my local machine.  It is failing on an assert for isclose with a mysterious difference?   I thought I removed the randomness from the test but maybe it is something else.   

Let me make sure I don't have a version skew.  I'll rebuild my local copy and rerun the local pytest to look for any obvious issues.

wangyinz:
This looks good. I think the only issue is whether we should move the db related function to under the db module. Not sure, but I will merge it for now.

PR #359, opened 2022-12-12
Enhance ensemble

pavlis:
This branch has some major changes I've worked out over the past several months.  (The initial commit times are about a month after I started working on this branch.)   I stress that time because there is a good chance it will break something that has evolved since I forked this branch in early September.  The main changes are:
1.  We had a stub for a gap-handling version of TimeSeries  called `TimeSeriesWGaps` in the C++ API since the earliest versions of MsPASS.  By that I mean we had include files but no implementation of the API we defined.   I implemented that feature.
2. There is a new C++ test program for `TimeSeriesWGaps` that is as complete as I could make it.  
3. The main purpose of this exercise was creation of a single function I called `TimeIntervalReader`.   The concept is something heavily used in Antelope that Danny Harvey always called his "station-channel-time" read library.  It is intended as a low-level reader for large, continuous data sets commonly stored as day files in miniseed directories.   Once an index was built this reader could be used to read data in a common time range specified by a start and end time.  

Known deficiencies:
1.  The `TimeSeriesWGaps` object does not have a pickle interface defined.   I did that because I am not sure we should encourage use of the C++ objects directly in python.   
2. I did nothing to enable the test of the C++ test program for `TimeSeriesWGaps`.   If it isn't automatic it won't be run in the testing here on github.  
3. What is here will not yet work on wf_miniseed data because I haven't yet fixed the problem where posting endtime was disabled in the miniseed indexer.   Once I verify this passes github tests that will need to be fixed before this branch is merged.  
4. I forgot to run black on the new python code so expect a ton of format errors.  

From the deficiencies it should be clear this is mostly pushed now for testing and for discussion within the group.


wangyinz:
Closing it due to overflow of automated review messages. Please open a new PR to be merged.

PR #361, opened 2022-12-14
Test for code formatting as PR

wangyinz:
This is to remove the use of reviewdog in current workflow that could easily overflows the comments in a PR.


wangyinz:
I also changed the docker actions so that the arm64 arch is only built in the weekly schedule.

PR #373, opened 2023-01-09
New ensemble

Yangzhengtang:


python/mspasspy/preprocessing/seed/gather.py
```diff
@@ -0,0 +1,1103 @@
...
+            group = zarr.group(store=store)
+            zarr_arr = group.create_dataset(
+                [doc["zarr_arr_name"]],
+                shape=ensemble_object.member.shape,
+                chunks=ensemble_object.member.data.chunks,
+                dtype=ensemble_object.member.dtype,
+                overwritebool=True,
+            )
+            dask.array.to_zarr(ensemble_object.member.data, zarr_arr, overwrite=True)
+    ensemble_col.update_one(filter=key, update=data, upsert=True)
```
pavlis, line 1103: This is looking very good.  A few basic additions:
1.  See comment about about wrapper needed to make this more robust.  Can't assume input is regular enough in several ways.  The best solution may be to have a set of functions to regularize start times (requires widowing) and sample intervals (requires recently added resample or decimation operators).   
2. Missing or incomplete docstrings for most of the functions/methods in this module.

pavlis (changes requested):
See summary comment at the end gather.py.

pavlis:
This is looking pretty good.   I don't a line-by-line review of the code is of any help from me in particular at this point.  Other than a pat on the back I have a couple small suggestions:
1.   I think it is unnecessary to have the "Old" qualifier in both function names and docstrings in this gather.py file.   We intend to keep both Ensemble and Gather data objects so calling one "Old" will be confusing to a reader and imply it might be prone to deprication.  That is not in the plan for sure as they have different uses.   
2. I am 100% sure the new gather.py file belongs somewhere else than preprocess/seed.   It really has nothing to do with seed data.  I leave it to our master organizer, @wangyinz, to decide where the file should go but definitely belongs somwhere else.   I think we probably should be mspasspy.seismic.gather which seems a logical variant of mspasspy.ccore.seismic.

PR #379, opened 2023-01-17
added dataframe version in read_distributed_data

Aristoeu:
update `read_distributed_data` and add `write_distributed_data`


Aristoeu:
update doc:
**Reader**
* `read_distributed_data`:  input is cursor and database or dataframe, output is rdd/bag of objects (TimeSeries/Seismogram)
  * This function is to divide the process of reading into two parts: reading from database and reading from file, where reading from database is done in sequence, and reading from file is done with dask/spark. The two parts are done in two functions: `read_to_dataframe`, and `read_files`
  * The input can be database or dataframe. If it is database, the function will firstly read from the database sequentially, and save the metadata to a dataframe. Then we have the dataframe. Use the information in the dataframe to read from files using dask/spark distributedly, and generate the objects to the container. This step uses map in dask/spark to improve efficiency.

* `read_to_dataframe`: input is cursor and database, output is dataframe of metadata
  * Firstly construct a list of objects using cursor. Then for each object, constrcut the metadata and add it to the list. Finally, convert the list to a dataframe.

* `read_files`: input is metadata, output is object (TimeSeries/Seismogram)
  * This is the reader for constructing the object from storage. Return type is a complete mspass object.


**Writer**
* `write_distributed_data`: input is rdd/bag of objects (TimeSeries/Seismogram), output is a dataframe of metadata
  * The function can be thought of as writing the entire data set from a parallel container to storage. From the container, it will firstly write to files distributedly using spark/dask, and then write to the database sequentially. The two parts are done in two functions: `write_files`, and `write_to_db`. It returns a dataframe of metadata for each object in the original container. The return value can be used as input for `read_distributed_data` function.
* `write_files`: This is the writer for writing the object to storage. Input is an object (TimeSeries/Seismogram), output is the original object with some more parameters added.
* `write_to_db`: input is an object (TimeSeries/Seismogram), output is metadata.
  * This function will write to the doc and to the database, then return the metadata of the object.

wangyinz:
I think the history is messed up in this branch. The three commits above [1a45770](https://github.com/mspass-team/mspass/pull/379/commits/1a457701d891c809863a6dcecfb45535903a6b78) should not be here. Please clean this up.

Aristoeu:
Yes, there are some commits with resample.py, I have reverted them.

Aristoeu:
Should the new reader replace the old one, or put it in a new module `io` in the doc? 
The writer is also not a method of Database, should it also be added to the `CRUD Operations` page in the doc?

wangyinz:
Let's not replace them for now and put them in the new module. But for the documentation, we could add the new implementation to the existing contents, which is the `CRUD Operations` section. It may be necessary to also have a separate section on the topic of IO that does not involve the database.

python/mspasspy/io/distributed.py
```diff
@@ -0,0 +1,1322 @@
...
+    return list_.map(
+        lambda cur: read_files(
+            Metadata(cur),  # convert dict to metadata
+            gridfs.GridFS(db)
+            if (isinstance(db, Database) and cur["storage_mode"] == "gridfs")
+            else None,
+            # if storage mode is gridfs, pass a GridFS object, it can not be put in the dataframe because rdd/bag
+            # can't pickle _thread.RLock objects
+            aws_access_key_id=aws_access_key_id,
+            aws_secret_access_key=aws_secret_access_key,
```
wangyinz, line 176: These two arguments are missing the docstrings. I noticed it was missing from the old version, but we should add them.

python/mspasspy/io/distributed.py
```diff
@@ -0,0 +1,1322 @@
...
+                        atomic_type = AtomicType.TIMESERIES
+                    else:
+                        atomic_type = AtomicType.SEISMOGRAM
+                    if not collection:
+                        collection = db.database_schema.default_name("history_object")
+                    # load history if set True
+                    res = db[collection].find_one({"_id": history_object_id})
+                    # retrieve_history_record
+                    if retrieve_history_record:
+                        h = pickle.loads(res["processing_history"])
```
wangyinz, line 467: Never use a single letter variable name in any code. This will make the code hard to read and maintain. Long and explicit words are always preferred. So, instead of calling `h`, this should be something like `processing_history_record`

python/mspasspy/io/distributed.py
```diff
@@ -0,0 +1,1322 @@
...
+    gfsh=None,
+    aws_access_key_id=None,
+    aws_secret_access_key=None,
+):
+    """
+    This is the reader for constructing the object from storage. Return type is a
+    complete mspass object. The logic of is same as Database.read_data().
+
+    :param md: the metadata for the object to be read.
+    :type md: :class:`mspasspy.ccore.utility.Metadata`.
```
wangyinz, line 523: The docstring needs some more details and there are arguments missing.

python/mspasspy/io/distributed.py
```diff
@@ -0,0 +1,1322 @@
...
+        name enforcement.  This parameter allows an alernate collection to
+        be used with or without some different name and type restrictions.
+        The most common use of anything other than the default is an
+        export to a diffrent format.
+    :param data_tag: a user specified "data_tag" key.  See above and
+        User's manual for guidance on how the use of this option.
+    :type data_tag: :class:`str`
+    """
+    # 1. write to file system, distributed
+    t = data.map(
```
wangyinz, line 738: Again, don't name it `t`

python/mspasspy/io/distributed.py
```diff
@@ -0,0 +1,1322 @@
...
+        )
+    # convert list of metadata to dataframe
+    return pd.json_normalize(map(lambda cur: cur.todict(), r))
+
+
+def write_to_db(
+    db,
+    mspass_object,
+    mode="promiscuous",
+    storage_mode="gridfs",
```
wangyinz, line 790: We want to make files the default here. Also, `write_to_db` is called after `write_files`, which means the input already contains the info on how the data is saved. I guess this is because the in `write_files` the return type is still a mspass object, but still the metadata of each mspass object should have the storage mode and other relevant keys (e.g., dir, dfile) defined. Also, I think we should at least make it optional for `write_files` to return dataframe instead, which will make it the exact reverse of `read_files`. Also, `write_to_db` should be a reverse to `read_to_dataframe`.

wangyinz (changes requested):
Very good, but there are still some minor issues. Please correct them in the inline comments

python/mspasspy/io/distributed.py
```diff
@@ -0,0 +1,1322 @@
...
+from mspasspy.db.schema import DatabaseSchema, MetadataSchema
+from mspasspy.util.converter import Textfile2Dataframe
+import dask.bag as daskbag
+import dask.dataframe as daskdf
+import dask
+import pyspark
+from pyspark.sql import SQLContext, SparkSession
+
+
+def read_distributed_data_new(
```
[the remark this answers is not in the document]
Aristoeu: The old one in database.py is removed. But many tutorials import `read_distributed_data` from database.py, and it should also be changed.

Aristoeu:
Cleaned the history

PR #383, opened 2023-01-26
Extend resample

pavlis:
This branch adds a resample function that is designed for use in a map operation.   The function uses a decimator when that is feasible but switches to resampling when a decimator is not possible (upsampling or noninteger decimation factor).   Also added tests for the new function.

Note I messed up when I created this branch.  I did a pull a few days ago to master but failed to create a new branch at that time.  My commits were added to master before I forked this branch.   I hope this doesn't create problems in the commit history.

### DIFF
--- a/python/mspasspy/algorithms/resample.py
+++ b/python/mspasspy/algorithms/resample.py
@@ -76,6 +76,55 @@ class BasicResampler(ABC):
             self.dt = 1.0 / sampling_rate
             self.samprate = sampling_rate
 
+    def target_samprate(self):
+        return self.samprate
+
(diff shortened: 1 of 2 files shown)


pavlis:
I would note one other thing about merging this branch.   At present the previous resample.py module sphynx documentation that should have been created for the resample module is on mspass.org.   There may be an index entry that needs to be added in the docs tree that I'm not aware of.

pavlis:
I think this is ready to merge.  I didn't add tests to the mspass_func_wrapper but presume that isn't necessary as they have their own tests.

PR #387, opened 2023-01-30
Update window.py

pavlis:
I ran across this code when I was trying to remember now to use the mspass_func_wrapper and noticed:
1. I does not have docstring
2. It is not correct as previous version of this had a bug in how it set the start of time window.   You can see the one line change I made.  It may break the tests.

I actually think this function is not needed.   To my mind it would be harder to remember a function with a similar but different name than to just remember what a TimeWindow object is or use a start and end time explicitly set as two parameters.

### DIFF
--- a/python/mspasspy/algorithms/window.py
+++ b/python/mspasspy/algorithms/window.py
@@ -382,6 +382,9 @@ def WindowData_with_duration(
     alg_id=None,
     dryrun=False,
 ):
+    """
+    THIS FUNCTION NEEDS A DOCSTRING.   I DON'T WANT TO WRITE IT AS I DON'T SEE WHY IT EVEN EXIST.  THERE IS A BUG IN THE CODE BELOW.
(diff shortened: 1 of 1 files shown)


pavlis:
Yup, tests broken as I expected.  I thought the tests were checking for the wrong answer that the previous code was returning.   I think the correct fix is to just remove that entire function and the tests that reference it.

wangyinz:
Thanks for catching this. I re-reviewed PR #387 against current `master` (`45325873`). The old `WindowData_with_duration` helper targeted by this patch no longer exists on `master`; the windowing API has since been refactored around explicit `WindowData` / `WindowDataAtomic` start-end handling.

This PR is now 461 commits behind `master`, is not mergeable, and the patch includes a placeholder docstring for code that has since been removed or reworked. I recommend closing this as superseded rather than merging the stale patch.

PR #398, opened 2023-02-25
Improve Window - cleaned up

pavlis:
There was a pending pull request that had been sitting for almost a year called "improve_window" that I just deleted.  This is a replacement for that pull request.  The reason the other was never done was it had improperly mixed in some python code for the normalize module which caused all kinds of conflicts in a simple merge.  The normalize module was stabilized some time ago so all I did was pull some changes in the C++ code for windowing to this new branch.   I expect all the texts here should come up clean as the only changes in in one C++ file that should not break anything.

As a reminder the fix here was purely an efficiency change.   The previous code called the copy constructor before windowing data.   That caused unnecessary memory bloat when cutting up very long files like day volumes.   @wangyinz assuming this doesn't break any tests you should quickly look what I did and read the comments I added before merging this branch.

### DIFF
--- a/cxx/src/lib/algorithms/slice_and_dice.cc
+++ b/cxx/src/lib/algorithms/slice_and_dice.cc
@@ -3,6 +3,7 @@
 #include "mspass/algorithms/TimeWindow.h"
 #include "mspass/seismic/TimeSeries.h"
 #include "mspass/seismic/Seismogram.h"
+#include "mspass/seismic/keywords.h"
 namespace mspass::algorithms
 {
 using namespace std;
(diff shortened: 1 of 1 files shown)


pavlis:
@wangyinz I urge you to simply merge this branch when you get a chance.  One oddity to look at however, is why the python formatter created an empty pull request.   That is, as far as I can tell it generated a pull request with no changes flagged and no new commits.   I expected a null result as the changes here were only in c++ code.  I don't think it should have generated that format pull request at all in this situation.   I would just delete the phantom pull request, but you probably should verify I'm reading that right and figure out why it did what it did.

PR #404, opened 2023-02-27
fix no history in new io module

Aristoeu:
Set objects alive and load history in `read_to_dataframe` method

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -4333,6 +4333,7 @@ class Database(pymongo.database.Database):
         if not format or format == "binary":
             try:
                 # create directory if not exists
+                dir = os.path.abspath(dir)
                 if not os.path.exists(dir):
                     os.makedirs(dir)
                 # This function has overloading.  this might not work
--- a/python/mspasspy/io/distributed.py
+++ b/python/mspasspy/io/distributed.py
@@ -197,7 +197,7 @@ def read_distributed_data(
 
 
 def read_to_dataframe(
-    db,  # db
+    db, 
     cursor,
     mode="promiscuous",
     normalize=None,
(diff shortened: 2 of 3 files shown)


pavlis:
The latest fix still generates an error with even a basic save_data using gridfs.   I think the section you have for history has an issue.  I got this message when the workflow did it's first call to db.save_data using gridfs:
```
InvalidDocument: cannot encode object: ProcessingHistory({'status': ProcessingStatus.UNDEFINED, 'uuid': 'UNDEFINED', 'algorithm': 'UNDEFINED', 'algid': 'UNDEFINED', 'stage': -1, 'type': AtomicType.UNDEFINED}), of type: <class 'mspasspy.ccore.utility.ProcessingHistory'>
```
Note the job wrote data to gridfs as the fs.files and fs.chunks collections are both not empty while the wf_TimeSeries collection this should have written is empty.   That may help you figure out the logic.  

I'm going to try the same workflow with file io to see what it does.

pavlis:
The file version fails for a different reason.   It didn't get as far.  In this case it might be classified as a user error, but not really.   The following section fails:
```
if not format or format == "binary":
            try:
                # create directory if not exists
                if not os.path.exists(dir):
                    os.makedirs(dir)
                # This function has overloading.  this might not work
                foff = _fwrite_to_file(mspass_object, dir, dfile)
            except MsPASSError as merr:
                mspass_object.elog.log_error(merr)
                if kill_on_failure:
                    mspass_object.kill()
                return 0, 0
```
with this message posted to the jupyter notebook:
```
/usr/local/lib/python3.6/dist-packages/mspasspy/db/database.py in _save_data_to_dfile()
   4335                 # create directory if not exists
   4336                 if not os.path.exists(dir):
-> 4337                     os.makedirs(dir)
   4338                 # This function has overloading.  this might not work
   4339                 foff = _fwrite_to_file(mspass_object, dir, dfile)

/usr/lib/python3.6/os.py in makedirs()
    218             return
    219     try:
--> 220         mkdir(name, mode)
    221     except OSError:
    222         # Cannot rely on checking for EEXIST, since the operating system

FileExistsError: [Errno 17] File exists: '/wftmp'
```
What happened here is the "dir" parameter I passed to Database.save_data was "./wftmp".   We clearly need a different, more elaborate algorithm to handle how "dir" is treated.   First thing to do is do some separate tests and reading about how to handle a directory name that has  "./" as the name.   I'm pretty sure I can make this run right now by changing the "./" to a valid path, but I'll let you know.

pavlis:
Confirm changing the dir to /home/wftmp gets the process past the file writing.  It puts data in a set of files but now exits with the same message as above when it enters the section handling history.   

Note this workflow never enables any history so I wonder if you are simply failing to test if the history container is empty or not.  I don't remember how to do that, but that may be the solution for that issue.

Aristoeu:
Thanks for testing! I failed to test `save_data` after `read_distributed_data`, I add this test and found the problem is I saved a "history" attribute in the metadata. When saving the metadata to the database, this attribute should be deleted.
To handle the directory name issue, I use `os.path.abspath(dir)` to first convert `dir` to an absolute path. This should handle a directory name that has "./".

PR #410, opened 2023-04-12
fix ensemble converter bug

pavlis:
This branch fixes a serious bug in handling ensemble data.  When obspy functions were applied to ensembles the previous version  obspy functions would create output that caused
the database save method to abort.   The problem was related to
a subtle mishandling of an internal set container used to handle the
conversion of ensemble Metadata.  This commit fixes the problem
but hasn't been verified with pytest.  It shouldn't fail but could.  My fix was tested on a subset of the 2012 usarray data and it solved the problem.   Look at my fixes with a bit of skepticism to see if there is a cleaner solution.  The way the decorators are interwoven made this hard to track so changes from what I suggest should be done cautiously.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -2821,13 +2821,13 @@ class Database(pymongo.database.Database):
         Reads an subset of a dataset with some logical grouping into an Ensemble container.
         It has the same fucntion as read_ensemble_data(), but is more efficient in reading binary files.
         The improvement is achieved by avoiding duplicate open and close for the same file.
-        MsPASS Objects are grouped firstly by files they are in. If two objects are in the same file, 
+        MsPASS Objects are grouped firstly by files they are in. If two objects are in the same file,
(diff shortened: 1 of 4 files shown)


wangyinz:
Thanks for tracking this down. I re-reviewed PR #410 against current `master` (`45325873`). The core issue this addressed, temporary `CONVERTER_ENSEMBLE_KEYS` metadata leaking through ObsPy ensemble conversion/copy-back, appears to already be handled on `master`: `converter.py` cleans the key in `Stream2TimeSeriesEnsemble` and `Stream2SeismogramEnsemble`, and `decorators.py` ignores/erases the key in `timeseries_copy_helper` and `seismogram_copy_helper`.

This branch is now 422 commits behind `master`, is not mergeable, and its `timeseries_copy_helper` version drops the `ts1.data = ts2.data` copy that current `master` preserves. I recommend closing this as superseded rather than merging it. A fresh follow-up regression test for the temporary-key cleanup would still be useful if we want to lock the behavior down.

Issue #409, opened 2023-04-10
undocumented limitation in Database.read_ensemble_data_group


pavlis:
I had a set of miniseed files with the data assembled into files organized by event.  i.e. each file had the data associated with a single earthquake.   I tried to use the new `read_ensemble_data_group` as I thought it might improve performance.   That function behaved in a somewhat unexpected way.   It am not totally sure what it did as I only have the result from a workflow that tried to do something with that data.  Looking at the code I suspect it read garage into data arrays with fread of raw miniseed packets into the vector container for the TimeSeries objects it was trying to load into an ensemble.   

As  minimum this is an undocumented limitation of this method.   It is probably better called a bug as the code should handle format variations and I don't think it does. It seems to be making an assumption the data are raw binary dump (fwrite) of the array of doubles that is the TimeSeries object container for sample data.  It is ok to make that assumption BUT ONLY IF the documentation says that loudly and forcefully.   

An initial fix for this could just test for a format attribute in the metadata of each member of the ensemble before trying to load the data.

wangyinz:
The current ensemble reader handles this format distinction explicitly. `Database._group_mdlist` groups members by both `storage_mode` and `format`; only file-backed `binary` groups use the optimized raw `_fread_from_file` path, while formatted files such as MiniSEED are passed through the atomic formatted reader. Mixed storage/format groups are supported by `_construct_ensemble`. Closing as completed.

PR #413, opened 2023-04-16
added doc and warning for read_ensemble_data_group()

Aristoeu:
Updated doc and added check for file format in `read_ensemble_data_group()`. The new function can only read from binary files. #409

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -47,7 +47,7 @@ jobs:
         echo "SPARK_HOME=/opt/spark-3.0.0-bin-hadoop2.7" >> $GITHUB_ENV
         export PATH=$PATH:/opt/spark-3.0.0-bin-hadoop2.7/bin
         echo "/opt/spark-3.0.0-bin-hadoop2.7/bin" >> $GITHUB_PATH
-        python -m pip install --upgrade --upgrade-strategy eager pyspark pytest-spark pybind11
+        python -m pip install --upgrade --upgrade-strategy eager pyspark pytest-spark pybind11[global]
     - name: Install
       run: python -m pip install --global-option=build --global-option=--debug .
     - name: Test with pytest
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -35,7 +35,7 @@ jobs:
     - name: Install Python Dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install --upgrade --upgrade-strategy eager pytest pytest-cov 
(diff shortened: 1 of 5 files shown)


Aristoeu:
An error about pybind11 occurs when [installing mspass](https://github.com/mspass-team/mspass/actions/runs/4715478781/jobs/8362495556), although I didn't change anything about pybind11:
```
By not providing "Findpybind11.cmake" in CMAKE_MODULE_PATH this project has
        asked CMake to find a package configuration file provided by "pybind11",
        but CMake did not find one.
      
        Could not find a package configuration file provided by "pybind11" with any
        of the following names:
      
          pybind11Config.cmake
          pybind11-config.cmake
      
        Add the installation prefix of "pybind11" to CMAKE_PREFIX_PATH or set
        "pybind11_DIR" to a directory containing one of the above files.  If
        "pybind11" provides a separate development package or SDK, be sure it has
        been installed.
```
Adding `pip install pybind11[global]` in `python-package.yml` can solve it, but I have no idea what happens here.

wangyinz:
The error is from:
```
      [ 22%] Performing download step (git clone) for 'pybind11'
      Cloning into 'pybind11-src'...
      HEAD is now at 80dc998e chore: bump versions for 2.10.1
      [ 33%] Performing update step for 'pybind11'
        File "/opt/hostedtoolcache/Python/3.8.16/x64/bin/pip3", line 5, in <module>
          from pip._internal.cli.main import main
      ModuleNotFoundError: No module named 'pip'
        File "/opt/hostedtoolcache/Python/3.8.16/x64/bin/pip3", line 5, in <module>
          from pip._internal.cli.main import main
      ModuleNotFoundError: No module named 'pip'
      gmake[2]: *** [CMakeFiles/pybind11.dir/build.make:124: pybind11-prefix/src/pybind11-stamp/pybind11-update] Error 1
      gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/pybind11.dir/all] Error 2
      gmake: *** [Makefile:91: all] Error 2
```
which is at the update step of CMake to install pybind11. The command running here is:
https://github.com/mspass-team/mspass/blob/a6b046800ac4e33fe3f143715717088850ba6662/cxx/cmake/pybind11-download.cmake#L14
I guess the question here is why the `pip3` command failed in the build. This makes no sense. Also, I think it is probably OK to remove the `pytest` installation here. I don't think it is necessary. Please remove it and test out whether that is true.

Aristoeu:
It works! It seems pip3 is the cause, removed the pip3 command and there is no failure.

PR #415, opened 2023-04-17
Add pyproject.toml

Yangzhengtang:


wangyinz (changes requested):
The new pyproject.toml file looks good. But I am still confused on what has been changed in the build process here. It seems to be that the requirement.txt was still used in both the GitHub Action workflow and the Dockerfile build. I thought we should remove that and use the toml file instead. 

Also, the log of the Action still shows the following message
```
DEPRECATION: --build-option and --global-option are deprecated. pip 23.3 will enforce this behaviour change. A possible replacement is to use --config-settings. Discussion can be found at https://github.com/pypa/pip/issues/11859
```
which I was aiming to get updated in this branch. Could you please fix that one too? Thank you!

wangyinz:
OK, I see what the issue here is. There seems to be no support in pip to install just the dependency from the toml file: https://github.com/pypa/pip/issues/8049

However, it seems to me that dependency is automatically handled by pip when using the toml file. Could you please confirm? If so, we can safely remove the requirement file. Alternatively, we can do something in this answer: https://stackoverflow.com/questions/73600082/how-to-reference-a-requirements-txt-in-the-pyproject-toml-of-a-setuptools-projec . The goal here is to have only one single list of dependency.

PR #419, opened 2023-04-22
fix cflags, sse2 and avx2 when building numcodecs in arm

Aristoeu:
When running MongoDB on arm, there were flags to disable sse2 and avx2, because of ENV variables defined in dockerfile. 
Removed the ENV variables and only set it inside a RUN command. They should not be present in the resulting Docker image.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -84,7 +84,7 @@ jobs:
         if: ${{ github.event_name != 'schedule' }}
         with:
           context: .
-          platforms: linux/amd64
+          platforms: linux/arm64
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -79,12 +79,12 @@ jobs:
 
       # Build and push Docker image with Buildx (don't push on PR)
       # https://github.com/docker/build-push-action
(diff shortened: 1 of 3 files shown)


Aristoeu:
Updated Dockerfile can be built successfully in arm, but there is a new [issue](https://github.com/mspass-team/mspass/actions/runs/4770892355/jobs/8485843142) when building Dockerfile_mpi. I didn't change anything about it, the latest test also runs well. Maybe the problem is with the base container.

wangyinz:
It seems somehow the `container-base` labeled the ubuntu20 as the latest instead of the ubuntu22 (see [here](https://github.com/SeisSCOPED/container-base/pkgs/container/container-base)). I think it is probably due to the wrong setting of the latest tag in the workflow. Could you please go there and remove the unnecessary "latest" tag in that workflow? Thank you!

Aristoeu:
Yes, the latest tag is not correct. Removed latest on ubuntu20 [here](https://github.com/SeisSCOPED/container-base/pull/6)

wangyinz:
Thanks for fixing it quickly! I have that PR merged just now.

Aristoeu:
if `cflags` is not defined, [numcodecs](https://github.com/zarr-developers/numcodecs/blob/2ad7a129e196b40bcbb389d6ec2b06bb5f169a81/setup.py#L12) add sse2 and avx2 flags, which are not supported in arm. So in dockerfiles we export cflags=-O3 if targetarch is arm.

numcodecs can not add support for arm now, related issue: https://github.com/zarr-developers/numcodecs/issues/431

PR #427, opened 2023-05-05
Add memory use methods

pavlis:
This branch adds a significant new capability that fixes a serious issue in previous versions.  Prior to this fix schedulers could not determine memory use correctly for seismic data objects.   The reasons are subtle but boil down to the fact that our data objects are implemented with C++ using pybind11 to create python bindings.   C++ dynamic memory allocation makes determining memory use from any object (class) that contains a dynamically allocated container difficult.   

Computing the exact memory use of a data object in C++ is expensive when the data contain std::map, std::multimap, and std::set containers like the mspass TimeSeries and Seismogram objects.   I used an approximation here where such containers use a constant multiplied by the count returned by their size method to estimate the memory use by such containers.   The assumption is the dominate memory use is the arrays used to store the sample data.  Those are computed exactly.   

The C++ code implements a method called `memory_use` in each of the following data objects:  `TimeSeries, Seismogram, TimeSeriesEnsemble,SeismogramEnsemble`, and `TimeSeriesWGaps`.  (Note `SeismogramWGaps` has not been implemented at all yet and exists only as as an include file defining the api.)  I used the pybind11 wrapper code to alias the memory_use method call to the special symbol `__sizeof__` that sys.getsizeof uses to estimate memory use in dask.   An important gap we don't know is if spark also uses sys.getsizeof to manage memory.

I implemented a C++ test program to do the most through tests of this new feature.   That was done because at the C++ layer I could do more and thorough tests because the python bindings to no provide access to some key symbols needed for the tests.  I only added some simple tests to test_ccore to validate sys.getsizeof behaves as expected.


pavlis:
PS to initial comment.   Once we assimilate this branch we need a way to verify it really solves the memory problems we were seeing with dask.   When that is verified we need to document this change in two ways:
1.  Add comments to the issues page related to memory problem (I didn't look up the number).
2. Add a section to the user manual on memory management.  Now that we understand this problem we can write something coherent.

Issue #407, opened 2023-03-08
Multitaper Estimator does not match mtspec


pavlis:
I am leaving this for the record for any current or future user's utilizing the multitaper spectral estimation code in mspass.   I am the author of that set of code (in C++) so I need to fix this problem.   

The issue is that it is very clear something is wrong with the code in the C++ class MTPowerSpectrumEngine with bindings to python with the same name.   I discovered this while working on magnetic data  I was processing with MsPASS for a paper I'm working on looking at seismic noise in the ultra low frequency band with periods of 20 min and longer.   The science problem is beside the point, but what I needed to do was compute spectra I could compare to seismic data spectra computed previously  with Prieto's mtspec program.  Krischer has a github repository for python bindings to the original FORTRAN implementation found [here](https://github.com/krischer/mtspec).   I loaded mtspec and processed the exact same signal with MTPowerSpectrumEngine and mtspec's mtspec function (package name and the main function are the same) and I got the figure below:

![mspass_multitaper_mtspec_compare](https://user-images.githubusercontent.com/184871/223691031-a55af5e1-4c63-452e-ad1a-abc721fe36e5.png)

The blue curve is the spectrum estimation with mtspec and the orange curve the spectrum computed by MTPowerSpectrumEngine.  They are obviously very different, although not absurdly so.   Direct calls to the C++ class and deconvolution algorithms based on multitapers should be consider broken until I resolve this issue.   I will need to do some digging and develop a specialized test to fix this problem.

Note I don't think an easy solution is to use mtspec to replace MTPowerSpectrumEngine.  The reason is that mtspec python is, I think, largely wrappers to create bindings for the FORTRAN code.   I doubt seriously it would function in dask/spark as it is unlikely they will work with pickle.  They might, but that will also require some testing.

pavlis:
For the record I think this problem is linked to two features of the multitaper method implemented in Prieto's mtspec package and possibly also the matlab implementation.  I can't be sure about that later as it isn't open source like Prieto's package is.   There are two fundamental issues I see that need to be addressed.
[...]

PR #428, opened 2023-05-07
Enhance multitaper engine

pavlis:
This branch has fixes to multitaper code to correctly handle zero padding in spectrum estimates.  I went way beyond what was needed for that and completely redesigned the PowerSpectrum object.   I also added new tests for new methods in PowerSpectrum.

This branch is a step forward to address problems raised in issue #407.  It is not complete but I urge it be merged once conflicts are resolved.  The final step needed is to assimilate support for German Prieto's multitaper library.  I have wrapper code to make it plug compatible but that feature need to resolve some issues I need Prieto's help to resolve.   After that the plug in will need some tests.  The related issue for not putting that revision here is it will require a change to the docker file to install prieto's code in the container.  We may want to put that into an alternative container than the base mspass container.


pavlis:
I see there is a problem with a prototype database handle I put into this branch.  That needs to be taken out for this intermediate fix.  It would be better added in a different branch and in a different place.

pavlis:
@wangyinz I think this branch is ready to merge.   Not sure why the codevcov tests are flagging this as failure or if that matters.   I assume you can sort that out.  It is mysterious to me. 

There also appears to be a black format issue detected downstream.  I thought I fixed those, but apparently I missed something.   Not sure what the preferred path to fix this is with the current setup of this repository.  Let me know if you want me to fix the black format issues.

PR #437, opened 2023-06-16
Fix splicing

pavlis:
This is an important bug fix.   The previous version of the C++ function splice_segments had a subtle, undetected bug that this branch fixes.   One of the dangers in multiple inheritance in OOP is method names can overlap.  In this case, the bug was caused by using the wrong "size" method reference for TimeSeries objects used in the algorithm. 

This also fixes two typos that caused cause the python merge function to abort in most situations.   

This particular bug fix is a case-in-point of why MsPASS needs a lot more exercising with real data with real data problems.   I found this while working on real data downloaded from IRIS with obspy.

@wangyinz if this passes tests as it should you should merge this asap.  This is a major bug that makes an important low-level function invalid.


pavlis:
@wangyinz unexpected error emerged from pytest run.   I hadn't expected that and failed to run a local pytest.  I'll deal with that directly and hopefully we can then merge this branch.

wangyinz:
The error in the test was due to the deletion of the data copy line in the `timeseries_copy_helper`, which makes the copied data empty. I got that fixed and had this branch merged.

PR #455, opened 2023-09-05
Adding Tacc specific startup script

jthet:


Dockerfile
```diff
@@ -203,6 +203,7 @@ RUN chmod +x /usr/sbin/tini
 
 # Add startup script
 ADD scripts/start-mspass.sh /usr/sbin/start-mspass.sh
+ADD scripts/start-mspass-tacc.sh /usr/sbin/start-mspass-tacc.sh
```
wangyinz, line 206: We don't necessarily want to include that script here in the main container image.

scripts/start-mspass-tacc.sh
```diff
@@ -0,0 +1,345 @@
...
+    ssh -q -f -g -N -R $STATUS_PORT:$NODE_HOSTNAME:8787 login$i > /dev/null 2>&1
+done
+# creating notebook password, giving link
+MSPASS_JUPYTER_PWD=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 20 | head -n 1)
+echo "Created reverse ports on Stampede2 logins"
+printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
+echo "Welcome to the MsPASS for Tapis" | sed  -e :a -e "s/^.\{1,$(tput cols)\}$/ & /;ta" | tr -d '\n' | head -c $(tput cols); echo -e "\n"
+echo  "The Link below will open a MsPASS Jupyter Notebook" | sed  -e :a -e "s/^.\{1,$(tput cols)\}$/ & /;ta" | tr -d '\n' | head -c $(tput cols);echo -e "\n"
+echo "http://frontera.tacc.utexas.edu:$LOGIN_PORT/lab?token=$MSPASS_JUPYTER_PWD"| sed  -e :a -e "s/^.\{1,$(tput cols)\}$/ & /;ta" | tr -d '\n' | head -c $(tput cols);echo -e "\n"
+printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
```
wangyinz, line 31: To my understanding, the lines here are newly added, right? There is some new handling of the Jupyter's URL compared to the one in https://github.com/mspass-team/mspass/blob/23898f45753603606c529b0034e3bac7eaf80ac2/scripts/tacc_examples/single_node.sh#L31-L42

What I would suggest is instead of handling the URL here inside of the startup script of the container, we should do it inside the job script instead (i.e., [single_node.sh](https://github.com/mspass-team/mspass/blob/master/scripts/tacc_examples/single_node.sh) and [distributed_node.sh](https://github.com/mspass-team/mspass/blob/master/scripts/tacc_examples/distributed_node.sh)). Then, we don't need to create a dedicated container for this specific purpose anymore. 

Obviously, this do mean that we need to change how the main container is built so that we could specify a password by giving it an environment variable, and I do agree that we should incorporate that change inside the mspass container in general. 

I can see that doing it inside of the job script means that we need some additional mechanism inside Tapis to change how containers are launched. We can discuss that further in our next meeting, but for the MsPASS repo, I figured it is better to use job script instead of creating dedicated container image.

wangyinz (changes requested):
Thanks a lot for contributing your code! This is very good and I actually learned something new here: the way of creating a random password for Jupyter is very cool. Please read my inline comments and let me know what you think. Thank you!

wangyinz:
This is out of date and the changes are merged in #615 already. Thanks a lot for your contribution and initiating this!

PR #463, opened 2023-10-22
fix seismogram array bug in fileio

Aristoeu:
#461 fixed the index error in TimeSeries ensemble, I forgot the case for Seismogram ensemble.

The fix is the same as #461, adding a foff(0) for the dead seismogram , it will not be used but can prevent the index error.


Aristoeu:
The error is fixed in #472

PR #468, opened 2023-11-07
Update README.md with a pointer for contribution guides.

JiaoMaWHU:
Update our home page per discussions in https://github.com/mspass-team/mspass/discussions/465


wangyinz:
Thanks for adding this!

PR #472, opened 2023-11-20
Major Revision of database and distributed

pavlis:
This is the major revision of the database.py and distributed.py module that has been in the works for months.  It was originally forked from a now ancient version of master so merging will be a challenge.   The list is long, but here is a summary of the main changes:
1.  The core `Database` class was drastically changed.  My original objective was to reduce redundant code I saw in the older code.  That is why this branch has the now misleading name "cleanup_database".   In the process I streamlined common code into some new private members.   I realized that those changes made it relatively easy to make the `save_data` and `read_data` methods more generic.  They now handle atomic and ensembles through the same methods.  The older versions with "ensemble" in their name were retained but now generate a "this method is deprecated" warning.  A big change in behavior for the `save_data` method is that by default it will now only return the ObjectId of the document that defines that datum (ensembles return a list of ObjectIds by default).  There is a new "return_data" boolean that can be set true to mimic the current behavior.  
2. Handling of ensembles is now standardized in Database and made to be more like atomic data.  That includes changes that make ensembles more like atomic data wrt the live/dead concept.   Old scripts with ensembles will break with this version as "live" is no longer a method but an attribute of the class.   
3. The use of the `Undertaker` class has been extended. It should now be viewed, as the name suggests, as the standard way to handle dead data.   The Database class no contains an Undertaker instance as `self.stedronsky` (Stedronsky was the local undertaker when I was growing up in rural South Dakota - a bad programming joke).  
4. The distributed.py module has been almost completely rewritten.  Like Database the readers and writers now handle ensembles as well as atomic data bar/rdd containers.   The implementation details are important but best punted to "read the docstring".    New revisions in the user manual in a related branch also document the new api.    An important new feature is saves (`write_distributed_data`) for both atomic data and ensembles use a bulk write to reduce database traffic on a save.  Both also add a new "post_elog" boolean that can be used to further reduce database traffic by posting error log data as subdocuments in the wf documents.  Note that feature has not yet been added to the standard schema definition.   
5. The pytest file for distributed is completely new.   The one for database.py has not changed in content but has major changes to match revisions to the behavior of Database methods.   

There are probably others I have forgotten.  Point is there are huge changes here.  

I'm submitting this pull request, but currently there are some additional things that most definitely will need to be done:
1.  Some tests that work on my local system are failing on github.  I will need to resolve that.
2. All the python code needs reformatting with black.  I don't want to do that until I resolve 1. 
3. All the rest of the development team need to peruse this material and suggest any changes before we even attempt to merge this.


pavlis:
There are two issues I need help with:
1.   The first instance of this pull request is failing with a setup configuration error I do not comprehend.   I think @wangyinz is most likely to know how to fix this issue.
2.  Not yet pushed but I did some additional local runs of pytest on this entire branch.   I should have done that before, but only ran tests on things I had changed.   I ran into a generic problem I will expand on below.

The generic problem I'm seeing is I don't  think we are handing the setup and teardown of pytest scripts that require spark correctly.   In case someone in the group doesn't know this, spark hands around this object they call the "context" that, I guess, defines the configuration of the cluster a parallel job is to run in.   A confusing property of context is that it must be unique to a particular "run".   I think a confusion we all have is what "run" means ("run" is my name for this - not necessarily anything you will find online).   I think we are getting messed up in pytest runs by inconsistent handling of context.  A case in point is that I just tried to run pytest locally on a file I never changed in this branch.  The test files is "python/tests/test_reduce.py".   When I run it locally it fails with the following message:
```
__________________________________________________________ ERROR at setup of test_map_spark_and_dask __________________________________________________________
file /home/pavlis/src/mspass/python/tests/test_reduce.py, line 59
  def test_map_spark_and_dask(spark_context):
E       fixture 'spark_context' not found
>       available fixtures: anyio_backend, anyio_backend_name, anyio_backend_options, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, doctest_namespace, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```
The reason for that failure is that indeed that file has this function:
```
def test_map_spark_and_dask(spark_context):
    l = [get_live_timeseries() for i in range(5)]
    # add net, sta, chan, loc to avoid metadata serialization problem
    for i in range(5):
        l[i]["chan"] = "HHZ"
        l[i]["loc"] = "test_loc"
        l[i]["net"] = "test_net"
        l[i]["sta"] = "test_sta"
    spark_res = spark_map(l, spark_context)
    dask_res = dask_map(l)

    ts_cp = TimeSeries(l[0])
    res = signals.filter(
        ts_cp, "bandpass", freqmin=1, freqmax=5, object_history=True, alg_id="0"
    )
    assert np.isclose(spark_res[0].data, ts_cp.data).all()
    assert np.isclose(dask_res[0].data, ts_cp.data).all()
```
and at least in this branch the variable `spark_context` in the function def line is not defined in the file.   For this to ever have worked there must be something hidden in the pytest configuration that sets `spark_context`  for the pytest run.   If my guess is correct, I have no idea how to find where that is done so can someone help me out here?   There are a series of similar failures on my local branch that I think are all related to this same issue. 

Now something I learned in developing this branch is hinted by that error message.   The right way to handle this problem is with fixtures - at least multiple sources on the web advise that approach.   In particular, I can resolve the above error if I add at the top of the test file the following:
```
import pyspark
@pytest.fixture(scope="module")
def spark():
    sc = pyspark.SparkContext("local", "io_distributed_testing")
    yield sc
    sc.stop()
```
and modify the function that was generating the error above with this slight variant:
```
def test_map_spark_and_dask(spark):
    spark_context = spark
    l = [get_live_timeseries() for i in range(5)]
    # add net, sta, chan, loc to avoid metadata serialization problem
    for i in range(5):
        l[i]["chan"] = "HHZ"
        l[i]["loc"] = "test_loc"
        l[i]["net"] = "test_net"
        l[i]["sta"] = "test_sta"
    spark_res = spark_map(l, spark_context)
    dask_res = dask_map(l)

    ts_cp = TimeSeries(l[0])
    res = signals.filter(
        ts_cp, "bandpass", freqmin=1, freqmax=5, object_history=True, alg_id="0"
    )
    assert np.isclose(spark_res[0].data, ts_cp.data).all()
    assert np.isclose(dask_res[0].data, ts_cp.data).all()
```
Noting the only changes are the first two lines.   Also, to get the whole test file to run I had to do a similar change to several other test functions that reference `spark_context`.  They all required exactly the same change.   It now runs, but I'm worried I'm going to break a bunch of other things if I change this file and others generating this kind of error.   I really think this is the right way to do all these tests, but until I understand the logic of how this worked before I do not want to change the other test files having this same issue when I run pytest locally.

pavlis:
This branch is getting close to being ready to merge.   Current status:

1. I did some minor revisions locally before merging master with this one.   Merge with master was much easier than I had feared.   
2. The big residual problem is with spark.   After a major revision of my development system I now have a close clone to the master docker container setup for testing.   I pushed this version today to verify the tests fail on github the same way they do here.   That seems to be true.   There is some residual problem with initialization of the full test suite causing these problem.  When I run the single file python/tests/test_distributed.py by itself it succeeds with no errors.  When run in the full chain it fails.  I am guessing there is some initialization I don't understand in the overall setup that is doing a global initialization of SparkContext.  That is wrong and needs to be fixed, but I need help to do that as I can't find it. 
3. I need to go back and do some black formatting, but that is minor.

pavlis:
The latest update is close, but still has two residual problems I need help fixing:
1.   Assuming I get the same result as my local test two pytest files are failing.  Both errors are in sections of MsPASS I have never dealt with before.   Like too many of our test scripts there are no comments or docstring info to guide me so I am tossing it back to someone else in the group to fix three offending sections.  (warning;  checks are in progress as I write this 3 may be wrong - that is what happened locally).   
2.  Outside pytest I was working on our tutorial notebooks.   I discovered a problem with the new `Database.save_data` method that must be fixed.   When running the "getting_started" tutorial in "notebooks" I found that when that notebook downloads data with obspy's get_waveform method, converts the data to TimeSeries objects, and runs save_data, the channel information is stripped (net, sta, chan) when the default mode is "promiscuous".    Fixing the problem is problematic only because my local system remains a bit unstable after upgrading to Ubuntu 22.04 and I'm unable to make the local mspass installation run with my favorite ide spyder.   I'll hack on this and may solve it before item 1 is resolved, but I wanted to make it clear that that is a big problem that needs to be fixed before this should be considered for merging to master.

pavlis:
Testing this with our getting_started tutorial with the docker container revealed a blemish I don't now how to fix.   When running a dask parallel run I get a ton of warnings posted with this line:
```
/tmp/ipykernel_223/2302293414.py:46: DeprecationWarning: `alltrue` is deprecated as of NumPy 1.25.0, and will be removed in NumPy 2.0. Please use `all` instead.
```
Probably a setup change we need to fix.  I suspect the current docker container will generate the same error since the parts being referenced in this branch are now identical to master.   Not sure how you can test that other than running the same tutorial.

pavlis:
@wangyinz can you look at the latest run here and see if you can figure out what is wrong?   Spark is throwing some error and I'm not sure how to even debug the error.

wangyinz:
The error is actually at https://github.com/mspass-team/mspass/actions/runs/7409175173/job/20158904073#step:11:314 in the error logs. It says the root cause is `Database.read_data:  arg0 has unsupported type=<class 'mspasspy.ccore.seismic.TimeSeries'>`. I can see why it errors out this way. Previously we have the following:
https://github.com/mspass-team/mspass/blob/5ec0bd0176ea14206296fcbaa7a09722aae305ab/python/mspasspy/db/database.py#L483-L486
This way, it supports not only dict but really any types that supports the `[ ]` operator. It is leveraging the duck typing feature of Python. However, in this branch, we have strict type check instead and that's reason why it failed. While the TimeSeries or Metadata does appear to be a dict, it is not a dict type object. Well, I think what we should do here is to relax the strict type checking for dict and change that into the old way to accept any dict-alike types.

pavlis:
Ok that explains that error.   BTW the reason I didn't catch it is I get a different error when I run pytest locally.  I have some weird version skew problem that is causing a "codex error" in the Database constructor.   I forget that and didn't even look at the output on github to realize it contained a different error.   Warning is that you may see some additional failed pushes before I resolve this. 

A different issue is I don't think the approach used in the old version for duck typing is the right one for this particular problem.  Part of that is a constraint I put on what the new implementation of `read_data` assumes about what it is given.   It needs to define operator[] (In C++ language) but testing to fetch a specific attribute, "_id", seems not very robust.   I'm also not sure just defining operator[] is enough any longer.  The new `doc2md` function centralizes the behavior and the input must resolve to a means to run this command:  `md=Metadata(doc)` that is the default promiscuous mode.  In cautious and pedantic mode it uses a loop over the keys and runs `md[k] = doc[k]` which pretty much means operator [].   

The fix I made is to allow dict or Metadata in the same isinstance test.   It will then also accept a TimeSeries or other seismic data object as an input pattern because all are subclasses of Metadata.   In the version I just made, however, I warn in the docstring that you should not use a seismic data object as arg0 to `read_data` as it is a bit intrinsically dangerous as it violates an implicit assumption of the algorithm.

pavlis:
Hmm - I do not quite understand why the tests just failed.  The last commit was some minor changes with black to a few test files,.  Now it is failing with a file exists error.  @wangyinz how are such things managed on github?  I think this is in the tests for the new distributed module where it tests reading and writing to files.  I thought I hat the fixtures set up to remove files on exit and it seemed to do that locally, so I'm a bit puzzled.  Not sure what the solution is.  

On a different note I see someone in the group (maybe me) need to write a pytest script to test Undertaker.  It is being tested only indirectly by database reader and writer tests handling killed data.  We really should have a comprehensive test file.

wangyinz:
This is a weird one. You can see that the "pull_request" workflow failed but the "push" workflow ran through. I can see the difference is [here](https://github.com/mspass-team/mspass/actions/runs/7415202255/job/20177877873?pr=472#step:3:432). It actually did a automatic merge into the master branch and that is what is failing. I guess this is really just telling us there will be problems if not merging it carefully.

PR #485, opened 2024-01-17
Fix bug in handling starttime and endtime

pavlis:
Previous version was failing to set starttime and endtime on saves.  Also data loaded and cut had incorrect starttime and endtimes saved because they werne't being replaced.


wangyinz:
fix in #486

Issue #484, opened 2024-01-17
endtime handling 


pavlis:
I don't know if we had this issue before or if I created it with the new Database implementation that was just merged to master.   What I'm finding is that save_data is not saving an important metadata attribute with the key "endtime".   That absolutely needs to be stored in wf documents as a very very common search is to select waveforms in a time interval.  If endtime is not defined that becomes really awkward.   

@wangyinz confirm you agree that saving endtime should be always done and I'll fix it.

pavlis:
I looked at the code and realized this was an easy patch to fix.  I just did it here on github and issued a pull request. 

Note the problem was even worse than stated above.  If a datum was windowed and then saved the start time would also be wrong because it wasn't being replaced in the document constucted form the datum.

wangyinz:
This was fixed in the save path.

The current `save_data` implementation always regenerates `starttime` from `mspass_object.t0` and `endtime` from `mspass_object.endtime()` (unless explicitly excluded), so windowed data cannot retain stale timing metadata. Database tests assert the saved timing fields.

Closing as completed.

PR #486, opened 2024-01-17
Fix bug in handling starttime and endtime

wangyinz:
This is fixing #485 #484

### DIFF
--- a/data/yaml/mspass.yaml
+++ b/data/yaml/mspass.yaml
@@ -66,6 +66,10 @@ Database:
           concept:  Time shift applied to define relative time of 0.0
           aliases:  t0shift
           constraint: required
+        endtime:
+          type: double
+          concept: Time of last sample of data (epoch time or relative to some other time mark)
+          constraint: optional 
         utc_convertible:
           type:  bool
           concept:  When true starttime_shift can be used to convert relative time to UTC.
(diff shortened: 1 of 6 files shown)


wangyinz:
I am not sure what to do with the error. The test that failed basically set `starttime` in `exclude_keys`. With the new changes, `exclude_keys` do not work on `starttime` and `endtime`. While I don't think people would ever do that, it does complicates things as `exclude_keys` now has exceptions.

wangyinz:
Maybe those keys should be added in `md2doc`? What do you think? @pavlis

pavlis:
Three attributes in wf documents are special and need to be handled with care.   They are:  starttime, endtime, and npts.  The reason they are so special in MsPASS is that all three must be closely linked with three internal attributes that are part of the C++ class (attributes of the class):  npts, dt, and t0.  The endtime attribute is more of a sanity check as endtime is a method of `BasicTimeSeries` that is the superclass of both `TimeSeries` and `Seismogram`.   If we have tests listing any of these attributes int he exclude list the tests need to be changed.   

These cannot be handled in md2doc.  The reason is that the "md" in md2doc is for Metadata.  The starttime attribute in Metadata can become disconnected with the value of the "t0" attribute in a TimeSeries or Seismogram.   I think a problem can happen because one can set t0 with a constuct like `d.t0=t` or `d.set_to(t)`.  The later resets the "starttime" Metadata field but the first does not.   For that reason `save_data` must force starttime.   endtime is different because it is always derived, but is worse because i could also be stale when any of t0, npts, or dt change.

PR #494, opened 2024-01-20
bug fix for write_distributed_data

pavlis:
This fix should correct the bug I was chasing yesterday that caused me to start the discussion topic #493.   The previous version was not handling dead data correctly and was failing when the database routine _save_sample_data was referencing a null matrix write_distributed_data was creating due to a mistake in the logic of handling dead data.  

I tested most of this code by hacking the container and resolving the problem with the parallel workflow.  I did not change the test script for distributed.py to test this condition.   We should do that later.  To get this fixed I plan to verify it recreates the successful run on a part of the 2012 usarray data set using the volatile container this pull request will create.  I'll leave a note in this pull request when I've verified that.  Before that, however, be aware I didn't even do a syntax check on this so there could be dumb errors.  Point is do nothing until I give a green light.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -6358,6 +6358,9 @@ class Database(pymongo.database.Database):
                         foff = fh.tell()
                     f_byte = io.BytesIO()
                     for i in range(len(mspass_object.member)):
+                        # silently skip any dead members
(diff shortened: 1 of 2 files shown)


pavlis:
Green light to merge this with master.   Fixed the problem I was having with write_distributed_data.

PR #496, opened 2024-01-21
Fix typo

pavlis:
Fix a typo where argument had "file" instead of "files".   The plural is what was defined previously in the api and is what Database expects.


pavlis:
@wangyinz I think I made this fix backwards.   Looking back in the repository I think we may have used `storage_mode=="file"` instead of `storage_mode=="files"` that this set of commits was trying to impose.   

I think this needs to be backward compatible.  If you confirm my observation that it seems we used the singular (i.e. "file") before I'll drop this branch and fix it the other way.

wangyinz:
yeah... that's right. It's always been "file" here.

PR #499, opened 2024-01-27
Fix typos incorrectly using plural of file

pavlis:
This branch fixes a typo using the plural in the `storage_mode` argument.   Code in database.py and docstrings of distributed.py incorrectly used the plural "files" instead of "file".

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -1099,7 +1099,7 @@ class Database(pymongo.database.Database):
             an abstraction that even in FORTRAN days hid a lot of complexity.
             A call to this function supports multiple save mechanisms we
             define through the `storage_mode` keyword. At present
-            "storage_mode" can be only one of two options:  "files" and
+            "storage_mode" can be only one of two options:  "file" and
             "gridfs".  Note this class has prototype code for reading data
             in AWS s3 cloud storage that is not yet part of this interface.
             The API, however, was designed to allow adding one or more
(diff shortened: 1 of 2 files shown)


pavlis:
@wangyinz this is failing for a mysterious reason on an import error.

wangyinz:
That error is due to the updated moto module, which was out today, that changed the api. You can see the change here: https://github.com/getmoto/moto/blob/5.0.0/CHANGELOG.md. I fixed that accordingly. However, now the distributed test also failed...

wangyinz:
So, the new error is at line: https://github.com/mspass-team/mspass/blob/f72037521d4a70c6047a8b59c95f6488ae7cfe33/python/mspasspy/db/database.py#L7394

which complains: 
```  
File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/mspasspy/db/database.py", line 7394, in _construct_ensemble
    ens_tmp, abortions = self._load_ensemble_file(this_mdl)
TypeError: Database._load_ensemble_file() missing 3 required positional arguments: 'object_type', 'dir', and 'dfile'
```

and that makes sense because the `_load_ensemble_file` is defined as: https://github.com/mspass-team/mspass/blob/f72037521d4a70c6047a8b59c95f6488ae7cfe33/python/mspasspy/db/database.py#L7187

I don't understand how this was not failed before.

pavlis:
Ah. I see where this is coming from now.   This is happening in a new section of save_data for ensembles that adds a "feature" that was not present before.  The docstrings in the chain explain it somewhere but the idea is to efficiently save data in an ensemble with mixed storage modes.  The main idea i to sort out all data going to a single file so we can use the faster ensemble write written in C.  This particular problem is an example of something I noted in recent meetings.  This write of Database has untested features.   This is one of them - ensembles with mixed storage modes.   

I may fork a completely new branch to address this and create the tests I should have made in the first place.   The good news is the bug won't effect most workflows and there is a (slow) workaround to not set storage_mode to anything but the default

PR #500, opened 2024-01-28
Bug fix handling data_tag

pavlis:
This fixes a bug created by an api mismatch between Metadata and dict.  Code used erase on a dict which throws an error when it is executed.


wangyinz:
This one failed for the same reason as #499

PR #501, opened 2024-02-01
bug fix to Database._convert_mseed_index

pavlis:
This is a bug that refuses to die.   Thought we had killed in months ago.   Adds "endtime" attribute to dict returned by this method.


pavlis:
I just noticed we had a partial fix of this only two weeks ago in a different part of the new Database class implementation.   This is in a completely different algorithm.  Should be able to merge this once an issue in this master is resolved with a mock module.

PR #502, opened 2024-02-03
Fix bug in handling ensemble files with mixed storage mode

pavlis:
I think this will fix the bug in reading ensembles from files.   I haven't run these through pytest yet so may require a second iteration to fix this.   This is a pretty serious bug as reading from miniseed files fails until this is fixed.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -7261,9 +7261,9 @@ class Database(pymongo.database.Database):
         # Because this is private method that should only be used
         # interally we do no type checking for speed.
         nmembers = len(mdlist)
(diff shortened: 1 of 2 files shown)


pavlis:
Hmm - looks like the fixes I made now pass the existing tests.  Something is failing still, however, with spark setup in a completely different test file.  Looks to me like it may be related to the mock_s3 version change but I'm not at all sure.

wangyinz:
The problem has nothing to do with mock_s3, but I don't understand it neither. The problem is that the monkeypatch over SparkConf that worked before doesn't for now. I don't see how that could happen, but I fixed it by patching the SparkSession.builder instead.

PR #508, opened 2024-02-11
Fix data gap

pavlis:
This started out as a simple bug fix for a mistake in the pybind11 code for this auxiliary class.   I realized, however, that in working with a prototype for surgical mutes that this was a very useful container that might work well as a component of a generic surgical mute algorithm.   Working on the prototype with a data set from Homestake I realized it lacked a couple important capabilities:
1.  The previous version required gaps intervals to not overlap.  An attempt to add a gap definition that overlapped an existing one would cause nothing to happen.   That happened because the original concept was recorded data gaps, which are effectively always immutable.   For a more generic use that needed to be relaxed.  I did that here by adding some minor complexity to the add_gaps method.   Tests show it that functionality now works.
2. I saw a need for the C++ operator += (need for the using += in python of course) that would cleanly merge two different DataGap containers.  Once functionality 1 was added this was trivial.   Well, not really as there was an odd property of the std::set container that complicated the code requiring a complete copy of the left hand side before merging the right hand side.  What I did works, but I'm not quite sure why it was necessary.   I considered it harmless, however, as this container should rarely be huge enough that such a copy operation is a serous performance problem.

### DIFF
--- a/cxx/include/mspass/seismic/DataGap.h
+++ b/cxx/include/mspass/seismic/DataGap.h
@@ -73,6 +73,16 @@ This function provides a common mechanism to define such a gap in the data.
       could be used to add that functionality.
       */
       void clear_gaps(){if(!gaps.empty())gaps.clear();};
+      /*! Return number of defined gaps. */
+      int size() const{return gaps.size();};
+      /*! Return the subset of gaps within a specified time interval. 
+       *
+       * \param tw TimeWindow defining range to be returned.  Note 
+       * overlaps with the edge will be returned with range outside the
+       * range defined by tw.  If the tw is larger than the range of 
+       * the current content returns a copy of itself. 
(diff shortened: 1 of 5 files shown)


pavlis:
Looks like these changes all passed the tests clean as I expected.  Wanted to comment that the lines Codecov flags are harmless.  They are present because I intentionally omitted the list list of TimeWindow constructor.   It was a relic of the older seispp api and seemed less useful for the python api.   If you, @wangyinz, think it would be  useful thing to add it wouldn't be hard.  More important is the TODO you will see the pybind11 code - this should really have a pickle api.   Won't really need that, however, unless I decide to move forward with this prototype surgical mute using this as the container.  Until then I'd suggest this thing is kind of a niche component.

PR #513, opened 2024-02-22
Fix graphics

pavlis:
Two things in this set of changes:
1.  Fixed a long standing bug that messed up scaling of plots in the "wtvaimg" mode.    Fix may not be perfect, but is an improvement over present form that produces really bad plots in that mode most of the time.   The possible residual issue is that we may need to handle the "aspect" parameter for image plotting more elegantly.   I'm not sure with these changes a user can actually change the aspect ratio manually.   I don't expect that to be a very common need though so I didn't think the effort required to test all the possibilities was worth the trouble.
2. I changed the default style to "wtva" (wiggle trace variable area) instead of "wtvaimg" (the style that previously was giving trouble).   Maybe should be the simplest "wt" but in most cases wtva plots are a better data visualization that wt - an option though.

### DIFF
--- a/python/mspasspy/graphics.py
+++ b/python/mspasspy/graphics.py
@@ -1038,7 +1038,7 @@ class SeismicPlotter:
             origin_position = "lower"
         pyplot.imshow(
             work,
-            aspect='auto',
+            aspect="auto",
             cmap=self._color_map,
             origin=origin_position,
             extent=extent,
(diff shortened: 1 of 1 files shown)


pavlis:
@wangyinz I urge you to merge this branch with master.   It fixes a long standing bug in graphics with the "wtvaimg" style of plot.   The only check that is flagging anything is coverage.  It is complaining the new sections aren't being tested.   I have no idea how to create a pytest function to test this kind of functionality.   The best way to test the graphics module is to run the graphics tutorial and verify it works correctly.   Anyway, I would urge you to review what I did and merge this branch with master.

PR #515, opened 2024-02-23
Build conda package and upload to anaconda

Aristoeu:
Added a workflow to build initiate conda environment, build conda package, and upload conda package to anaconda.

`meta.yaml` and `build.sh` are configuration files to build conda package. `MANIFEST.in` is a necessary file to build python package. 

For now the conda package is uploaded to [my account](https://anaconda.org/cxwang/mspasspy), and the token is written explicitly, it should be a secret defined in github repo.

I directly build the conda package from source code, rathen than building the wheel and uploading the wheel to pypi. I tried to build wheels for manylinux, but there are still problems:
- cannot find -lpython3.8. It seems manylinux docker containers don't have the lpython, and no `libpython3.8.so` or `libpython3.8.a`. I didn't find a good way to install them when building the wheel
- openblas error. It still throws many errors after I specified the lowest target version


Aristoeu:
Different python versions, test using docker, then arm

Aristoeu:
https://anaconda.org/cxwang/mspasspy

Aristoeu:
Tried to change cmake version to debug the slash missing bug for arm64, but the version is always set to 3.18.4-2+deb11u1

Issue #511, opened 2024-02-19
problem with rotate method and function


pavlis:
I discovered what could either be called a bug or a feature in the `rotate` method of `Seismogram`.  It carries over to the rotate function which has a wrapper defined for map operators.   

If could be called a "feature" if I make the documentation clearer.  It turns out the angle of rotation is positive clockwise.   That is the standard convention for azimuth angles, but is reversed for spherical coordinates.  i.e. in very math course the angle phi spherical coordinates is positive in the anticlockwise direction.   Stock geographic coordinates we use in seismology always have the x (x1) axis positive in the local east direction and y (x2) positive in the north direction.   This small code fragment will show you that the convention used for rotate is positive clockwise and measured in radians:
```
from mspasspy.ccore.seismic import Seismogram
from mspasspy.algorithms.basic import rotate_to_standard,rotate
import numpy as np
import math

d=Seismogram(4)
d.dt=1.0
d.set_live()
phid=30.0
for j in range(d.npts):
    d.data[0,j]=math.cos(np.radians(phid))
    d.data[1,j]=math.sin(np.radians(phid))
phi=-30.0
print(d.data)
dtrans = rotate(d,np.radians(phi))

print(dtrans.data)
```
If you run that you'll see the input data has vectors that are  [sqrt(2),0.5,0] and after rotate they become [1,0,0].  The math side is clear.  For seismology this would come up if the "radial" direction for local wave propagation is at 60 degrees azimuth ( convention with R + in the direction of propagation.  In that case P waves would have horizontal motion mainly in the R-Z plane.  Standard vertical rotation convention is RTZ meaning x,y,z->R,T,Z.  That means after the rotation x (x1) should point in the R direction.   Confusing signs, I know, but the little test script shows the way you would transform the coordinates to RTZ would be to compute the azimuth difference between E (azimuth of 90) and the direction of propagation (+R).   For above the azimuth of the propagation direction is 60 degrees so the azimuth relative to E is -30 degrees because azimuth is measured positive clockwise.  

The other perspective is this is a bug from mixing up the forward an inverse transform of the transformation matrix for this problem.   The standard formula for rotation by angle phi with the normal math convention is:
```
|  a   -b   0 |
|  b   a    0|
| 0    0     1|
```
[...]

Issue #512, opened 2024-02-20
Segmentation fault during ctest in local build


JiaoMaWHU:
When compiling mspass locally, I was able to finish
```
cmake ..
make
make install
```
and also installing the python package
```
pip3 install ./ -v
```

However, when I ran `ctest` (this command is included in the [cmake workflow file](https://github.com/mspass-team/mspass/blob/master/.github/workflows/cmake.yml#L42)), the following error happened,
```
ctest --rerun-failed --output-on-failure                      
Test project /Users/jiaoma/gitCode/mspass/cxx/build
    Start 6: test_bundle
1/1 Test #6: test_bundle ......................***Exception: SegFault  0.07 sec
Starting test_bundle program
Input ensemble summary
net sta chan loc npts fill_value
AA ABC HHN 00 1
AA DEF HHN 00 2
AA BBLONG HHN 00 3
AA ABC HHN EMPTY 4
AA DEF HHN EMPTY 5
AA BBLONG HHN EMPTY 6
XY ABC HHN 00 7
XY DEF HHN 00 8
XY BBLONG HHN 00 9
XY ABC HHN EMPTY 10
XY DEF HHN EMPTY 11
XY BBLONG HHN EMPTY 12
AA ABC HHZ 00 13
AA DEF HHZ 00 14
AA BBLONG HHZ 00 15
AA ABC HHZ EMPTY 16
AA DEF HHZ EMPTY 17
AA BBLONG HHZ EMPTY 18
XY ABC HHZ 00 19
XY DEF HHZ 00 20
XY BBLONG HHZ 00 21
XY ABC HHZ EMPTY 22
XY DEF HHZ EMPTY 23
XY BBLONG HHZ EMPTY 24
AA ABC HHE 00 25
AA DEF HHE 00 26
AA BBLONG HHE 00 27
AA ABC HHE EMPTY 28
AA DEF HHE EMPTY 29
AA BBLONG HHE EMPTY 30
XY ABC HHE 00 31
XY DEF HHE 00 32
XY BBLONG HHE 00 33
XY ABC HHE EMPTY 34
XY DEF HHE EMPTY 35
XY BBLONG HHE EMPTY 36
Running best case test for bundle_seed_data
net sta chan loc npts fill_values
XY DEF 00 20 8 32 
XY DEF EMPTY 23 11 35 
XY BBLONG 00 21 9 33 
XY BBLONG EMPTY 24 12 36 
XY ABC 00 19 7 31 
[...]

PR #516, opened 2024-02-24
Extend 3c transforms

pavlis:
Fixes and extensions to address Issue #511.   Five major changes/additions:
1.  Adds new transform_to_RTZ function
2. Adds new transform_to_LQT function
3. Changes the api for free_surface_transformation to add more flexible input.  Default now assumes fetching required data from Metadata to make the function easier to use in a map operator. 
4. Complete pytest functions for the above. 
5. Cleaned up a confusion on the sign of the angle of rotation for the overloaded C++ function.  That is now largely hidden behind the higher level functions anyway but the most important fix is making the sign convention unambiguous in the doxygen comments.

Note I'm pushing this to verify all the changes match my local test runs.    I want to do a final sanity check against data before suggesting this really be merged.

### DIFF
--- a/cxx/src/lib/seismic/CoreSeismogram.cc
+++ b/cxx/src/lib/seismic/CoreSeismogram.cc
@@ -621,7 +621,11 @@ void CoreSeismogram::rotate(const double nu[3])
 /* simplified procedure to rotate only zonal angle by phi radians.
  Similar to above but using only azimuth angle AND doing a simple
  rotation in the horizontal plane.  Efficient algorithm only
- alters 0 and 1 components. */
+ alters 0 and 1 components. 
+
+Note sign is spherical coordinate form with phi positive anticlockwise.
+Sign in rotate with spherical coordinate is different because phi is 
+converted to azimuth.  Note that is just the sign of b in the code.*/
 void CoreSeismogram::rotate(double phi)
 {
     if( (u.size()[1]<=0) || dead()) return; // do nothing in these situations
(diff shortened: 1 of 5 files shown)


pavlis:
Oh yes, forgot to run the C test function.   That should also adress issue #512.

pavlis:
Assuming the current set of tests pass on github this branch is ready to merge.   The pytest functions for the new set of Seismogram wrapper functions are fully tested, although we should verify if the coverage gizmo flags anything.  I also tested them with my tutorial notebook on three component data transformations and I get the answers I expect.  Hence, unless something unexpected pops up in the tests currently running I advise you, @wangyinz, to merge this branch with master.

pavlis:
Well, obviously that last push didn't pass the pytest run.   Rather mysterious, but my fault for not explicitly running the test under pytest and not checking the exception handler code.   I forgot that that only could be tested under pytest.   

Working on that so ignore this until you see it pass all these tests.   Sorry, @wangyinz , for being overconfident this was ready to merge

pavlis:
@wangyinz  I think you should merge this branch.   I noticed codecov flagged some missing testing for error handler lines that do kill if the rotate_to_standard throws an exception.   I am 100% sure those handlers work.   There is one other that issues a rare complaint message that is very unlikely to ever be executed.   I'm sure the syntax is right and it won't fail if used.   Adding a test for such a rare event seems unnecessary to me.  If you disagree I'll try to create something.

PR #517, opened 2024-03-01
initial commit for adding the p wave pick function

JiaoMaWHU:
## Summary
- This change implements a P wave arrival-time pick function based on the probability prediction from machine learning model.
  - It leverages the SeisBench PhaseNet model for the probability prediction. 
  - A probability threshold is used to choose picks.
  - Time window is supported for focusing on a certain period of the input.

## Testing
- unit test added for inputs that include mseed file, and trace object obtained using obspy.clients.fdsn.Client

### DIFF
--- a/python/mspasspy/algorithms/ml/arrival.py
+++ b/python/mspasspy/algorithms/ml/arrival.py
@@ -25,7 +25,7 @@ def annotate_arrival_time(
     :param timeseries: The time series data to predict the arrival time.
     :param threshold: The probability threshold (0-1) to filter p-wave picks. \ 
         Any picks with probability less than the threshold will be removed. Default value is 0.2. 
-    :param time_window: The relative time window to filter the predicted arrival time. \ 
+    :param time_window: The time window (in utc timestamp) to filter the predicted arrival time. \ 
         If not provided, the whole time series will be used.
     :param model: The model used to predict the arrival time.
     :param model_args: arguments to initialize a new model if not provided
--- a/python/mspasspy/algorithms/ml/arrival.py
+++ b/python/mspasspy/algorithms/ml/arrival.py
@@ -1,11 +1,12 @@
 import numpy as np
 import seisbench.models as sbm
 
+from seisbench.models.base import WaveformModel
 from obspy.taup import TauPyModel
 from obspy.geodetics import gps2dist_azimuth, kilometers2degrees
 from mspasspy.ccore.seismic import TimeSeries, TimeSeriesEnsemble
-from mspasspy.ccore.algorithms.basic import TimeWindow, WindowData
-from seisbench.models.base import WaveformModel
+from mspasspy.ccore.algorithms.basic import TimeWindow
+from mspasspy.algorithms.window import WindowData
 
 
 def annotate_arrival_time(
@@ -41,11 +42,15 @@ def annotate_arrival_time(
         # 'stead' model was trained on STEAD for 100 epochs with a learning rate of 0.01.
         # use sbm.PhaseNet.list_pretrained(details=True) to list out other supported models
         # when using this model, please reference the SeisBench publications listed at https://github.com/seisbench/seisbench
-        pretrained_model = "stead" if model_args["name"] == None else model_args["name"]
+        pretrained_model = "stead" if (model_args == None or "name" not in model_args) else model_args["name"]
         model = sbm.PhaseNet.from_pretrained(pretrained_model)
 
     # apply the window if provided and convert time series to stream
-    windowed_timeseries = WindowData(timeseries, time_window.start, time_window.end) if time_window else timeseries
+    windowed_timeseries = (
+        WindowData(timeseries, time_window.start, time_window.end)
+        if time_window
+        else timeseries
+    )
     ts_ensemble = TimeSeriesEnsemble()
     ts_ensemble.member.append(windowed_timeseries)
     stream = ts_ensemble.toStream()
--- a/python/mspasspy/algorithms/ml/arrival.py
+++ b/python/mspasspy/algorithms/ml/arrival.py
@@ -8,7 +8,12 @@ from seisbench.models.base import WaveformModel
 
 
 def annotate_arrival_time(
-    timeseries: TimeSeries, model: WaveformModel = None, model_args: dict = None
+    timeseries: TimeSeries,
+    threshold=0.2,
+    window_start_offset_sec=0,
+    window_length_sec=None,
+    model: WaveformModel = None,
+    model_args: dict = None,
 ):
     """
     Predict the arrival time of the seismic phase using the provided seisbench WaveformModel.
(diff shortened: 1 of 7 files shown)


python/mspasspy/algorithms/ml/arrival.py
```diff
@@ -8,7 +8,12 @@
 
 
 def annotate_arrival_time(
-    timeseries: TimeSeries, model: WaveformModel = None, model_args: dict = None
+    timeseries: TimeSeries,
+    threshold=0.2,
+    window_start_offset_sec=0,
```
JiaoMaWHU, line 13: we can use the timewindow object here

python/mspasspy/algorithms/ml/arrival.py
```diff
@@ -8,7 +8,12 @@
 
 
 def annotate_arrival_time(
-    timeseries: TimeSeries, model: WaveformModel = None, model_args: dict = None
+    timeseries: TimeSeries,
+    threshold=0.2,
```
JiaoMaWHU, line 12: 0.2 may be adjusted in threshold default value

python/mspasspy/algorithms/ml/arrival.py
```diff
@@ -44,40 +49,33 @@ def annotate_arrival_time(
...
-    degdist = kilometers2degrees(georesult[0] / 1000.0)
+    # Step 5: Convert arrival times to a more convenient format if necessary
+    filtered_arrival_time_strs = [time.strftime("%Y-%m-%d %H:%M:%S.%f") for time in filtered_arrival_times]
 
-    arr = model.get_travel_times(
-        distance_in_degree=degdist, source_depth_in_km=srcdep, phase_list=["P"]
-    )
-    atime = srctime + arr[0].time
-    d["Ptime"] = atime
+    timeseries["p_arrival_picks_time"] = filtered_arrival_time_strs
```
JiaoMaWHU, line 81: css 3.0 arrival table uses float data format and absolute time, it could be possible for us to use relative time

JiaoMaWHU:
to fix the docker issue, we can try

1. exclude cuda library from seisbenc
2. clean up cache: https://github.com/mspass-team/mspass/actions/caches
3. try removing https://github.com/mspass-team/mspass/blob/7f959aa19e4fc3393bced4061067032df7b6db95/.github/workflows/docker-publish.yml#L108C1-L109C59

JiaoMaWHU:
Hi @wangyinz please take a look at this pr, it's ready for review. Thank you.

python/mspasspy/algorithms/ml/arrival.py
```diff
@@ -0,0 +1,80 @@
...
+def annotate_arrival_time(
+    timeseries: TimeSeries,
+    threshold=0.2,
+    time_window: TimeWindow = None,
+    model: WaveformModel = None,
+    model_args: dict = None,
+):
+    """
+    Predict the arrival time of the P wave using the provided seisbench WaveformModel.
+    The arrival time will saved as metadata in the input TimeSeries object and can be accessed using \
```
wangyinz, line 21: will be saved

python/mspasspy/algorithms/ml/arrival.py
```diff
@@ -0,0 +1,80 @@
...
+    :type timeseries: mspasspy.ccore.seismic.TimeSeries
+    :type threshold: float
+    :type time_window: mspasspy.ccore.algorithms.basic.TimeWindow
+    :type model: seisbench.models.base.WaveformModel
+    :type model_args: dict
+    """
+
+    # Check the input arguments
+    if not 0 <= threshold <= 1:
+        raise ValueError("Threshold must be between 0 and 1")
```
wangyinz, line 38: instead of raising an error, maybe this should be a warning instead and we will set the threshold to 0 or 1 if the range is exceeded.

python/mspasspy/algorithms/ml/arrival.py
```diff
@@ -0,0 +1,80 @@
...
+        model = sbm.PhaseNet.from_pretrained(pretrained_model)
+
+    ts_ensemble = TimeSeriesEnsemble()
+    ts_ensemble.member.append(timeseries)
+    stream = ts_ensemble.toStream()
+
+    # apply the window if provided and convert time series to stream
+    start_time = stream[0].stats.starttime
+    windowed_stream = (
+        stream.trim(starttime=start_time + time_window.start, endtime=start_time + time_window.end)
```
wangyinz, line 55: Is this assuming a relative time window is being used? I am a little confused on how stream.trim works. In our data object, we do support both relative time and absolute time reference.

python/mspasspy/algorithms/ml/arrival.py
```diff
@@ -0,0 +1,80 @@
...
+    # Step 2: Find all the index with probability value greater than the threshold
+    indices = np.where(data > threshold)[0]
+
+    # # Step 3: Calculate the corresponding time
+    p_wave_picks = trace.times()[indices]
+
+    # Step 4: Save the arrival time in the relative format
+    # Adjust the arrival time if a time window is provided
+    if time_window:
+        p_wave_picks += time_window.start
```
wangyinz, line 79: This is related to the relative time question above. This should at least be explained in the comments.

python/tests/algorithms/ml/test_arrival.py
```diff
@@ -0,0 +1,82 @@
...
+from mspasspy.util.converter import Trace2TimeSeries
+from obspy import Stream, UTCDateTime
+from obspy.clients.fdsn import Client
+
+sys.path.append("python/tests")
+
+from helper import get_live_timeseries
+
+pn_model = sbm.PhaseNet.from_pretrained("stead")
+
```
wangyinz, line 16: Maybe we should add a test with real data. We do have some real waveform data in the [data dir](https://github.com/mspass-team/mspass/tree/master/python/tests/data), and you might want to reuse one of them for the test. I think this will be helpful with the relative time vs absolute time confusion I had above. Also, please add some necessary comments to the test code too - I know we were not doing great commenting the test code in the existing code base, but we should improve it from now on.

setup.py
```diff
@@ -98,5 +98,8 @@ def build_extension(self, ext):
     ),
     package_data={"": ["*.yaml", "*.pf"]},
     include_package_data=True,
-    install_requires=["pyyaml"],
+    install_requires = [
+        "pyyaml",
+        "seisbench==0.4.1" if sys.version_info < (3, 9) else "seisbench>0.4.1"
+    ]
```
wangyinz, line 104: Is this still needed? I thought the new setup just used the dependency from the requirement.txt. I would think the pyyaml one was a relic from previous versions.

python/mspasspy/algorithms/ml/arrival.py
```diff
@@ -0,0 +1,80 @@
...
+    if model == None:
+        # 'stead' model was trained on STEAD for 100 epochs with a learning rate of 0.01.
+        # use sbm.PhaseNet.list_pretrained(details=True) to list out other supported models
+        # when using this model, please reference the SeisBench publications listed at https://github.com/seisbench/seisbench
+        pretrained_model = "stead" if (model_args == None or "name" not in model_args) else model_args["name"]
+        model = sbm.PhaseNet.from_pretrained(pretrained_model)
+
+    ts_ensemble = TimeSeriesEnsemble()
+    ts_ensemble.member.append(timeseries)
+    stream = ts_ensemble.toStream()
```
wangyinz, line 50: It seems to me that we should support directly applying to a ensemble object.

wangyinz (changes requested):
Thank you so much for drafting up this new feature! I think this is a great start. Please don't forget to rebase the code after you address the comments.

python/mspasspy/algorithms/ml/arrival.py
JiaoMaWHU: could you elaborate a bit more? seisbench model does require a stream input

python/tests/algorithms/ml/test_arrival.py
JiaoMaWHU: sure, I can use real data to test and provide more comments for the test cases.

are you aware of any existing code that imports those data into mspass formats? that would be really helpful to me

setup.py
JiaoMaWHU: sure

python/mspasspy/algorithms/ml/arrival.py
wangyinz: Yeah, that is confusing... I am actually talking about the `tref` in our base object: https://www.mspass.org/_static/html/classmspass_1_1seismic_1_1_basic_time_series.html#abb88ddef3bab3b3205dfe6c6cfdb91aa

There is a short section in our document that explains it: https://www.mspass.org/user_manual/data_object_design_concepts.html#handling-time

python/mspasspy/algorithms/ml/arrival.py
JiaoMaWHU: todo saving the prob as well
```
pick -> {time, prob, ....}

[v] pick_time -> {prob, ....}
```

python/mspasspy/algorithms/ml/arrival.py
JiaoMaWHU: timeseries and time_window all in UTC

python/tests/algorithms/ml/test_arrival.py
JiaoMaWHU: https://docs.obspy.org/packages/autogen/obspy.io.mseed.core._read_mseed.html

python/mspasspy/algorithms/ml/arrival.py
JiaoMaWHU: change the input as TimeSeriesEnsemble, so that picking for multiple ts can be done together

JiaoMaWHU: I would like to keep time series as the input in this pr. after kicking off the experiment we can add a new method to support timeseries_ensemble inputs

JiaoMaWHU:
Hi @wangyinz could you please review this pr? thank you

wangyinz:
Sorry for the delays! I just merged it.

requirements.txt
```diff
@@ -11,11 +11,12 @@ boto3
 botocore
 moto
 requests 
-setuptools
+setuptools<60.0
```
wangyinz, line 14: @JiaoMaWHU Could you please clarify why this is needed? Limiting the version of setuptools has created warnings from the Dependabot.

Issue #520, opened 2024-03-08
bug in merge


pavlis:
I am pretty sure I just found an off by one error bug in our merge function for handling continuous data.   Working on processing very long time series data from Homestake where the merge window was of the order of 3 months long.  The window happened to cross the 2015-2016 new year where all the input files have an break.   36 channels of data after merging all have a 1 sample gap at the year boundary.   

This is almost certainly an off by one error in the C++ code that does the heavy lifting for this function.  It is my problem to fix it, but I wanted to put this here for the record.

pavlis:
I am becoming skeptical this problem is caused by merge.  I am suspecting we may have an issue with our miniseed implementation.   I'm seeming some very inconsistent data.  A critical reason to be sure the problem is our implementation is that the gap I'm seeing at the year boundary in the data read by MsPASS does NOT appear in data indexed with Antelope's equivalent `miniseed2db`.  i.e. I can run their graphical display ("dbpick") at the year boundary are the data are continuous.   

There is definitely an issue with the index we produce with index miniseed.   The index for one channel has only two documents.  Here is the output I see that shows the issue:
```
for doc in cursor:
    print(UTCDateTime(doc['starttime']),UTCDateTime(doc['endtime']))

2015-03-25T16:36:24.000004Z 2015-12-31T23:45:44.000004Z
2016-01-01T00:01:12.000000Z 2016-11-30T07:47:33.000000Z
```
Notice there is about a 15 minute gap between the endtime of the first and the starttime of the second.   Weirder is obspy's reader, which we use to read the actual data files, gives a different answer.   Here is the output of different combinations of things produced by reading the data linked to the first document above.
```
d1=db.read_data(doc,collection='wf_miniseed')

print(UTCDateTime(d1.endtime()))
2015-12-31T23:06:46.000004Z

print(UTCDateTime(d1['endtime']))
2015-12-31T23:45:44.000004Z

print(UTCDateTime(d1['last_packet_time']))
2015-12-31T23:06:47.000000Z
```
The important thing is the that the output of the endtime method is different from the time defined by the "endtime" Metadata attribute by several minutes.   

So, the problem seems to actually be in the miniseed data.  The inconsistency of the obspy reader result and the time computed by the indexer are going to be particularly difficult to resolve.   I'll start with our indexer on these files to see if I can figure out what it is doing wrong,.

pavlis:
[...]

PR #523, opened 2024-03-22
Improve index mseed

pavlis:
Revisions to address #520.   

1.  Major reworking of miniseed C++ indexing function.
2. Add code to Database to fix bug in handling time from mismatch of output of our indexer and what  obspy's reader does.   Revision has the readers output always override the Database posting elog messages when that is necessary.  
3. Added a prototype C++ test code for indexer.   That is incomplete and subject to a decision about how to handle data files for testing and what additional tests are needed to certify the miniseed indexing and reading.


pavlis:
Hmmm  - someone else on the development team needs to pull this branch and see if you can recreate the pytest failures.  When I run the tests locally I cannot recreate the problem that is test_normalize is showing.    It doesn't make a lot of sense that normalize should have a problem anyway the changes in this branch have no conceivable way they could impact the test that is failing.  Very strange

pavlis:
Ignore that last request for an independent test of the test failure.   In trying to use this on some other data I discovered an off by one error.  I'm working to fix it.

pavlis:
There definitely was an off by one error that I think I have resolved in a version I have not yet pushed.  I need some help, however, before I can resolve why test_normalize is failing with this branch.   What I know is this:

- I honestly don't understand how `test_normalize` ever worked for the section that is failing - a test of `OriginTimeMather`.   The reason it is failing is that `OriginTimeMatcher` is trying to compare the "starttime" value of wf_miniseed documents (in a Dataframe and in MongoDB) with "time" values stored in the database "source" collection.   The problem is there are no records that match within hundreds of seconds.   What is mysterious is that the database the test is using, I think, is static and doesn't even use the new version of the code to index miniseed that this branch aims to clean up.   Why is it just failing now as I don't think the changes I've made impact that static data?
- I can't really fix this easily for a couple of reasons.  First, I have no idea where that data came from and what its properties are that make it appropriate for this test.   Hence, I can't try to recreate the database to see where a mismatch has occurred.    Second, I don't want to mess with this database as I fear I'll break something else.   

I can't resolve this easily without some help.   Who created the db_dump data in the first place?  What is it?  What other tests will I break if I change it?  I could work around the problem by changing the tests for `OriginTimeMatcher` but I think the tests I see are fine and do what they need to do.   I'm just confused by the data used for the tests.

PR #526, opened 2024-03-27
Improve datascope interface

pavlis:
New more generic handling of Antelope (Datascope) database tables.  The main class this adds should become the new basis for import and export from Antelope systems.   

Fellow developers should comment on the api design as I developed this largely independently.   Also check that the docstrings make sense and are instructive as they have also only been seen by me to date.   

I think the tests are fairly complete but we'll see what codedev says.


pavlis:
I see my initial tests were not even close.   Forgot about additional methods I added to the class that were of secondary use.  I was just testing the two main methods I thought most important.  Anyway, on this - next version will fix that inadequacy.

pavlis:
Codedev shows some rarely hit sections of this set of code that aren't being tested.   Would be a pain to address them as most would require finding a Datascope database with a particular set of problems error handlers are aiming to catch.   

There is also a bit of anomaly with how to handle a script I put into this repository that won't actually run without a license for the Antelope software.   That script belongs in the antelope contrib repository.   I'm going to put it there eventually, but we should have a copy here.  What I don't know for sure is how that should be handled within MsPASS.  A simple solution is to put a README file in the directory where the command line tool script is found.   Maybe there is some fancier, standard way to do that kind of thing on github I'm not aware of.  @wangyinz please decide how you want to handle this.   We could just duck this problem for now and punt it down the road until I get the same script in antelope contrib.

pavlis:
@wangyinz do not merge this yet.   In writing a new page for the user manual on this new class I realized there was a dumb mistake in the api for the df2table method.   It currently requires a Datascope database name but it should default to the name of the instance of the class from it is called.   When I fix that it should be ready to merge.

PR #528, opened 2024-04-10
Documentation edits

pavlis:
Adds a new page on generic problem of importing tabular data including new feature for handling Antelope called DatascopeDatabase.  Also fixes a number of small errors sphynx detected I didn't notice earlier.  

Note sphynx is still throwing some errors that are related to two issues:
1.  Some formatting errors in docstrings it is cross referencing
2. References to functions or classes not yet in master as the additions are in pending pull requests.


wangyinz:
The error from cmake seems to be a bug in their latest release and it has been fixed. Basically, it was giving the wrong path to the pybind11 we installed.

PR #533, opened 2024-05-05
Fix several bugs in OriginTimeMatcher

pavlis:
Fixes a couple bad bugs in OriginTimeMatcher.   If this passes tests as it should it should be merged immediately.

### DIFF
--- a/python/mspasspy/db/normalize.py
+++ b/python/mspasspy/db/normalize.py
@@ -2357,7 +2357,7 @@ class OriginTimeDBMatcher(DatabaseMatcher):
         t0offset=0.0,
         tolerance=4.0,
         query=None,
-        attributes_to_load=["lat", "lon", "depth", "time"],
+        attributes_to_load=["_id", "lat", "lon", "depth", "time"],
         load_if_defined=["magnitude"],
         aliases=None,
         require_unique_match=False,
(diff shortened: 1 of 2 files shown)


pavlis:
Obviously my assumption these fixes would pass the tests was incorrect.   Obviously not ready to merge

pavlis:
The latest push of this is still running, but the fixes here have revealed a couple minor issues that may come back to haunt us:
1.  There is a bit of an inconsistency in this api using a DataFrame input versus a MongoDB instance in the "_id" value that is special to MongoDB.   I had to modify our tests to match a change I had to make to OriginTimeMatcher to make it work correctly.   I am confident the change I made was the right one because it wouldn't work correctly with an id matching algorithm otherwise.  Loading _id is the correct default because 99% of uses should load the source data from MongoDB.   The only exception at this time is if one wanted to load a Datascope catalog.  I modified the docstring to give that warning, but that is really deeply buried in the documentation.
2. We probably should implement a "find_doc" method for the `OriginTimeMatcher` that this branch fixes.  If we did that it could be used with the `bulk_normalize` function, but if you use it now that function raises a MsPASS exception and aborts saying it needs a "find_doc" method.   The difference between find_one and find_doc are somewhat trivial.   I should have probably implemented it but didn't think it was a particularly high priority so punted it down the road.  If the group thinks that should be addressed now I can do that but it should probably done in separate branch as this branch addresses a critical bug fix - OriginTimeMatcher is completely broken until this is merged.

PR #536, opened 2024-05-15
Fixing issues with conda build

wangyinz:
This branch will build all conda packages within the container. This should simplify the process and avoid some unknown issues with the versioning of the conda package the master branch currently have.


wangyinz:
Turns out the versioning didn't work because of the default behavior of the checkout step. It only checkout the current commit from the repo, which breaks the assumption of the versioning mechanism where we need to use the tags to identify the number. There are more discussions here: https://github.com/actions/checkout/issues/217

PR #548, opened 2024-06-12
Bug fix in DatascopeDatabase constructor

pavlis:
Two line change to fix an error in the constructor of this class.   Not sure how the previous version passed pytest.


pavlis:
covdecov is just confirming why this bug wasn't discovered earlier.   The offending section seems to have not been covered by the tests.

pavlis:
PS:  I would ask @wangyinz to go ahead and merge this anyway as I want to use this feature for the upcoming short course.

Issue #551, opened 2024-07-22
WindowData option to consider 


pavlis:
In working on a new multichannel cross-correlation gizmo I realized a minor issue with our `WindowData` function.  It is currently completely dogmatic in handling a situation that probably should not be viewed as black and white.   The issue is what should `WindowData` do when the interval requested is bigger than the range of the data being windowed?  That could be either a starttime or endtime inconsistency.  Currently in either case `WindowData` will kill the output and log a message to elog.  

I would suggest we add an option to `WindowData` to be less dogmatic in this situation.   The current function signature is this:
```
def WindowData(d, win_start, win_end, t0shift=None, 
   object_history=False, alg_name='scale', alg_id=None, dryrun=False)
```
I suggest it become this:
```
def WindowData(d, win_start, win_end, t0shift=None, 
   # suggested additions 
    short_segment_handling=["kill","log"],
    object_history=False, alg_name='scale', alg_id=None, dryrun=False)
```

The name `short_segment_handling` as suggest may be a bit to complicated.  The idea is the pair of keywords define how this problem is handled (the alternative is two kwarg with different names).   The idea is that component 0 defines how the issue should be handled and the second the algorithm for recovery if appropriate.  The acceptable options would be

1. ["kill","log"] - the current behavior and the clear default to be backward compatible.
2. ["truncate","log"] truncate the output if required but log a complaint
3. ["truncate","silent"] truncate silently
4. ["pad","log"] zero pad undefined section to return the right number of samples expected from window range and log as complaint.
5. ["pad","silent"] same as 4 but do it silently
6. ["gap","log"] - return result as a `TimeSeriesWGap` object with nonoverlap defined as a gap.  Log a complaint
7. ["gap","silent"] - like 6 but do it silently.

That sounds complicated but would be pretty easy to implement.   It probably should be done as two different arguments rather than the pair as suggested.  0 is basically how to handle the problem and 1 is whether or not it should initiate an error log message.  Hence, rather than a pair when I think about this it should probably be:
```
def WindowData(d, win_start, win_end, t0shift=None, 
   # suggested additions 
    short_segment_handling="kill",
    log_recoverable_errors=True,
    object_history=False, alg_name='scale', alg_id=None, dryrun=False)
```
Noting name implies proper behavior here that only recoverable errors should be allowed to be handled silently.   

I can do this change as part of the extensive changes I'm doing anyway for the multichannel correlation and new deconvolution algorithm with very little effort if the group thinks this is a good idea.

[...]

PR #552, opened 2024-07-29
Extend window data

pavlis:
This set of changes implements suggestions proposed in issue #551.   This also required very large changes to the pytest file for the WindowData function.   I think the revisions are pretty complete, but we should review the coverage data before this is merged.   It passed pytest locally so hopefully the only holdup may be in any missed coverage.

### DIFF
--- a/cxx/src/lib/algorithms/slice_and_dice.cc
+++ b/cxx/src/lib/algorithms/slice_and_dice.cc
@@ -52,7 +52,8 @@ CoreSeismogram WindowData(const CoreSeismogram& parent, const TimeWindow& tw)
 	CoreSeismogram result(parent);
   result.u=dmatrix(3,outns);
 	result.set_npts(outns);
-	result.set_t0(tw.start);
+  /* Using the time method here preserves subsample timing.*/
+	result.set_t0(parent.time(is));
   // Perhaps should do this with blas or memcpy for efficiency
   //  but this makes the algorithm much clearer
   int i,ii,k;
@@ -106,7 +107,8 @@ CoreTimeSeries WindowData(const CoreTimeSeries& parent, const TimeWindow& tw)
 	CoreTimeSeries result(parent);
 	result.s.reserve(outns);
 	result.set_npts(outns);
-	result.set_t0(tw.start);
+  /* Using the time method here preserves subsample timing.*/
+	result.set_t0(parent.time(is));
 	// Necessary to use the push_back method below or we get leading zeros
 	result.s.clear();
 
@@ -175,7 +177,8 @@ Seismogram WindowData(const Seismogram& parent, const TimeWindow& tw)
 	would be lost.  Reason is the constructor uses CoreTimeSeries. */
 	BasicTimeSeries btstmp(dynamic_cast<const BasicTimeSeries&>(parent));
 	btstmp.set_npts(outns);
-	btstmp.set_t0(tw.start);
+  /* Using the time method here preserves subsample timing.*/
+	btstmp.set_t0(parent.time(is));
 	/* WARNING MAINTENANCE ISSUE:  this is less than ideal fix for a problem
 	found when debugging the revision of this algorithm to improve its
 	performance May 2022.  the constuctor called here assumes the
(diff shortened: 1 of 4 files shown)


pavlis:
The previous commit had no real errors but I had neglected to remove the debug scaffolding.   The version I just pushes fixes that mistake and should pass tests cleanly.    

The black formatter was detecting an error in this pull request but the error is not mine.  It is in db/client.py.   I think it is a relic of a last minute change before the Earthscope short course. @wangyinz  you may want to resolve that issue either in this branch or separately,.

wangyinz:
This looks good to me and I have merged the black format fix here since this is more straightforward. The coverage is only at 86.02%, but I think that is good enough.

wangyinz:
I fixed a minor issue on the inconsistent `alg_name`. There is still one issue: the `alg_name` of `WindowDataAtomic` will always be overridden and changed to `WindowData`, but I think this might be OK. What do you think?

pavlis:
I hadn't noticed that about `alg_name` but I see what you mean.   I think your changes handle this correctly,  If you run `WindowData` the `alg_name` will be set as "WindowData" by default but if you run `WindowDataAtomic` directly it gets set to "WindowDataAtomic".   I didn't actually test that but I think that is what will happen.   Hence, I think this is good.  

I also see now the coverage missed by the test functions.   All are rare but probably should be verified.   I'd kind of prefer to punt them down the road.   One is incorrect because the coverage gizmo can't know that the C++ functions should never throw an exception because the python wrappers detect and handle those issues preventing the C++ functions from ever being called with bad input.   the comments around the error handler note that point which is about the best we can do.  

So, I suggest you merge this unless you think I need to add test code for those uncovered error handlers.

pavlis:
@wangyinz hold off on merging this until I tell you it is ok.  I may have found an off by one error in the "pad" mode.

pavlis:
There is indeed a problem.    It is subtle floating point problem the tests didn't detect that effects the new algorithm for both "pad" and "truncate" mode.   It uses a floating point test for < or > of a window timer compared to data start or end time.   The actual interval needs to be rounded to the nearest sample.   When data come from UTC clocks the data start times are not synchronously sampled to the sample grid.   My tests didn't detect that because I was only using simulation data with synchronously sampled data.  

Let me fix this problem and extend the tests to handle the issue.   I'll try to improve the coverage since I need to repair the test functions anyway.   Glad I tried this on the new cross correlation stuff or I wouldn't have detected this problem.

pavlis:
Almost have this but while I remember note that the coverage gizmo is wrong about coverage.  The tests I have for arguments are flagged as untested but they actually are tested.  The problem is that they are actually intercepted by the decorators and whatever approach that gizmo uses it is confused by the redundant error handlers.  Consequently, all I'm adding for the next revisions is a test where I dither the start time by a fraction of a sample.

pavlis:
I fixed the somewhat minor issue to have the WindowData function preserve subsample timing.  Before we merge this though there is a point of discussion to consider.   The current implementation for ensembles has an undesirable side effect that I think is a bad idea.   In particular, this block of code runs the atomic version of window data on each member:
```
for i in range(len(mspass_object.member)):
                if mspass_object.member[i].live:
                    mspass_object.member[i] = WindowDataAtomic(
                        mspass_object.member[i],
                        win_start,
                        win_end,
                        t0shift=t0shift,
                        short_segment_handling=short_segment_handling,
                        log_recoverable_errors=log_recoverable_errors,
                        object_history=object_history,
                        alg_name=alg_name,
                        alg_id=alg_id,
                        dryrun=dryrun,
                    )
```
Because it uses the construct 
```
mspass_object.member[i] = WindowData()
```
and `mspass_object` is the input that does an in place modification.   That reduces memory footprint, which can be a factor with large ensembles as we know, but has an undesirable side effect that if you use a construct like this:
```
enscut = WindowData(ens,stime,etime)
```
after that runs enscut and ens will be references to the same object.  

That is definitely the wrong default behavior, but I could see making that an option.   If you concur I will add an "in_place_return" option for ensembles that would be False by default.   True would probably be the preferred option for map operators.

wangyinz:
Yeah, we shouldn't do in place return here. I think making that an option is definitely needed.

pavlis:
Ok, fixed that last problem.   I probably should have added a test for the new option, which I elected to actually call "overwrite_members" out of fear the symbol would collide with what I suggested above that is used in the mspass_func_decorator.   Probably wouldn't have mattered but I think that name is clearer anyway.  

I think this should be ready to merge now

pavlis:
This branch should be ready to merge.   The latest push was just a trivial format fix with black.   This should resolve the subtle rounding problem with the previous version we discussed in our last group meeting,.

PR #558, opened 2024-08-07
Addition of Multichannel cross-correlation processor for phase arrival time estimation

pavlis:
This is a new implementation of a batch version of my old dbxcor program for MsPASS.   The version I'm initially pushing is incomplete lacking two major things:
1.   I need to complete a separate test function for `align_and_stack` for the long list of exceptions it handles via either an exception mechanism or an error log entry.  There are a lot that have to be handled for this complex algorithm I added from my previous experience with dbxcor.
2. There are additional functions that are needed in association with the `align_and_stack` function that is the only thing currently finished.  These are:  (1) A simple generic function that takes the output of this function and computes measured arrival times using computed cross-correlation shifts.   The issue is making it "generic".   A topic for upcoming group meetings.   (2) There is a need for related function(s) to automate what dbxcor required the user to input.   The first one I'm planning is a envelope-based method to estimate what we called the "correlation window" in dbxcor.   The second is a generic function to extract an initial seed estimator for the robust stacking algorithm - the signal used as the seed for the first stage correlation.   That is a very different problem also open for discussion in the group.  Finally, there is a need for a picker to compute a time shift to apply to ensemble outputs (a third thing the user had to pick graphically in dbxcor). 

For now I'm pushing this branch to make sure pytest on github is consistent with my local test and to provide a copy for the group to discuss.


pavlis:
This is not yet ready to merge.   The tests are complete for the top-level function called `align_and_stack`.  Lacking are tests for other functions in this module not used directly by `align_and_stack`.  There are also handlers inside functions called by `align_and_stack` that aren't being fully tested for the range of things that can go wrong in this complex algorithm.   At our next group meeting we need to discuss how to proceed.

pavlis:
Assuming this commit passes the pytest runs as it did for me locally this is almost but not quite ready to merge.   The topic of discussion topic #562 needs to also be resolved.  There are some functions in new module this branch contains that are more generic and belong somewhere else.    

Another issue before this should be merged is I want to exercise this algorithm more extensively on real data.   I need to verify its behavior is what I expect when working with small events where the algorithm needs to automatically discard low snr data.   When I get a working container for this branch I want to do those tests before we put this in the distribution.   

When this feature is fully integrated and tested it justifies a new release.  If this works out well we should consider a publication on the implementation as it is would have wide use.

pavlis:
I think this branch is now ready to merge.   I suggest the following sequence:

1. @wangyinz should review this content and verify there isn't something I missed.  Also consider if the function names are rational.   One thing I did that I wasn't sure was correct is add the lines to have sphynx add the new stuff to the python api section of the user manual.
2. When satisfied merge the branch to master.
3.  I'm starting work on a user manual page and tutorials linked to this new material.   That is best done on a separate branch.   The tutorial will be a full workflow I'm going to develop to apply this new algorithm to a decent sized data set on an HPC system.  Getting this merged is pretty much a prerequisite to building that more easily.    I could use the dev container from this pull request and will do that if you want to hold up merging this for any reason.

wangyinz:
I don't see anything obvious so I merged it. It sounds all good to me. Thank you so much!

PR #567, opened 2024-10-04
CNRDeco Initial Version

pavlis:
This branch is incomplete but when ready we may want to assimilate it anyway.  It has these elements:
1.  Draft version of a new CNRDecon implementation to replace the prototype called CNR3CDecon
2. Bug fixes for the multichannel correlation module found while testing the CNR code
3. Some enhancements to the util/seismic
4. A new method for BasicTimeSeries inherited by all atomic data called "time_axis" useful for raw matplotlib plotting


pavlis:
@wangyinz the tests are failing on this branch because of the problem with the symbol "database" that i believe is known to be connected to a version skew with MongoDB.   I have no idea how to fix that here on github.   I could use your help to fix this.

wangyinz:
hmmm... I am not sure what that is, but if you rebase your code with the latest master, it shouldn't have any dependency issues. However, I looked at the logs and I can see that the code is actually failing the test is several places: 
```
FAILED python/tests/algorithms/test_MCXcorStacking.py::test_align_and_stack - assert 0.0031451296528622673 < 0.002
FAILED python/tests/algorithms/test_MCXcorStacking.py::test_align_and_stack_error_handlers - ValueError: v cannot be empty
FAILED python/tests/algorithms/test_RFdeconProcessor.py::test_RFdeconProcessor - TypeError: cannot pickle 'mspasspy.ccore.algorithms.deconvolution.MultiTaperXcorDecon' object
FAILED python/tests/test_ccore.py::test_TimeSeries - assert 104 == 100
```
You can see the details in https://github.com/mspass-team/mspass/actions/runs/11181775844/job/31086542446?pr=567, I'm actually not sure where the symbol issues are.

pavlis:
@wangyinz you may want to merge this to master subject to a couple of details to finalize:

1. To get the pytest suite to run with test_RFdeconProcessor.py I had to completely disable the pickle tests.  The reason is some time earlier in these revisions for the new "CNRDeconEngine" in this branch I modified the RFDeconProcessor class to cache an instance of the C++ object.   My recollection was I did that to match the way CNRDeconEngine works.  The reason to do so is efficiency as it always uses a multitaper algorithm that requires computing the Slepian tapers.   Slepian functions are expensive to compute being of the same order of magnitude as the deconvolution operation.  Hence, caching them makes sense.  It does, however, require the pickle operation on the operator.   I remember now the reason I changed RFDeconProcessor was that the multitaper algorithms in that class have the same issue.  Fixing pickle of that class faces exactly the same problem as CNRdeconEngine so I reasoned if I had to do one I might as well improve all of them.   RFDeconProcessor still works but cannot be run as a map function until this is resolved.  
2.  CNRDeconEngine is not fully tested.  There is no pytest setup at all.   My plan is to test it next on the same trial data set I've been using to test multichannel correlation function (son of dbxcor).  The point is that it is not ready for prime time but isn't too far from completion.   Putting it in master is not a huge problem as it is orthogonal to anything else as far as I know.
3. Perhaps the biggest reason to merge this branch is all the secondary problems I found in running the multichannel correlator code to prep the test data set for CNRDeconEngine processing.   You can look through the commits and see what I'm talking about.  

So, I'll let you decide if you want to merge this or wait until I've cleaned up CNRDeconEngine a bit more.   The key reason to merge is to include the side fixes I found in working on this branch.   If we do that I'll fork a new branch from master that will be more focused on cleaning up all the deconvolution code (including pickle) and not have all this other stuff mixed in.   Let me know if you want to do that before proceeding though as I need to fix a few black format issues - didn't do that before pushing the latest changes.

PR #568, opened 2024-10-20
Fix "sampling_rate" not updating along with "dt"

Cloud1e:
I fixed the issue of the attribute "sampling_rate" not updating when "dt" changes.
This happens in resample in resample.py and TimeSeries2Trace in converter.py.
I also edited the test code for the resample and it passed all the cases.

### DIFF
--- a/python/mspasspy/algorithms/resample.py
+++ b/python/mspasspy/algorithms/resample.py
@@ -229,6 +229,21 @@ class ScipyResampler(BasicResampler):
             )
             mspass_object.set_npts(n_resampled)
             mspass_object.dt = self.dt
+            # Check for "sampling_rate" attribute
+            if mspass_object.is_defined("sampling_rate"):
+                sampling_rate = mspass_object["sampling_rate"]
+                # Check if sampling_rate is consistent with 1/dt
+                if abs(sampling_rate - 1.0 / mspass_object.dt) > 1e-6:
+                    # Record inconsistency in error log (elog)
+                    message = "sampling_rate inconsistent with 1/dt; updating to 1/dt"
+                    mspass_object.elog.log_error("resample",
+                                                 message,
+                                                 ErrorSeverity.Complaint)
+                # Update sampling_rate to 1/dt
+                mspass_object["sampling_rate"] = 1.0 / mspass_object.dt
+            else:
+                # Set sampling_rate to 1/dt
+                mspass_object["sampling_rate"] = 1.0 / mspass_object.dt
             # We have to go through this conversion to avoid TypeError exceptions
             # i.e we can't just copy the entire vector rsdata to the data vector
             dv = DoubleVector(rsdata)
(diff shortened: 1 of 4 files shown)


python/mspasspy/algorithms/resample.py
```diff
@@ -229,6 +229,21 @@ def resample(self, mspass_object):
...
+            # Check for "sampling_rate" attribute
+            if mspass_object.is_defined("sampling_rate"):
+                sampling_rate = mspass_object["sampling_rate"]
+                # Check if sampling_rate is consistent with 1/dt
+                if abs(sampling_rate - 1.0 / mspass_object.dt) > 1e-6:
+                    # Record inconsistency in error log (elog)
+                    message = "sampling_rate inconsistent with 1/dt; updating to 1/dt"
+                    mspass_object.elog.log_error("TimeSeries2Trace",
+                                                 message,
+                                                 ErrorSeverity.Invalid)
```
wangyinz, line 241: The `ErrorSeverity` should be `Complaint` instead of `Invalid`. Also, the algorithm name here is wrong. This is inside `resample` not `TimeSeries2Trace`. There are similar issues to the rest of the changes here.

python/mspasspy/algorithms/resample.py
```diff
@@ -229,6 +229,21 @@ def resample(self, mspass_object):
...
+                # Update sampling_rate to 1/dt
+                mspass_object["sampling_rate"] = 1.0 / mspass_object.dt
+            else:
```
wangyinz, line 244: The else branch here is not covered by test. Also, I think we need to add tests for the converter, too. We also need to test for the elog here. I am actually not sure whether elog is correctly retained after the conversion to Trace and back.

wangyinz (changes requested):
This is great! Thanks for fixing it! I only have two minor comments. We can merge it after you address it. Thank you!

wangyinz:
btw, I see the python builds failed due to `AttributeError: module 'pymongo' has no attribute 'database'`, which is something we currently have not fixed for the API change of the new pymongo. The commit https://github.com/mspass-team/mspass/commit/8ef0ae3e60882c82973e01c12523f89cb1910447 address the problem by temporarily limit our pymongo to an old version... We might need to look into implementing the changes now...

python/mspasspy/algorithms/resample.py
Cloud1e: I fixed it and have pushed it.

python/mspasspy/algorithms/resample.py
Cloud1e: I fixed it and pushed it. I think elog is correctly retained during my tests.

Cloud1e:
> btw, I see the python builds failed due to `AttributeError: module 'pymongo' has no attribute 'database'`, which is something we currently have not fixed for the API change of the new pymongo. The commit [8ef0ae3](https://github.com/mspass-team/mspass/commit/8ef0ae3e60882c82973e01c12523f89cb1910447) address the problem by temporarily limit our pymongo to an old version... We might need to look into implementing the changes now...

Sorry, I forgot to merge the master branch to my branch before pushing. Now I think it works fine.

Cloud1e:
> btw, I see the python builds failed due to `AttributeError: module 'pymongo' has no attribute 'database'`, which is something we currently have not fixed for the API change of the new pymongo. The commit [8ef0ae3](https://github.com/mspass-team/mspass/commit/8ef0ae3e60882c82973e01c12523f89cb1910447) address the problem by temporarily limit our pymongo to an old version... We might need to look into implementing the changes now...

I fixed this and now It works well.

PR #578, opened 2024-11-11
feat: make seisbench dependency optional 

JiaoMaWHU:
## What
To make the `seisbench` dependency optional so that the size of the standard installment can be reduced.

## Next steps:
Also need to update the container definition as part of this pr, so that dev container can have seisbench installed

## Testing
Ran the following commands and verified seisbench is not installed in the first one
```
python3 -m pip install .
python3 -m pip install '.[seisbench]'
```

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -49,7 +49,7 @@ jobs:
         echo "/opt/spark-3.0.0-bin-hadoop2.7/bin" >> $GITHUB_PATH
         python -m pip install --upgrade --upgrade-strategy eager pyspark pytest-spark
     - name: Install
-      run: python -m pip install -C--global-option=build -C--global-option=--debug -v .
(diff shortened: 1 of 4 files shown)


wangyinz:
I reran the test in this branch and it failed too. So, there is probably something changed from the remote data.

JiaoMaWHU:
> I reran the test in this branch and it failed too. So, there is probably something changed from the remote data.

Let me take a look. We can consider revert this commit and see if that would fix the issue.

JiaoMaWHU:
@wangyinz please take a look at https://github.com/mspass-team/mspass/pull/580

wangyinz:
Oh, I said already there is probably something wrong in seisbench's server. I don't think reverting this PR will do anything. It is going to fail regardless. I think we just need to remove the dependency of its remote ML model.

PR #581, opened 2024-11-16
Add Janitor class

pavlis:
This implements a new `Janitor` python class.  This can be used to solve a long standing problem with handling Metadata attributes in a workflow.  Attributes can become stale and/or inappropriate for the object containing them.  An example is "chan" values left as debris by the `bundle` function.   This class implements generic methods to handle cleaning out such debris.   

This initial push should be viewed as a draft from coimments by the group.  Colleagues please consider:
1.  Is the API complete?
2. Are the names rational?   They are kind of cutsey but that also makes them memorable. 
3. Read the docstrings and critically evaluate behavior of the methods.  

I think the tests may be broken on the very last thing run.   I'm pushing this anyway as I'm a bit suspect of my pytest setup and hoping I can get some insight running this on github.


pavlis:
If you can resolve the conflicts created by unresolved urls this branch inherited from master this should be ready to merge.

PR #601, opened 2025-01-23
update distributed_node.sh

Cloud1e:
update distributed_node.sh


wangyinz:
Oh, what I mean was to add this script to the pwmig's repo, and then change the existing script in this repo

wangyinz (approved):
Looks good to me! Thank you so much!

PR #615, opened 2025-02-08
For Tapis UI Execution

yiw008:
Dockerfile and script for executing the app on Tapis UI


.github/workflows/docker-publish.yml
```diff
@@ -130,6 +130,19 @@ jobs:
...
+        if: ${{ github.event_name == 'schedule' }}
+        with:
+          context: .
+          file: Dockerfile_tacc
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new
```
wangyinz, line 145: Adding a step here is not a good idea. It will make this job running for too long. Also, if the previous step fails, this image will not build. Anyway, what you should do is to add a new job. Also, the way you are specifying the tag here is problematic, we should tag it as "mspass:tacc" instead. 

I guess maybe you want to make sure this tacc image builds after the weekly image. But I think that should be done through job dependencies instead.

wangyinz (changes requested):
Thanks for the update! Please find my inline comment for changes requested.

PR #618, opened 2025-02-14
Enhance decorators - discussion #616

pavlis:
This implements a change to the mspass decorators to make it easier to adapt an algorithm that works only on atomic data to ensembles.  It basically automates adding a loop over ensemble members.   It implements ideas documented in discussion item #616.  

The change required small changes to most of the code in mspasspy.algorithms to add a kwarg switch to tell the decorator to enable the ensemble loop.

This initial push is virtually guaranteed to fail the pytest run here on github.   I did not run pytest on all the algorithms and the odds are high I made a mistake or one of the test make an assumption that is invalid with the decorators enabled.   Also there will be a gap in the tests to verify the ensemble functionality works for algorithms not previously working that way.  The coverage gizmo probably won't detect that since it is a decorator feature but that may be appropriate.

### DIFF
--- a/docs/source/user_manual/algorithms.rst
+++ b/docs/source/user_manual/algorithms.rst
@@ -5,7 +5,40 @@ Algorithms
 Overview
 -----------
 This page is a quick reference to algorithms available in MsPASS.
-The page is organized as a set of tables with hyperlinks to the docstrings
+The functions described here are all algorithms for handling MsPASS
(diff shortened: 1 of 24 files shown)


pavlis:
@wangyinz this is completed and passes all the tests including those I added for the changes to the decorator.  Note I also updated the algorithms page.   Unless you see a problem I think it can be merged to master.   It adds a fair amount of practical functionality when working with ensembles.

Issue #623, opened 2025-03-02
gil_scoped_release changes are problematic - current master is broken


pavlis:
Last week we merged in a branch that changed the pybind11 code for most C++ functions to use this kind of construct:
```
  m.def("_WindowData3C",py::overload_cast<const Seismogram&,const TimeWindow&>(&WindowData),
              "Reduce data to window inside original",
    py::call_guard<py::gil_scoped_release>(),
    py::return_value_policy::copy,
    py::arg("d"),
    py::arg("twin") )
  ;
```
I chose that particular example because it with version a processing script that uses `WindowData` (note `WindowData` is a thin python wrapper that invokes this C++ overloaded C++ function) will seg fault when the function is called.   I changed the binging code above to remove the gil_scoped_release line.  i.e.. 
```
  m.def("_WindowData3C",py::overload_cast<const Seismogram&,const TimeWindow&>(&WindowData),
              "Reduce data to window inside original",
    py::return_value_policy::copy,
    py::arg("d"),
    py::arg("twin") )
  ;
```
and the seg fault goes away.   More digging with spyder shows absolutely that it is that called to WindowData where the seg fault occurs.  

The code next fails on the following line which calls another C++ function modified with gil_scoped_release:
```
x=ExtractComponent(d,i)
```
After that call in a debugger I find:
```
IPdb [3]: print(d.t0,d.endtime(),d.npts)
-9.989461898803711 399.5605381011963 2201

IPdb [4]: print(x.t0,x.endtime(),x.npts,len(x.data))
-9.989461898803711 100.01053810119629 2201 2201
```
noting that dt=0.05 s so the values from `x` are correct but `d` the endtime method output for d is not correct.  

After much digging I think I understand the source of this problem.   You can get a good summary from the google AI with the search string:  "pybind11 gil_scoped_release_problem".  The AI mostly gets its information from [this page](https://pybind11.readthedocs.io/en/stable/advanced/misc.html).    My conclusion is that using gil_scoped_release on functions applied to our seismic data objects is a very bad idea because they all use the buffer protocol and link to python objects in Metadata.  The references warn that is dangerous and we got bit by it.  

The solution is clear.  I'm going to remove all calls to gil_scoped_release in the pybind11 code.   The gil sections we must retain are the ones in the pickle code, but they use a different mechanism anyway and the documentation doesn't suggest that approach is a problem.   We went down this road thinking these algorithms would work with python threading but we created a much worse problem.   Pretty much the entire code base is broken until we fix this.

PR #624, opened 2025-03-02
remove gil_scoped_release from all function binding code

pavlis:
This branch is a fix addressing issue #623.  See that issue page for why this is an essential fix - it appears the current master is completely broken until this is resolved.  

I didn't run the full test suite locally but did compile the C code successfully.   It is kind of surprising the pytest functions didn't show the problems that surfaced when I used the current branch but they did not.  For that reason I would be shocked if the tests here on github fail.


pavlis:
This branch should be merged asap.  As noted the entire code base seems to be broken until this is resolved.

PR #625, opened 2025-03-05
Bug fix for CNRDecon module

pavlis:
This is a fix isolated to the CNRDecon module.  It fixes an issue that caused seg faults due to a mistake in handling fft padding in the C++ code.  A  consequence was the output was then being returned of length defined by the fft size not the original data window size.   I elected to have the python code handle this by adding an option (default True) to window the output to the processing window size (signal_window parameter of the function input).


pavlis:
PS:   I forgot this also fixes a problem that certain extreme bad data (e.g. completely dead component with all 0 samples) could create undetected NaN or Inf sample values in the output.  I added a generic trap for that condition to handle it correctly with kill/elog approach of mspass

Issue #629, opened 2025-03-19
Bug in Database.read_data - miniseed read error handling


pavlis:
We have a minor bug in the read_data method of Database.  The following section (lines 4226 - 4230) needs an error handler:
```
           with open(fname, mode="rb") as fh:
                if foff > 0:
                    fh.seek(foff)
                flh = io.BytesIO(fh.read(nbytes))
                st = obspy.read(flh, format=format)
```
The obspy.read line threw the following exception:
```
raise InternalMSEEDParseTimeError(msg)
obspy.io.mseed.InternalMSEEDParseTimeError: julday out of bounds (wrong endian?): 16470
```
when it read the wrong file due to a bug in my job script.   The read section here that uses obspy.read needs an exception handler to do the standard MsPASS thing of kill and post the error message to elog.  We need to look at the obspy.read function and make sure all exceptions it can throw are handled.  

Note this bug is rare as I've read millions of miniseed files and only got this one due to a different mistake so it is not a pressing problem.

pavlis:
As best I can judge ObsPY uses a set of bare bones exception objects that are all subclasses of `Exception`.   I think `Exception` is a simple subclass of `BaseException` which is the base class for all python exception objects.   All the ObsPy exceptions seem to be little more than aliases for Exception.  Here is a typical definition you can find in their documentation:
```
class ObsPyException(Exception):
    pass
```
Not sure why they do that, but I am going to try fixing this bug by trapping the `Exception` object as it seems to be a base class for all ObsPy exceptions.

wangyinz:
Checked by Codex

Rechecked against current master: `obspy.read` in `_read_data_from_dfile` is now wrapped with `TypeError` and generic `Exception` handlers that kill the datum and log the read failure to `elog`. That appears to address the missing error handling reported here; leaving the issue open per triage policy.

wangyinz:
The missing error handling is now present.

`_read_data_from_dfile` wraps the formatted `obspy.read` call with both `TypeError` and generic `Exception` handlers. Either path kills the datum and records the underlying read failure in its error log; normal reconstruction runs only in the `else` branch after a successful read.

Closing as completed.

PR #630, opened 2025-03-20
normalize and database bug fixes

pavlis:
This branch contains a bug fix for two problems:
1.   I found the normalize function (in normalize module) did not work with ensembles because of a mistake I had made in application of the decorator extensions made recently.   
2. Issue #629 

In this initial push I haven't revised the test to check either of these fixes.  For 1 I'm not sure it is necessary as the only thing adding such a test would do was duplicate tests of the decorator and validate normalize does indeed work with ensembles correctly.  2 is more problematic as I am not sure how to do a test to check the error handlers I added.   The new handlers are there to handle corrupted or completely invalid data files.   I think just trying to read a junk file with content not matching the format would work, but I'm unsure how to do that.  I'm going to push this to see what happens and ask for help from the group.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -4226,128 +4226,158 @@ class Database(pymongo.database.Database):
             with open(fname, mode="rb") as fh:
                 if foff > 0:
                     fh.seek(foff)
-                flh = io.BytesIO(fh.read(nbytes))
-                st = obspy.read(flh, format=format)
-                if isinstance(mspass_object, TimeSeries):
-                    # st is a "stream" but it may contains multiple Trace objects gaps
-                    # but here we want only one TimeSeries, we merge these Trace objects and fill values for gaps
-                    # we post a complaint elog entry to the mspass_object if there are gaps in the stream
-                    if len(st) > 1:
-                        message = "WARNING:  gaps detected while reading file {} with format {} using obspy\n".format(
-                            fname, format
-                        )
-                        message += "Using specified fill defined in call to read_data"
-                        mspass_object.elog.log_error(
-                            "read_data",
-                            message,
-                            ErrorSeverity.Complaint,
-                        )
-                        st = st.merge(
-                            method=merge_method,
-                            fill_value=merge_fill_value,
-                            interpolation_samples=merge_interpolation_samples,
-                        )
-                    tr = st[0]
-                    # We can't use Trace2TimeSeries because we loose
-                    # all but miniseed metadata if we do that.
-                    # We do, however, need to compare post errors
-                    # if there is a mismatch
-                    if tr.stats.npts != mspass_object.npts:
-                        message = "Inconsistent number of data points (npts)\n"
-                        message += "Database npts={} but obspy reader created a vector {} points long\n".format(
-                            mspass_object.npts, tr.stats.npts
-                        )
-                        message += "Set to vector size defined by reader."
-                        mspass_object.elog.log_error(
-                            alg, message, ErrorSeverity.Complaint
-                        )
-                        mspass_object.set_npts(tr.stats.npts)
-                    if tr.stats.starttime.timestamp != mspass_object.t0:
-                        message = "Inconsistent starttimes detected\n"
-                        message += "Starttime in MongoDB document = {}\n".format(
-                            UTCDateTime(mspass_object.t0)
-                        )
-                        message += "Starttime returned by obspy reader = {}\n".format(
-                            tr.stats.starttime
-                        )
-                        message += "Set to time set by reader"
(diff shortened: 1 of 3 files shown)


wangyinz:
So, the problem is in the added `args` and `kwargs` in normalize. However, I am totally puzzled. I realized the code in your original change does not make sense at all, where you have: 
```
@mspass_func_wrapper
def normalize(
    mspass_object,
    matcher,
    kill_on_failure=True,
    *args,
    handles_ensembles=False,
    checks_arg0_type=False,
    handles_dead_data=True,
    **kwargs,
):
``` 
But, how can you place `args` after `kill_on_failure=True`? The `kill_on_failure` is a `kwargs` in this context. I then checked the other PR that was merged making these changes to all functions (https://github.com/mspass-team/mspass/pull/618/files), and I can see that issue is almost everywhere. How did they work at all? I suspect there is something implicitly assumed that I don't understand....

wangyinz:
Also, the pickle error comes back, but it is at a different place: https://github.com/mspass-team/mspass/actions/runs/13996038010/job/39191344594. Note that here is the old one: https://github.com/mspass-team/mspass/actions/runs/13709553232/job/38342891382
 
I am again puzzled. By looking at the pybind11 code, I see no serialization supported for the `FFTDeconOperator` class: https://github.com/mspass-team/mspass/blob/db2ab1b03ebd180f6ea8aa379479fb337019ba8c/cxx/python/algorithms/deconvolution_py.cc#L257-L264

The only class that supports serialization is it child `CNRDeconEngine`: https://github.com/mspass-team/mspass/blob/db2ab1b03ebd180f6ea8aa379479fb337019ba8c/cxx/python/algorithms/deconvolution_py.cc#L319-L378

How did it work at all?

pavlis:
1. For the *args and **kwargs point in the fix for normalize, I was just using multiple patterns for how we used that decorator before.   What I do know from other examples I worked with in debugging the revision to  the function and method decorators, that if you don't do it that way it won't work.   
2. Let me look at the issue with FFTDeconOperator.   Not serializing something required there may be the issue for that older problem.   The same class is a base class for all the algorithms supported through the RFDecon function.    They work so maybe those definitions had serialization code for FFTDeconOperator.    On the other hand, this is confusing because I am pretty sure I remember worrying about serialization of that class as it was a little weird.  More when I've reviewed the situation.

pavlis:
I looked at the issue with pickle of the deconvolution operator and realized why you would be confused.   The main thing to realize is that all the deconvolution operators, including the problematic CNRDeconEngine, use boost serialization at the C++ level and then build the interface to pickle in pybind11 with the pybind11 incantations with the boost serialization code.  Let me show a typical example.   Here is the private section of the LeastSquareDecon class:
```
private:
  int read_metadata(const mspass::utility::Metadata &md);
  int apply();
  double damp;
  friend boost::serialization::access;
  template <class Archive>
  void serialize(Archive &ar, const unsigned int version) {
    ar &boost::serialization::base_object<FFTDeconOperator>(*this);
    ar &boost::serialization::base_object<ScalarDecon>(*this);
    ar & damp;
  }
```
and here is the way pickle is set up for LeastSquareDecon in the pybind11 code:
```
.def(py::pickle(
      [](const LeastSquareDecon &self) {
        pybind11::gil_scoped_acquire acquire;
        stringstream sstm;
        boost::archive::text_oarchive artm(sstm);
        artm<<self;
        pybind11::tuple r_tuple = py::make_tuple(sstm.str());
        pybind11::gil_scoped_release release;
        return r_tuple;
      },
      [](py::tuple t) {
        pybind11::gil_scoped_acquire acquire;
        stringstream sstm(t[0].cast<std::string>());
        boost::archive::text_iarchive artm(sstm);
        LeastSquareDecon lsd;
        artm >> lsd;
        ShapingWavelet sw=lsd.get_shaping_wavelet();
        ComplexArray w(*sw.wavelet());
        pybind11::gil_scoped_release release;
        return lsd;
      }
    ))
```
What that shows is the operator is serialized with boosts serialization using the ` boost::archive::text_iarchive` class.   The include file shows that be base classes are serialized using the boost api.   

The C++ part is certified to work with the C++ test program  in cxx/test/decon_serialization.   What is different about CRNDeconEngine to cause it to fail remains a mystery to me.   

Maybe another set of eyes looking for dumb mistakes in the pybind11 code is a possible next step to address that.  For now it looks like you fixed my mistake with the nomalize function and the only residuals is this pickle error coming back.

pavlis:
One additional point about the pickle error with CNRDeconEngine.   It is not a random problem as when I tried running CNRRFDecon on 2013 extended usarray data as a parallel job (with map operators) it failed in some cloudpickle function.

PR #632, opened 2025-03-31
fix error in use of *args and **kwargs

pavlis:
Fixes a mistake I made in an earlier commit in use of *args in functions using the mspass_func_decorator.


pavlis:
@wangyinz it looks like my previous approach to handling *args and **kwargs was correct and the more commonly stated idea of how to handle *args and **kwargs is not appropriate for the way we define these decorators.   To be clear, normal suggested use for placement of *args and **kwargs is this kind of setup (made up example):
```
def normal_usage(a,b,*args,x=“foo”,y=“bar”,**kwargs):
```
With the mspass_func_decorator or mspass_method_decorator this proper usage is like this example from our current code base:
```
@mspass_func_wrapper
def snr(
    data_object,
    noise_window=TimeWindow(-130.0, -5.0),
    signal_window=TimeWindow(-5.0, 120.0),
    noise_metric="mad",
    signal_metric="mad",
    perc=95.0,
    *args,
    handles_ensembles=False,
    checks_arg0_type=False,
    handles_dead_data=False,
    **kwargs,
):
```
In this branch I changed this to the following which is like the made up example below:
```
@mspass_func_wrapper
def snr(
    data_object,
    *args,
    noise_window=TimeWindow(-130.0, -5.0),
    signal_window=TimeWindow(-5.0, 120.0),
    noise_metric="mad",
    signal_metric="mad",
    perc=95.0,
    handles_ensembles=False,
    checks_arg0_type=False,
    handles_dead_data=False,
    **kwargs,
):
```
which seems to be incorrect.  I think the essence of the problem is that the decorator causes the kwargs between the change to be dropped.  We should do a few additional tests to verify that suggestion is true, but it rings true with my memory from when I revised the mspass_func_decorator and mspass_method_decorator.   

In any case, i think this is a dead branch but one we should retain for this record.   I have no idea how to do that, but as the github keeper I’m hoping you know how to do that.  We definitely DO NOT want to merge this branch into master.

wangyinz:
I found that the fix here are actually needed. The use of arg and kwarg here is correct. The previous error is actually due to incorrectly calling the target functions, and they didn't fail previously because of the wrong placement of arg and kwarg. So, basically, this helps us identified bugs. I am merging it now. @pavlis

PR #633, opened 2025-04-01
Fix ApplyCalib constructor bug

pavlis:
This fixes a rare bug for bad channel Metadata.  Hit an example where the other calibration data seeded valid but the sensitivity value was set to 0.0.   That caused a divide by 0 error exception when sensitivity is converted to calib (calib = 1/sensitivity).   This adds a small error handler that should take care of the problem.  Probably should have  pytest for this exception, but this is an rare error and indicates something got by IRIS’s QC procedures to have an invalid entry like that.


pavlis:
Note for the record to put on the development agenda.   The constructor for the ApplyCalib class should print more information in verbose mode when it finds a problematic response object.   Right now it just prints a message describing the problem but it really should print some or all of the problematic document it parses.   The example that drove this fix is a case in point.  Somewhere in about 40,000 document is the problem child but I can’t find it without deserializing the obspy Response structure for each document.

pavlis:
This seems to be failing only because of the codecov issue I predicted and the same error @JiaoMaWHU reported in our meeting yesterday.   I would urge you to go ahead and merge this bug fix and sort out what is happening with the test that is failing.

JiaoMaWHU:
> This seems to be failing only because of the codecov issue I predicted and the same error @JiaoMaWHU reported in our meeting yesterday. I would urge you to go ahead and merge this bug fix and sort out what is happening with the test that is failing.

ack. I tried to fix the test using an old version of the boto library, which didn't work. I will continue trying to fix this test.

PR #640, opened 2025-04-18
Adds IU launchers including prototype python mspass launcher classes

pavlis:
This branch changes nothing in the mspass code base.  It adds a new directory to the scripts directory called "IU_examples".   That base directory contains shell scripts used on the IU cluster to launch mspass.   They are variants of tacc versions.   They are useful documentation for users to see or how to set up a local cluster configuration with the unix shell. 

The second part of this commit is more important.  That is found in the "python" directory of IU_examples.   The file called `launcher.py` contains two prototype classes to replace shell launchers with python launchers:
1.  `HPCClusterLauncher` is designed to replace the awkward shells used at present to launch mspass on HPC clusters.   It uses a configuration file that eliminates the error prone problem of editing a shell script to get a functional job script.   It also provide significant additional functionality with methods to monitor status of the cluster and a clean shutdown mechanism. 
2. `DesktopLauncher` is essentially a python class that automates the docker-compose sequence desribed in our user manual on [this page](https://www.mspass.org/getting_started/deploy_mspass_with_docker_compose.html).  Like HPCClusterLauncher the class constructor is driven by a configuration file.  

The commit also has sample configuration files and test launch python scripts.   Note this these are best for now as launcher.py does not match our current pytest setup as it would normally be run outside the container.

This pull request should pass all existing tests since it makes not changes to the code base.  It should not, however, be merged until our next group meeting where I will demonstrate its use and current limitations.   A key design point is how to utilize these prototypes as the basis for a cleaner front end for MsPASS.


pavlis:
Given the discussion in the meeting this morning  and the fact that the python launcher.py is stated as a pure prototype I think you should just merge this for the record.   A comment I just added to a discussion page assumes that.

PR #684, opened 2026-02-04
Add sliding window

pavlis:
This branch implements the sliding window algorithm with Futures whose development is described in Discussion #656.  I've also included a complete pytest suite for the function to verify all features work - at least on simple data.   

mspass-team members please evaluate the api critically.  Also because this is a generic function are there additional sanity checks on the arguments that need to be added?

Hopefully this will pass tests clean, but it is possible the use of a dask.distributed.LocalCluster may behave differently on github than on the machine I ran pytest.


pavlis:
I fixed the problems that were causing the tests to break.   In our meeting Friday all agreed otherwise this was ready to merge so I'll proceed.

PR #687, opened 2026-02-11
Chenbo/cnrdecon dask serialization

### LINKED ISSUE #680: CNRRFDecon Parallel Processing: Serialization Error with CNRDeconEngine
## Current Status

Attempting to reproduce the serialization issue mentioned in previous discussions regarding parallel execution of `CNRRFDecon_2012.py` using Dask futures.

## Problem Encountered

The script fails immediately (~30 seconds) with a serialization error when trying to submit tasks to Dask workers:

```
TypeError: cannot pickle 'PyCapsule' object
```
```
[ERROR] TypeError: ('Could not serialize object of type LLGExpr', "LLGExpr(dsk={'process-6d2a9153537296ebef741c845dde190b': <Task 'process-6d2a9153537296ebef741c845dde190b' process(..., ...)>})")
```

## What Was Modified

Converted the serial processing loop to parallel processing using Dask futures:
- Added `dask_client = mspass_client.get_scheduler()`
- Replaced `for sid in sidlist:` loop with `dask_client.submit()` and `as_completed()`
- Added test mode with data tag management for idempotent testing

The script fails before any meaningful CPU usage can be observed due to the immediate serialization error.

## Questions

1. **@wangyinz @pavlis**: Have I successfully reproduced the serialization issue you mentioned? The error occurs when trying to pass `engine`, `db`, and `dbmd` objects to worker processes.

2. **Solution Approach**: Since `CNRDeconEngine` is similar to `db_handle` (both contain non-serializable C extension objects), should we use a Dask serialization plugin similar to the PR #674 approach for `db_handle`?

3. **Configuration**: Does the `workers_per_node: 144` and `primary_node_workers: 10` configuration look correct for Vista?

Cloud1e:
# Fix amplitude functions pickle serialization for Dask

Fixes #680

## Problem

When using amplitude functions (`MADAmplitude`, `PeakAmplitude`, `RMSAmplitude`, `PercAmplitude`) in Dask parallel workflows, cloudpickle fails to serialize them when captured in closures, throwing `TypeError: cannot pickle 'PyCapsule' object`. This happens because pybind11 function bindings are exposed as PyCapsule objects, which cannot be pickled.

## Solution (amplitude functions)

Following @pavlis Gary's suggestion (based on Gemini's recommendation), we wrap each amplitude function in a stateless functor class and implement `py::pickle` support. Cloudpickle can then serialize the functor by class name and empty state, and reconstruct it on worker nodes.

**Implementation details:**
- Add C++ functor structs (`MADAmplitudeFunctor`, `PeakAmplitudeFunctor`, `RMSAmplitudeFunctor`, `PercAmplitudeFunctor`) with `operator()` overloads for `CoreTimeSeries` and `CoreSeismogram`; each delegates to the existing C++ function. This gives us classes that pybind11 can bind instead of plain functions that become PyCapsules.
- Replace `m.def("XAmplitude", ...)` with `py::class_<XAmplitudeFunctor>` bindings so Python gets class instances, not raw PyCapsules.
- Implement `py::pickle` with empty state (`__getstate__` returns `()`, `__setstate__` does nothing) since the functors hold no state.
- Register `__call__` overloads so the instances are callable like functions.
- Set `m.attr("XAmplitude") = XAmplitudeFunctor()` so existing `MADAmplitude(data)`, `PeakAmplitude(data)`, etc. usage continues to work unchanged.

## Other fixes

**setuptools pin:** CI was failing with `ModuleNotFoundError: No module named 'pkg_resources'`. ObsPy depends on `pkg_resources`, which was removed from setuptools 70+ ([obspy/obspy#3681](https://github.com/obspy/obspy/issues/3681)). Pin `setuptools>=58,<70` in `requirements.txt` and `meta.yaml`. Can be reverted once ObsPy ships a release that no longer depends on pkg_resources.

**mongodb-github-action:** Upgraded `supercharge/mongodb-github-action` from 1.3.0 to 1.12.1 to fix Docker API 1.40/1.44 mismatch in CI.

## Testing

- Added `python/tests/algorithms/test_amplitudes_pickle.py` with cloudpickle round-trip tests for all four amplitude functions.
- Tested locally with Dask/cloudpickle serialization.
- Tested on TACC Vista with parallel Dask workflows.
- Verified API compatibility: `MADAmplitude(data)`, `PeakAmplitude(data)`, etc. work unchanged.
- All existing tests pass.

## API Compatibility

The Python API remains unchanged. Users can still call `MADAmplitude(data)`, `PeakAmplitude(data)`, `RMSAmplitude(data)`, `PercAmplitude(data, perf)` exactly as before, but now they work seamlessly with Dask serialization.

### DIFF
--- a/cxx/python/algorithms/amplitudes_py.cc
+++ b/cxx/python/algorithms/amplitudes_py.cc
@@ -22,27 +22,73 @@ struct MADAmplitudeFunctor {
   }
 };
 
+struct PeakAmplitudeFunctor {
+  double operator()(const CoreTimeSeries& d) const {
+    return PeakAmplitude(d);
+  }
+  double operator()(const CoreSeismogram& d) const {
+    return PeakAmplitude(d);
+  }
+};
+
+struct RMSAmplitudeFunctor {
+  double operator()(const CoreTimeSeries& d) const {
+    return RMSAmplitude(d);
+  }
+  double operator()(const CoreSeismogram& d) const {
+    return RMSAmplitude(d);
+  }
+};
+
+struct PercAmplitudeFunctor {
+  double operator()(const CoreTimeSeries& d, double perf) const {
+    return PercAmplitude(d, perf);
+  }
+  double operator()(const CoreSeismogram& d, double perf) const {
+    return PercAmplitude(d, perf);
+  }
+};
+
 PYBIND11_MODULE(amplitudes, m) {
   m.attr("__name__") = "mspasspy.ccore.algorithms.amplitudes";
   m.doc() = "A submodule for amplitudes namespace of ccore.algorithms";
 
-  /* Amplitude functions - overloads */
-  m.def("PeakAmplitude",py::overload_cast<const CoreTimeSeries&>(&PeakAmplitude),
-    "Compute amplitude as largest absolute amplitude",
-    py::return_value_policy::copy,py::arg("d") )
-  ;
-  m.def("PeakAmplitude",py::overload_cast<const CoreSeismogram&>(&PeakAmplitude),
-    "Compute amplitude as largest vector amplitude",
-    py::return_value_policy::copy,py::arg("d") )
-  ;
-  m.def("RMSAmplitude",py::overload_cast<const CoreTimeSeries&>(&RMSAmplitude),
-    "Compute amplitude from rms of signal",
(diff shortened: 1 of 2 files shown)


wangyinz (approved):
Looks good to me. Thank you!

PR #711, opened 2026-04-16
Make MongoDBWorker pickle-safe for Dask register_plugin

Cloud1e:
## What

`MongoDBWorker` now implements `__getstate__` / `__setstate__` so only `dbclient_key` and `connection_url` get pickled. The real `DBClient` is still created in `setup()` on each worker.

## Why

Registering the plugin with Dask can pickle the `WorkerPlugin`. The default path sometimes pulls in things that don’t pickle cleanly (I’ve hit asyncio-related errors). This keeps registration reliable without changing how workers get a DB handle.

### DIFF
--- a/python/mspasspy/util/db_utils.py
+++ b/python/mspasspy/util/db_utils.py
@@ -82,7 +82,10 @@ class MongoDBWorker(WorkerPlugin):
         self.connection_url = url
 
     def __getstate__(self):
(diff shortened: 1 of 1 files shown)


wangyinz (approved):
Thanks for fixing the bug!

Just a note here. The fact that the code change is covered by the test already suggests that the dask plugin does implicitly serialize the object somehow. We didn't know this but the error we ran into also suggested this.

PR #715, opened 2026-04-22
Add Exception fallback to mspass decorator wrappers

Cloud1e:
## Summary
Hardens MsPASS decorator error handling so parallel/map-style callers always get a **MsPASS data object back** (with errors in `elog` and the datum marked **dead** where appropriate), instead of Python’s implicit **`None`**—which was causing downstream `AttributeError` (e.g. on `.live`).
### `mspass_func_wrapper` / `mspass_method_wrapper`
- **`RuntimeError`**, **`MsPASSError`** (non-`Fatal`), and **`Exception`** (when `data` is a known wrapper type): after logging to `elog` / `ensemble_error`, call **`data.kill()`** and **`return data`**.
- **`MsPASSError`** with **`Fatal`** severity: still **re-raised** (no kill/return on that path).
- **`Exception`** when `data` is **not** a supported MsPASS type: **re-raise** unchanged.
### `mspass_func_wrapper_multi`
- Same pattern for both inputs: log per atomic/ensemble branch, then **`data1.kill()`**, **`data2.kill()`**, **`return data1`** for **`RuntimeError`**, non-`Fatal` **`MsPASSError`**, and handled **`Exception`**.
- **`Fatal` `MsPASSError`**: still **re-raised**.
### `mspass_reduce_func_wrapper`
- Keeps the existing **`MsPASSError`** rule: only **`Informational`** / **`Debug`** are logged in the wrapper; other severities **re-raise**.
- For **`RuntimeError`**, logged **`MsPASSError`**, and handled **`Exception`**: after logging, **`data1.kill()`**, **`data2.kill()`**, **`return data1`**.
### Tests (`python/tests/util/test_decorators.py`)
- New **`dummy_*`** helpers and assertions to cover the above branches (including dual-arg, ensemble companion, `Fatal` vs non-`Fatal`, and reduce **`Exception`**).
- Existing ensemble **`check_arg0_tester`** expectations updated where the decorator now **logs** instead of propagating **`TypeError`**.

### DIFF
--- a/python/mspasspy/util/decorators.py
+++ b/python/mspasspy/util/decorators.py
@@ -155,7 +155,8 @@ def mspass_func_wrapper(
     :param kwargs: extra kv arguments
     :return: modified seismic data object.  Can be a different type than
        the input.   If function_return_key is used only the Metadata
-       container  should be altered.
+       container  should be altered.  On RuntimeError, MsPASSError (non-Fatal),
+       or generic Exception, arg0 is logged, killed, and returned.
     """
     # Add an error trap for arg0 if this function doesn't do that
     if not checks_arg0_type:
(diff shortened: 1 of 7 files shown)


wangyinz (approved):
Thank you for fixing this!

PR #719, opened 2026-05-21
Fix issue 710

### LINKED ISSUE #710: Bug in Stream2TimeSeriesEnsemble and Stream2SeismogramEnsemble
I am pretty sure there is a serious bug in Stream2TimeSeriesEnsemble and by association Stream2SeismogramEnsemble.    Neither sets the ensemble live attribute of the return.  As a result the output is always marked dead.   I encountered this with this section of code while experimenting with the new s3 reader using the earthscope sdk.   
```
            strm.sort()
            strm.merge()
            alldata = Stream2TimeSeriesEnsemble(strm)
            del strm
            print("converted to ensemble - number members=",len(alldata.member))
            print("ensemble live state=",alldata.live)
            for d in alldata.member:
                if "loc" in d:
                    loccode = d["loc"]
                else:
                    loccode = "NULL"
                print(d["chan"],loccode,d.t0,d.npts,d.live)
```
With some test data it yielded this output:
```
loaded stream of size= 16
stream of size after select= 3
converted to ensemble - number members= 3
ensemble live state= False
BHE 00 1263081600.0318 1727995 True
BHN 00 1263081600.0318 1727995 True
BHZ 00 1263081600.0318 1727995 True
```
That shows the converted ensemble had live members but was itself marked dead.   That is not the right behavior.  

There is a simple workaround for now, but we need to fix this asap.  Here is the workaround.  It requires the function with this import:
```
from mspasspy.util.seismic import number_live
```
Then right after the converter you run:
```
n=number_live(alldata)
if n>0:
    alldata.set_live()
```

pavlis:
As the name suggests this addresses issue 710.   The commit includes updates to the tests to validate changes.    As soon as the github tests pass this branch should be good to merge.

### DIFF
--- a/python/mspasspy/util/converter.py
+++ b/python/mspasspy/util/converter.py
@@ -756,6 +756,8 @@ def _converter_get_ensemble_keys(ens):
                 return d["CONVERTER_ENSEMBLE_KEYS"]
             else:
                 return list()
+    # need this to handle ensembles with all dead members
+    return list()
 
 
 def Stream2TimeSeriesEnsemble(stream):
--- a/python/mspasspy/util/converter.py
+++ b/python/mspasspy/util/converter.py
@@ -25,6 +25,7 @@ from mspasspy.ccore.seismic import (
     Keywords,
 )
 from mspasspy.ccore.algorithms.basic import _ExtractComponent
+from mspasspy.util.seismic import has_live_data
 
 
 def dict2Metadata(dic):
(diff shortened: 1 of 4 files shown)


pavlis:
That failure after fixing the typo was unexpected.   Digging suggests it is related to a more fundamental problem with how the converters handle dead data in ensembles.   I thought this would be a quick fix, but I am addressing that problem.   Team members I will give a report on this potential problem in our next weekly meeting.

pavlis:
I was overly pessimistic.   The existing code did not have a major problem.  The bug that caused the failures in the early push were an error in an internal function that didn't handled a ensemble with no live data.

PR #736, opened 2026-06-13
Add GID lag penalty QC and validation

wangyinz:
## Summary
- Add GID lag-weight penalty support for `none`, `boxcar`, `cosine_taper`, `shaping_wavelet`, `resolution_kernel`, and the new adaptive `adaptive_memory` mode.
- Implement adaptive penalty memory for both `TimeDomainGIDDecon` and `FrequencyDomainGIDDecon`, using resolution-kernel footprints, search-adjusted noise confidence, finite retention, and consistent iteration-cap behavior.
- Preserve decon QC metadata for downstream review, including configured/effective penalty widths, final lag-weight norms, adaptive confidence/retention fields, noise amplitude, valid-lag counts, NS-GID noise RMS, and compact resolution-kernel metadata.
- Update default GID parameter files to use `adaptive_memory` with scale `0.35`; keep `least_square` as the compatibility inverse default while documenting `adaptive_memory` + `ns_gid` as the noisy-production recommendation.
- Expand validation plots/tests for penalty comparisons, high-noise manual plot runs, complex colored-noise setup figures, and non-GID QC metadata preservation.

## Theory And Documentation
- Document the basic lag-penalty formulation and the full adaptive anti-cycling framework.
- Clarify that `adaptive_memory` is a guarded anti-cycling prior, not a guarantee of dominance over `none` for dense close-arrival support.
- Add benchmark-backed guidance: adaptive behavior matches/collapses toward `none` at extreme noise scales when selected peaks are not significant relative to the full lag search, while `none` can remain preferable for tightly clustered true arrivals.
- Explain that `shaping_wavelet`, `resolution_kernel`, and `adaptive_memory` derive effective support from kernels; `lag_weight_function_width` is preserved as fixed-width/QC context in those modes.

## Validation
- `conda run -n mspass python -m pytest python/tests/algorithms/test_TimeDomainGIDDecon.py python/tests/algorithms/test_FrequencyDomainGIDDecon.py -q`
- `conda run -n mspass python -m pytest python/tests/algorithms/test_decon_algorithm_validation.py -q`
- `conda run -n mspass python -m pytest python/tests/algorithms/test_decon_algorithm_validation.py -q --decon-validation-plots --decon-validation-plot-dir /tmp/mspass-review-decon-plots --decon-validation-noise-scale 100`
- `git diff --check`
- `git grep -n -E "group_sparse|GROUP_SPARSE|GroupSparse|group-sparse|group sparse" -- cxx data docs python/tests` returned no matches.

## Review Pass Notes
- Used sub-agents to review C++ behavior/QC, tests/plots, and docs/config consistency.
- Fixed review findings by adding `.pf` comments, clarifying the shipped default vs noisy-production recommendation, strengthening kernel-footprint tests, expanding adaptive high-noise assertions through noise scale `100`, and annotating normalized manual setup plots with measured RMS values.
- Latest pushed commit: `4010a61e35708cea4cf23d2223f619f4903d791b`.


pavlis:
This is really interesting you could implement that that fast.   Another application the AI?

wangyinz:
Yep... BTW, one thing that struck me while watching codex doing it is that I think we want to introduce some decaying factor to the penalty, such that a spike or its close neighbors can still be modified after certain number of iterations. What do you think?

pavlis:
Yes, that makes sense but a simpler way to do that is to make the penalty function less aggressive.   I don’t remember the exact formula and I’m too lazy to look it up but the penalty function was ad hoc.   I know it is pretty aggressive with the defaults.  We may need to experiment those parameters to set the defaults better.   The tests really, however, need to be done on real data not simulations, and that is harder to do.

Also give some thought to alternative penalty functions.  That was just the first thing I thought of.   I was mostly considering the penalty function concept, not the detail of what the form of the penalty function was.

Issue #757, opened 2026-07-22
RFdeconProcessor.actual_output() / output_shaping_wavelet() return _CoreTimeSeries instead of TimeSeries


Cloud1e:
After running `RFdecon(..., alg="TimeDomainGID", ...)`, calling `gid_engine.actual_output()` or `gid_engine.output_shaping_wavelet()` returns a low-level `_CoreTimeSeries` object instead of a full `TimeSeries`. Passing that directly into `WindowData()` (and presumably other higher-level API functions) fails:

```python
ao = gid_engine.actual_output()
print(type(ao))   # <class 'mspasspy.ccore.seismic._CoreTimeSeries'>

WindowData(ao, stime, etime)
```

```
TypeError: Illegal type for arg0=<class 'mspasspy.ccore.seismic._CoreTimeSeries'>.  Must be a MsPASS seismic data object
```

Wrapping manually works around it:

```python
ao = TimeSeries(gid_engine.actual_output())
```

but callers shouldn't need to know to do this. `actual_output()` / `output_shaping_wavelet()` should probably just return a full `TimeSeries` directly, consistent with the rest of the atomic data object API.

wangyinz:
Fixed in #766: `RFdeconProcessor.actual_output()` and `output_shaping_wavelet()` now return metadata-aware `TimeSeries` objects for both scalar and 3-C engines, allowing downstream windowing APIs to consume them. Closing as completed.

PR #759, opened 2026-07-30
Fix GID receiver-function timing, stopping, and diagnostics

wangyinz:
## Final update: target-specific MTZ presets and external-wavelet timing

Final commits: `e8bf75d8` (MTZ presets), `be9b30c9` (CNR/FFT hardening), `d1e4cf45` (external-wavelet timing and validation), and `7170bd48` (Sphinx-clean API documentation).

The public no-argument scalar workflow resolves the packaged `mtz_plane_wave_lowfreq` reference preset:

```text
analysis/output window   (-10, 120) s
source-wavelet window    (-5, 20) s
target dt                0.05 s
output shaping           zero-phase Ricker, 0.125 Hz
WaterLevel               0.1
frequency-domain LS      damping 1.0
```

This is explicitly documented as a **reference preset for MTZ plane-wave/Kirchhoff imaging**, not a universal receiver-function default. The 0.125-Hz Ricker is an objective-specific low-frequency reference for P410s/P660s imaging. Companion presets remain `mtz_plane_wave_highres` (0.25-Hz Ricker), `crustal_plane_wave` (0.5-Hz Ricker), and `raw_decon_diagnostic` (no shaping). Numerical PF values were not changed by the wording correction, and the source/package PF mirrors are identical.

The final API fixes also:

- correct `RFdeconProcessor.loaddata(TimeSeries, window=False)` to read `d.data`, with a direct regression test;
- add optional `wavelet_t0` to scalar, TD-GID, and FD-GID Python paths so a bare numpy/vector wavelet can be physically aligned to the analysis grid;
- keep a `TimeSeries` wavelet's own `t0` and `dt` authoritative;
- preserve legacy vector behavior when `wavelet_t0=None`, with one explicit `UserWarning`;
- validate nonfinite/invalid wavelets and datum/grid compatibility before mutating a reusable engine;
- preserve GID's P-relative lag-domain contract and reject unconverted epoch-time data with guidance to use `ator(P-arrival epoch)`.

The CNR fix removes the prior external-wavelet phase error and hardens datum/wavelet/PSD validation, zero/noise spectral limits, diagnostic timebases, transactional process state, copy/assignment/serialization semantics, and checked RAII ownership of GSL FFT resources. Scalar and 3C noise-spectrum producers reject nonfinite/overflowing inputs before mutating cached engines.

Final local verification:

```text
Python decon/RF/GID related suites           625/625 PASS
RFdeconProcessor full file                   113/113 PASS
Debug CTest                                    11/11 PASS
Release -O3 -DNDEBUG focused CTest               1/1 PASS
Valgrind                      74,261 allocs/frees, 0 leaks, 0 errors
git diff --check                                        PASS
```

Final MTZ waveform verification (0.05 s sampling, -10–120 s analysis, 0.125-Hz source) recovered injected phases at exactly `0.0/45.0/72.0 s`. Internal `TimeSeries` and external vector+`wavelet_t0` RFs are pointwise identical for scalar, TD-GID, and FD-GID paths; no sparse zero blocks or isolated late artifacts were introduced.

The unchanged final revision received two consecutive independent GPT-5.6 Sol High critical code-review rounds with no P0–P2 findings. A separate Sol High receiver-function seismologist review reported `FINAL WAVEFORM CLEAN 2`. All implementation changes were made by GPT-5.6 Terra High.

Remote status for `7170bd48`: Python 3.10/3.13, CMake, sdist, dependency alignment, Docker amd64/arm64 dev/latest, MPI, Codecov patch, and Codecov project all **PASS**.

## Summary

Fix generalized iterative receiver-function deconvolution in `TimeDomainGIDDecon` and `FrequencyDomainGIDDecon` for long mantle-transition-zone windows, physical wavelet/data time alignment, statistically valid convergence, complete QC, and long-noise FFT sizing.

Also fixes **#757** by changing the C++ scalar-deconvolution diagnostic API itself to return `TimeSeries`.

## Confirmed root causes and fixes

1. **Independent time origins were lost.**  The analysis record and short source wavelet were reduced to raw vectors before scalar inverse preprocessing.  With analysis `t0=-10 s` and wavelet `t0=-5 s`, their first samples were treated as simultaneous, so RF lag zero moved to about `+5 s`.  Both GID engines now construct a physical union grid, align/zero-pad by actual sample time, process every supported leaf consistently, and crop back to the unchanged analysis grid.  This applies to internal/external wavelets and noise-threshold preprocessing; it does not merely relabel `t0`.

2. **Residual-noise stopping used record-length-dependent L2 norms.**

   ```text
   old_legacy_ratio = ||residual||2 / ||noise||2

   residual_rms = ||residual||2 / sqrt(number_of residual time samples)
   noise_rms    = ||noise||2 / sqrt(number_of noise time samples)
   stopping_ratio = residual_rms / noise_rms
   ```

   The RMS ratio now drives stopping.  The old L2 ratio is retained as explicitly legacy QC.  Canonical and compatibility QC aliases are emitted together, including `ns_gid_residual_noise_rms_ratio`, `ns_gid_residual_rms_ratio`, `ns_gid_residual_rms`, and `ns_gid_noise_rms`.

3. **The sigma threshold had ambiguous statistics.**  `ns_gid_peak_sigma_threshold` now multiplies the RMS aggregate of three signed-component, median-centered MAD sigma estimates.  Zero/roundoff MAD falls back to the ordinary signed-component RMS aggregate; invalid or nonpositive scales fail explicitly.  The empirical 3C vector-amplitude quantile remains a separate maximum guard.  The same policy drives automatic group-sparse lambda.  Exact and compatibility QC names expose empirical, sigma, selected, robust, fallback, and per-component scales.

4. **Fractional convergence was evaluated after periodic ridge refitting.**  Every fifth accepted spike could therefore stop on the refit's state adjustment instead of the accepted candidate's improvement.  NS-GID now evaluates equation-15 fractional improvement from the candidate's pre-refit raw-L2 reduction.  Refit remains the next iteration state and has separate audit fields.

5. **Long noise could exceed the inverse-operator FFT allocation.**  Runtime sizing now includes the complete, untruncated noise record and grows the operator safely.

6. **Issue #757 returned reduced diagnostics.**  `ScalarDecon::actual_output()` and `output_shaping_wavelet()` now return `TimeSeries`; every override and pybind binding matches.  Python wrappers no longer construct temporary `TimeSeries(...)` objects.  Data, metadata, timebase, and live/dead state are preserved, and the direct engine objects pass directly to `WindowData`.

7. **`candidate_not_significant` could terminate on the wrong lag.**  The old loop maximized the lag-penalized score first and only then tested raw significance.  A strongly downweighted but raw-significant peak could therefore be hidden by an insignificant peak with a larger penalized score.  Both engines now first form the set `raw_amplitude >= threshold && lag_weight > 0`, then maximize the penalized score inside that set.  Final ridge refitting is followed by a fresh valid-lag scan; a reintroduced significant residual is reported as `post_refit_significant_candidate_remaining` with `converged=false`.  Invalid/zero analysis data and nonfinite internal/external noise or spectra fail with state-safe engine reuse.

8. **The fractional-improvement floor could be bypassed.**  Legacy leaf modes had been unintentionally changed from their historical post-acceptance Eq. (15) state test into a candidate-level gate.  NS-GID could reject its strongest decreasing but sub-floor candidate and then accept a weaker lower-ranked lag, producing false late support.  Legacy modes now accept the best decreasing candidate and apply their historical post-state test; NS-GID terminates on the first decreasing noise-significant sub-floor decision.  The final-refit audit preserves that ordered decision while scanning all remaining candidates read-only for global diagnostics.  Separate QC distinguishes compatibility pretrial scans, actual post-state tests, the decision candidate, and the globally best trial.

## Defaults and statistical interpretation

The PR ships the following conservative NS-GID profile:

```text
maximum_iterations                    100
residual_fractional_improvement_floor 1e-4
ns_gid_refit_interval                 5
ns_gid_residual_noise_ratio_floor     1.0
ns_gid_peak_sigma_threshold           3.0
ns_gid_peak_probability_threshold     0.995
ns_gid_use_empirical_noise_threshold  true
ns_gid_max_spikes                     0
ns_gid_ridge_beta                     1e-10
```

The multiplier `3.0` now means three robust scalar-component sigmas (with the documented fallback), not three vector-amplitude RMS units.  The operational peak threshold is still `max(empirical 3C amplitude quantile, scalar-sigma threshold)`, so it can exceed three scalar sigmas.  All other convergence defaults remain unchanged.

The exact current engine was screened on 332 real RF records from 10 station folds (31–39 records/fold), with analysis `(-10,160)`, wavelet `(-5,20)`, noise `(-200,-10)`, and `dt=.05 s`, for both engines on Vista ARM64.  Parameter sweeps reused these 332 records and are not thousands of independent observations.

- On the final commit `c608aeb4`, the exact shipped `q=.995, sigma=3, cap=100` profile has median iterations/spikes `25/26` (TD/FD), with q10 `3/3` and q90 `100/100`.  Median residual-energy fractions remain high at `.763/.762`.  Stops are TD `candidate_not_significant=260, max_iterations=57, residual_noise_floor=12, fractional_floor=3`; FD `259/57/12/4`.
- Lower `q=.8` and `q=.5` sweeps allow substantially more support and lower residual energy, but the added late energy is not station-coherent; `q=.5` often reaches the RMS noise floor before its probability threshold differs operationally from `.8`.  No tested lower probability threshold outperformed the independent scalar leaf references in both waveform repetition and H–κ behavior.
- Increasing the NS cap to 400 does not repair the default: the added support is dominated by event-specific/negative-time/boundary energy.  The paper-like `damping=1e7, Gaussian=0.2 s, epsilon=.05, Nmax=400` package reached median residual energy `.013` but had crustal LOO `.0008`, MTZ LOO `.0087`, H–κ stability `.16`, and 5/10 station peaks at or near the H-grid boundary.  Thus 400 remains an experiment-specific safety cap, not the global default.

These values are retained only as a conservative software-safety/candidate-generation preset, not as a scientifically validated RF or MTZ default and not as an automatic RF acceptance criterion.  The legacy `residual_L2_initial/final` fields remain for compatibility and single-record debugging only.  Cross-record QC should use the new dimensionless `residual_rms_final_fraction`, `residual_rms_reduction_fraction`, `residual_energy_final_fraction`, `residual_energy_reduction_fraction`, and the noise-normalized `ns_gid_residual_noise_rms_ratio`.  Very sparse, saturated, ringing, source-failed, transverse-dominated, or negative-lag-dominated records still require event-level editing or weighting.

### Synthetic regression

- The former raw-vector alignment maps controlled `0/45/72 s` phases to approximately `+5/+50/+77 s` for analysis `t0=-10 s` and wavelet `t0=-5 s`.
- Both engines now recover `0/45/72 s` within `0.10 s` under the exact shipped convergence defaults.
- Wavelet starts `-8/-5/-3 s` preserve the same physical phase times.
- Equivalent internal and external `TimeSeries` wavelets preserve the same lags.
- Noise windows `(-35,-5)`, `(-100,-10)`, and `(-200,-10)` give stable RMS stopping rather than raw-L2 length dependence.
- Direct `engine.actual_output()` and `engine.output_shaping_wavelet()` results are live `TimeSeries` objects with preserved data/timebase and pass directly to `WindowData`.

### Real-data result

All 332 selected records completed for both engines and all tested policies; the exact baseline used all 3,801 noise samples and `nfft=8192`.  The longitudinal/source response is centered at zero, and no systematic `+5 s` timing shift remains.  Station-dependent radial energy near `+5 s` is consistent with a crustal conversion rather than a global time-origin error, but is not assigned a phase here.

Final default-profile comparison against the preceding artifact:

| Metric | TD | FD |
|---|---:|---:|
| valid records | 332/332 | 332/332 |
| waveforms pointwise unchanged | 329/332 | 328/332 |
| median iterations/spikes | 25 | 26 |
| median residual-energy fraction | .763 | .762 |
| records moved from cap=100 to fractional-floor stop | 3 | 4 |

The changed records stop at 21–90 accepted spikes instead of blindly reaching 100.  They lose mainly boundary/late overfit support; the fix does not manufacture coherent converted phases.

Figure-8-style station summaries (event mean, pointwise IQR, min–max, and fold) and slowness-sorted gathers show that TD/FD are reproducible, but this archive does **not** establish P410s/P660s detections: corrected coherence remains near a random `1/N` baseline.  Predicted IASP91 tracks are references, not detections.

The final waveform-only seismology review found the direct P within ±0.25 s for 78.9%/80.7% of TD/FD records, with no systematic +5 s shift.  However, station-internal LOO correlation is only .039/.034 in 2–20 s and .016/.018 in 20–100 s; median stacks are approximately zero.  H–κ uses about 299 records and has median bootstrap stability .445/.348, but several station peaks lie at 58–59.5 km near the H-grid upper boundary and the station RF stacks do not show a stable Moho family.  H–κ is therefore a crustal diagnostic, not evidence for P410s/P660s or an automatic acceptance rule.  The shipped .995 profile is explicitly labeled conservative/safe rather than science-ready.

### Commands and results

```text
cmake --build build                                           PASS
ctest                                                       11/11 PASS
pytest TD/FD GID + RFdeconProcessor + algorithm validation 436/436 PASS
git diff --check                                             PASS
Vista ARM64 final default profile                  TD 332/332, FD 332/332 valid
GitHub Actions: Python 3.10/3.13, CMake,
                x86/ARM64 dev/latest, MPI, sdist              PASS
Codecov project                                   66.42% (+1.96%) PASS
Codecov patch                                     78.94% (target 64.45%) PASS
```

The final full diff received two consecutive independent GPT-5.6 Sol High CLEAN release audits after the last code change, with zero P0–P2 findings.  The implementation was produced by a separate GPT-5.6 Terra High agent.  A third Sol High reviewer acting as a receiver-function seismologist inspected only the final real-data waveforms and concluded that the stopping fix removes a few overfit tails without a dataset-level regression, while the unedited default RFs remain too sparse and incoherent for crustal or MTZ imaging claims.

## Main files changed

- C++ GID headers/implementations and scalar-deconvolution overrides
- pybind deconvolution bindings and Python GID wrappers/processors
- all mirrored TD/FD PF defaults
- TD/FD regression tests and serialization/API coverage
- deconvolution user documentation

## Remaining limitations

- The real-data screen is a bounded 332-record/10-station archive study, not a representative mantle benchmark.
- The H–κ pilot is limited by strongly one-sided backazimuth coverage; no stable simple-crust Moho family was established, so its maxima are not reported as crustal estimates or used for automatic editing.
- Deep-phase interpretation still requires moveout correction, ray-parameter/backazimuth balance, editing/weighting, and bootstrap or held-out-event stability.
- Prebuilt extensions and downstream custom `ScalarDecon` overrides must rebuild for the intentional `TimeSeries` virtual-API migration.


wangyinz:
Final validation for head d37ae6ce:\n\n- All GitHub Actions checks pass: CMake (push and PR), Python 3.10/3.13 (push and PR), code format, dependency alignment, sdist, and all Docker amd64/arm64/MPI builds.\n- Codecov patch passes at 95.65789% (33 changed lines uncovered).\n- Codecov project passes at 65.30%, up 0.84 percentage points from the 64.45% base.\n- The final CI corrections were implementation-neutral test fixes and received three additional rounds of independent GPT-5.6 Sol High review; every round reported GO with no P0-P3 findings.\n- The seismology review remained strictly limited to the real-data receiver-function waveforms, as requested.

PR #769, opened 2026-08-09
Keep long MongoDB scans alive

### LINKED ISSUE #293: cursor timeout may be a pervasive issue
I just hit  a MongoDB cursor timeout in a different context that is a big problem.   I was running a pure database script with MongoDB cleaning up a set of READONLYERROR attributes that had to be repaired.  (full up extended usarray 2012 data set) The problem was I was running that job on a database with over 3 M documents.   It will no doubt take hours to do the repairs and sure enough I got a timeout error when I ran this seemingly innocent script:
```
rename={ "READONLYERROR_net" : "net",
        "READONLYERROR_sta" : "sta",
        "READONLYERROR_chan" : "chan",
        "READONLYERROR_loc" : "loc"}
db.clean_collection('wf_TimeSeries',rename_undefined=rename,verbose=False)
```

A BIG PROBLEM here for our users is that we hide the cursor inside Database.  There is currently no way to tell clean_collection that it should create an immortal cursor when the user can anticipate a task will take more than 30 minutes (default cursor timeout) to do the updates.   

We immediately need to go through Database and in all places this can be an issue we need to add a "no_cursor_timeout" boolean argument (another name than MongoDBs might be better but may just cause confusion.  e.g. we could use a key arguments `use_immortal_cursor=False` or something else.   Suspect using the mongodb name would be less confusing to users)  

This also brings to the front the need to fully test the `with -mongoquery- as cursor` construct discussed elsewhere (I don't  know how you do internal links in github - sorry).   A pending question is how to assure an immortal cursor can actually be killed.   As discussed earlier if a parallel job did a lot of queries that never timed out we know from experience it can crash mongodb because the gods (the immortal cursors) have all become decadent and are doing nothing but blocking the entrance of valhalla for anyone else.  (Couldn't resist that imagery.  In computer terms the threads handling each connection never exit and the server crashes when some limit is hit).   From my reading I think the `with blah as cursor` construct solves this problem but that is purely theoretical and needs a test before we use it through all of MsPASS.  Advise we check that and fix this asap.

wangyinz:
## Summary

- add one managed cursor path for every cursor created inside `Database`
- opt in to `no_cursor_timeout` with an explicit MongoDB session, refresh that session between documents before the next cursor fetch, and close both cursor and session reliably
- expose the option through `clean_collection`, collection-wide maintenance/verification helpers, and `read_inventory`
- expose `--no-cursor-timeout` in `dbclean` and `dbverify`, with API and user-guide documentation
- make the remaining small inventory lookup cursors deterministically close on normal and early-return paths
- cover default and long-running cursor behavior plus the affected database and CLI paths

MongoDB logical session expiry can invalidate a cursor even when PyMongo's `no_cursor_timeout=True` is used on its own. This implementation follows MongoDB's documented pattern of pairing the option with an explicit, periodically refreshed session.

## Verification

- `48 passed` across `test_database.py`, `test_dbclean.py`, and `test_dbverify.py` against MongoDB 4.4
- focused managed-cursor unit tests: `2 passed`
- Sphinx HTML build completed with exit code 0 (only existing offline intersphinx warnings)
- Black, Python byte-compilation, and `git diff --check` passed

Fixes #293

### DIFF
--- a/docs/source/user_manual/mongodb_and_mspass.rst
+++ b/docs/source/user_manual/mongodb_and_mspass.rst
@@ -391,7 +391,7 @@ long.  A direct PyMongo loop should use a ``with`` statement so the cursor is
 always closed, although that context manager does not prevent expiry.
 Setting ``no_cursor_timeout=True`` alone is not sufficient for an indefinitely
 long scan because MongoDB's logical session timeout still applies.  The MsPASS
-``clean_collection`` method and the
+``clean_collection`` and ``read_inventory`` methods and the
 ``dbclean`` and ``dbverify`` commands provide an opt-in
 ``no_cursor_timeout``/``--no-cursor-timeout`` mode for large data sets.  That
 mode opens an explicit session, refreshes it periodically, and closes both the
--- a/docs/source/user_manual/mongodb_and_mspass.rst
+++ b/docs/source/user_manual/mongodb_and_mspass.rst
@@ -386,6 +386,20 @@ The examples above show the most common, simple usage.
 Both accept a required-or-defaulted filter and an optional projection;
 additional behavior is controlled by documented keyword arguments.
 
+MongoDB can expire a cursor when processing one returned batch takes too
+long.  A direct PyMongo loop should use a ``with`` statement so the cursor is
+always closed, although that context manager does not prevent expiry.
+Setting ``no_cursor_timeout=True`` alone is not sufficient for an indefinitely
+long scan because MongoDB's logical session timeout still applies.  The MsPASS
+``clean_collection`` method and the
+``dbclean`` and ``dbverify`` commands provide an opt-in
+``no_cursor_timeout``/``--no-cursor-timeout`` mode for large data sets.  That
+mode opens an explicit session, refreshes it periodically, and closes both the
+cursor and session when the scan ends.  See MongoDB's `cursor timeout
+documentation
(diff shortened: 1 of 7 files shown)


wangyinz:
Independent review of HEAD `88814279d4854945722b27064c15d9f7b5c09a5d` completed with **no actionable findings**.

The reviewer confirmed that this PR now adequately addresses #293: all cursors created inside `Database` use the managed path; long scans combine `no_cursor_timeout=True` with an explicit periodically refreshed session; cursor/session cleanup covers normal, early-exit, and exception paths; appended opt-in parameters preserve default compatibility; CLI propagation, tests, and documentation are consistent.

Independent read-only checks included the exact base-to-HEAD diff, all `Database.find` call sites, syntax/diff checks, and the 48-test related regression result. The PR remains draft and unmerged while final CI runs.

wangyinz:
Final CI for HEAD `88814279d4854945722b27064c15d9f7b5c09a5d` completed successfully.

All six workflows concluded `success`:
- Dependency Alignment
- code-format
- CMake build/test
- Python package on Python 3.10 and 3.13
- Docker builds
- source distribution build

Publish/deploy-only jobs and steps that do not run for this draft PR were conditionally skipped; there were no failed or pending executable checks.

Together with the independent review (no actionable findings), PR #769 has passed the requested review and CI gates for issue #293. It remains draft and unmerged.

PR #770, opened 2026-08-09
Allow forcing a PIC-safe bundled Boost

### LINKED ISSUE #446: mismatch in anaconda version of boost with our cmake build
I was getting some classic conflict problems from library incompatibilities on my local ubuntu machine I use for mspass development.   The problem was coming from a library I had installed via anaconda (matplotlib which is pretty fundamental) so I uninstalled and reinstalled the latest version of anaconda.  

Problem that cause is it breaks local compiles of the cxx code base.   ccmake shows me cmake is looking for boost in the libraries in the anaconda tree.   It apparently did that in the configuration step through the magic it uses for finding such packages.   Anyway, when I compile the cxx parts of mspass I'm getting this fatal error from the build process:
```
  ...
[ 75%] Linking CXX shared module io.cpython-311-x86_64-linux-gnu.so
/usr/bin/ld: ../../src/lib/libmspass.a(fileio.cc.o): relocation R_X86_64_PC32 against undefined hidden symbol `_ZTCN5boost10wrapexceptINS_12bad_any_castEEE0_NS_16exception_detail10clone_implINS3_19error_info_injectorIS1_EEEE' can not be used when making a shared object
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
```
A google search quickly shows this error is linked to a library being used in the link process that was not linked with the magic linux flag -fKpic.  From the error I appears to be from boost.  I've seen a problem before in boost with a problem mixing up static and dynamic libraries.  It looks to me like anaconda's build has that issue.   

I am recording this for the record.  I think the only solution is to build a local copy of boost manually and make sure the static libraries are not mixed up with or compiled with -fKpic.  I remember going into this dark alley a while back so I'm pretty sure I can solve it that way.  Alternatively, is there a way to set up cmake to force it to download and build boost?

wangyinz:
## Summary

- add a default-off `MSPASS_FORCE_BUNDLED_BOOST` CMake option that bypasses Conda/system Boost discovery
- after building the pinned Boost 1.86 dependency, clear stale Boost cache entries (including CMake 3.11's `Boost_LIBRARY_DIR`) and restrict the final lookup to the bundled install prefix
- prevent mixed Boost headers/libraries from separate prefixes
- document a CMake 3.11-compatible `cmake` and `ccmake` workflow for source builds

The bundled Boost path already builds static Boost.Serialization with `-fPIC`. This change gives developers an explicit escape hatch when an installed static Boost is incompatible with the Python extension link.

## Verification

- adversarial forced configure with both a different complete Boost 1.86 prefix and an explicit external `Boost_LIBRARY_DIR` still resolved headers, library directories, and `libboost_serialization.a` exclusively from the bundled build
- full forced C++ build completed successfully, including all Python extension link steps
- CTest: 11/11 passed
- default-off configure continued to select the compatible installed Boost
- Sphinx HTML build completed with exit code 0 (only existing offline intersphinx warnings)
- `git diff --check` passed

Fixes #446

### DIFF
--- a/cxx/CMakeLists.txt
+++ b/cxx/CMakeLists.txt
@@ -48,12 +48,16 @@ message (STATUS "YAML_CPP_INCLUDE_DIR = ${YAML_CPP_INCLUDE_DIR}")
 
 # Boost.Serialization archives are shared across MsPASS builds, so the Boost
 # version is a compatibility constraint rather than a minimum dependency.
+option (MSPASS_FORCE_BUNDLED_BOOST
+  "Ignore installed Boost libraries and build the pinned bundled Boost" OFF)
 set (Boost_USE_STATIC_LIBS ON)
 if (DEFINED ENV{MSPASS_BOOST_ROOT} AND NOT "$ENV{MSPASS_BOOST_ROOT}" STREQUAL "")
   set (BOOST_ROOT "$ENV{MSPASS_BOOST_ROOT}")
 endif ()
-find_package (Boost 1.86.0 EXACT COMPONENTS serialization)
-if (NOT Boost_FOUND)
+if (NOT MSPASS_FORCE_BUNDLED_BOOST)
+  find_package (Boost 1.86.0 EXACT COMPONENTS serialization)
+endif ()
+if (MSPASS_FORCE_BUNDLED_BOOST OR NOT Boost_FOUND)
   message (STATUS "Building Boost")
   include (cmake/boost.cmake)
   fetch_boost (
--- a/cxx/cmake/boost.cmake
+++ b/cxx/cmake/boost.cmake
@@ -33,9 +33,34 @@ macro(fetch_boost _download_module_path _download_root)
     endif()
 
     set (BOOST_ROOT ${BOOST_INSTALL_ROOT})
+    set (BOOST_INCLUDEDIR ${BOOST_INSTALL_ROOT}/include)
+    set (BOOST_LIBRARYDIR ${BOOST_INSTALL_ROOT}/lib)
     set (Boost_NO_BOOST_CMAKE ON)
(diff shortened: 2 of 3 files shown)


wangyinz:
Independent review completed for exact HEAD `aab34bac93fdc13797aadc59c4670b740a4faee6`.

Result: **no actionable findings; this revision fully addresses #446.**

The reviewer independently verified that:

- both the normal and cache forms of CMake 3.11's `Boost_LIBRARY_DIR` are cleared;
- an adversarial forced configure with an external `Boost_LIBRARY_DIR` and external `CMAKE_PREFIX_PATH` still resolves Boost headers, library directories, and `libboost_serialization.a` exclusively from the bundled prefix;
- all Python extension link commands use the bundled archive and the original `CMAKE_PREFIX_PATH` is restored;
- the documented commands are compatible with the project's minimum CMake 3.11;
- default-OFF behavior, `MSPASS_BOOST_ROOT`, static/PIC behavior, and later dependency lookup remain unchanged.

Local evidence for this exact revision also includes a successful forced build, 11/11 CTest tests, Sphinx HTML build, and `git diff --check`.

The PR remains draft and unmerged while CI for this exact HEAD completes.

wangyinz:
Final CI gate passed for exact HEAD `aab34bac93fdc13797aadc59c4670b740a4faee6`.

All pull-request workflow runs completed successfully:

- CMake — success
- Python package (Python 3.10 and 3.13 jobs, including pytest and documentation) — success
- Docker (release/dev, amd64/arm64, MPI/GeoLab paths) — success
- code-format — success
- Dependency Alignment — success
- Publish to PyPI workflow — success

Together with the completed independent review (no actionable findings), PR #770 has cleared the requested gate for issue #446. It remains **draft and unmerged**.

PR #772, opened 2026-08-09
Resolve packaged data without MSPASS_HOME

### LINKED ISSUE #550: MSPASS_HOME should not be required
Right now, in our C++ code, we rely on the following function to determine the default directory for the parameter files:
https://github.com/mspass-team/mspass/blob/bf7e2f05609888b56a238095ff2aa3527e304694/cxx/src/lib/utility/data_directory.cc#L7
Basically, a `MSPASS_HOME` environment variable needs to be set for any of those C++ code to work.

In the Python code, such requirement is avoided by the following constructs:
https://github.com/mspass-team/mspass/blob/bf7e2f05609888b56a238095ff2aa3527e304694/python/mspasspy/db/schema.py#L17-L26
Because we always install the `data` directory with the Python package, we can use `__file__`  (which is the pathname of the file from which the running Python module was loaded) to locate the relative path to the `data` directory .

I should also note that there are Python codes that didn't use this robust method:
https://github.com/mspass-team/mspass/blob/bf7e2f05609888b56a238095ff2aa3527e304694/python/mspasspy/preprocessing/css30/datascope.py#L87-L95

Now, I think we should improve the C++ code such that it is not necessary to define `MSPASS_HOME` especially when it is compiled and installed as a Python modules. This can be done by using CMake to set a variable to tell the code that it is being built as a Python module. Then, CMake can also set a global variable to code in the relative path to the data directory within the python installed directory.

wangyinz:
## Summary

- Keep the standalone C++ core free of Python runtime/link dependencies.  The Python `utility` binding registers the installed `mspasspy/data` path when the module initializes; `MSPASS_HOME/data` remains the core override when the environment variable is set.
- Let `AntelopePf` load the MsPASS base parameter file before its existing `PFPATH`/current-directory chain, preserving last-file-wins overrides.
- Accept relative parameter-file base names with or without the documented `.pf` suffix.
- Make `DatascopeDatabase` use installed package data when `MSPASS_HOME` is absent, and update the C++/Python/Conda/user-manual documentation.
- Add Python regression coverage plus a standalone CMake test that deliberately links without Python.

Fixes #550

## Validation

- Built all C++ libraries, tests, and Python extension modules from the implementation in `mspass/mspass:latest` (Python 3.12); exact HEAD `83d5d9ddd04d61570b13d45f35602823bbdeaa7f` adds only the independently requested API-documentation clarification.
- Passed the new standalone `test_data_directory` CTest; its dynamic dependencies contain no `libpython`.
- From an unrelated temporary directory, verified packaged `AntelopePf` loading with both `attribute_maps` and `attribute_maps.pf`, `PFPATH` merging/last-file-wins, and `MSPASS_HOME` replacing the packaged base.
- Verified `DatascopeDatabase` loads bundled `DatascopeDatabase.pf` with both `MSPASS_HOME` and `PFPATH` unset.
- Black, clang-format checks on changed C++ ranges/files, Python syntax compilation, a direct standalone no-Python compile/link/run, and `git diff --check` passed.

### DIFF
--- a/cxx/include/mspass/utility/AntelopePf.h
+++ b/cxx/include/mspass/utility/AntelopePf.h
@@ -65,13 +65,14 @@ public:
   /*! \brief Construct from a base pf name.
 
   This constructor acts like the antelope pfread function for
-  parameter files constructed in plain C.  A relative base name is first
-  read from the MsPASS data directory: MSPASS_HOME/data when MSPASS_HOME is
-  set, or the installed mspasspy package data otherwise.  The constructor
-  then follows the chain of directories defined by PFPATH; if PFPATH is not
-  defined, that path defaults to ".".  As with the antelope pfread procedure,
-  the last file read takes precedence, so PFPATH or the current directory can
-  override the MsPASS data file.
+  parameter files constructed in plain C.  A relative base name is first read
+  from the MsPASS data directory when available: MSPASS_HOME/data when
+  MSPASS_HOME is set, or the installed mspasspy package data after the Python
+  extension registers it.  A standalone caller without MSPASS_HOME instead
+  starts with the directories defined by PFPATH; if PFPATH is not defined,
+  that path defaults to ".".  As with the antelope pfread procedure, the last
+  file read takes precedence, so PFPATH or the current directory can override
+  the MsPASS data file.
   Further if pfbase begins with a slash (i.e. "/") it is assumed to be
   the actual file name to read and the PFPATH feature is disabled.
 
--- a/cxx/include/mspass/utility/AntelopePf.h
+++ b/cxx/include/mspass/utility/AntelopePf.h
@@ -65,15 +65,18 @@ public:
   /*! \brief Construct from a base pf name.
 
   This constructor acts like the antelope pfread function for
-  parameter files constructed in plain C.   That is, the argument is
-  a base name sans the .pf extension.   Like antelope's pfread it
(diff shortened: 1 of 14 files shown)


wangyinz:
Independent review gate passed for exact HEAD `83d5d9ddd04d61570b13d45f35602823bbdeaa7f`.

The reviewer rechecked the complete base-to-HEAD diff and reported no actionable findings. The review independently verified the standalone no-`libpython` contract, Python package-data registration, `MSPASS_HOME` selection, current-directory/absolute/multi-entry `PFPATH` behavior and last-file-wins, optional `.pf` suffix handling, Datascope fallback, import/reload/thread behavior, packaging, and API compatibility. The final documentation-only follow-up now accurately distinguishes Python-registered package data from standalone `PFPATH`/current-directory behavior.

Conclusion: this draft PR sufficiently addresses issue #550. It remains unmerged; the final gate is CI on this exact SHA.

wangyinz:
CI gate passed for exact HEAD `83d5d9ddd04d61570b13d45f35602823bbdeaa7f`.

All pull-request workflows completed successfully:

- Dependency Alignment — success
- Publish to PyPI validation — success
- code-format — success
- CMake build and CTest — success
- Docker matrix — success
- Python package — success (both Python 3.10 and 3.13 jobs, including full pytest; documentation build also succeeded)

Together with the independent no-findings review recorded above, the review/CI gate for issue #550 is complete. This PR intentionally remains draft and unmerged.

PR #773, opened 2026-08-09
Add regression coverage for empty distributed ensemble reads

### LINKED ISSUE #651: Possible bug in read_distributed_data
I encountered a relatively minor bug in read_distributed_data.   I encountered the problem in a workflow parallel in ensembles.  For that context read_distributed_data expects to receive a list of queries (python dict containers).  The bug I encountered is only encountered when I had an error in the usage that caused all the queries to return nothing.   The particular way that happened was to specify a 'data_tag' value that did not exist in the input defined by the "collection" argument.   In particular, I intended to have `collection="wf_TimeSeries"` but specified `collection="wf_Seismogram"` by mistake.   The data tag value I used did not exist in wf_Seismogram as that collection didn't even exist at that point in this workflow.  

I am not exactly sure what happened with this error.   What I do know is it created mysterious errors downstream and I'm not quite sure why.   The function `read_ensemble_parallel` should be called when arg0 of read_distributed_data is a list of dictionaries.   This is the entire body of that function:
```
    if sort_clause:
        cursor = db[collection].find(query).sort(sort_clause)
    else:
        cursor = db[collection].find(query)
    ensemble = db.read_data(
        cursor,
        collection=collection,
        mode=mode,
        normalize=normalize,
        load_history=load_history,
        exclude_keys=exclude_keys,
        data_tag=data_tag,
        aws_access_key_id=aws_access_key_id,
        aws_secret_access_key=aws_secret_access_key,
    )
    kill_me = True
    for d in ensemble.member:
        if d.live:
            kill_me = False
            break
    if kill_me:
        ensemble.kill()
    return ensemble
```
which shows that this is largely a wrapper on `Database.read_data` for the case with a cursor input.  When arg0 is a cursor this section of read_data should always be entered:
```
        elif isinstance(id_doc_or_cursor, pymongo.cursor.Cursor):
            for doc in id_doc_or_cursor:
                if data_tag:
                    # silently drop any documents that do not have data tag
                    # and match input value
                    if "data_tag" in doc and doc["data_tag"] == data_tag:
                        doclist.append(doc)
                else:
                    doclist.append(doc)
            reading_atomic_data = False
            # return default constructed ensemble with a error message if
            # editing for data tag above threw out all data
            if len(doclist) == 0:
                message = "Found no data matching data_tag={} in list created from cursor\n".format(
                    data_tag
                )
                message += "Returning an empty ensemble marked dead"
                if object_type is TimeSeries:
                    val = TimeSeriesEnsemble()
                else:
                    val = SeismogramEnsemble()
                val.elog.log_error("Database.read_data", message, ErrorSeverity.Invalid)
                return val
```
which shows it should always return a dead, empty ensemble container in this situation.   Debugging showed I actually got a dead Seismogram object, which makes no sense.   I double checked the pybind11 code and see no reason the call line `val = SeismogramEnsemble` above could generate anything but the expected empty, dead SeismogramEnsemble.   I can only guess that something downstream in this workflow was actually failing when it received the empty ensemble object and I misread teh exception messages.   Unfortunately, I'm not able to exactly recreate this error without a lot of headaches.  

The solution to this issue is to add a directed set of tests for read_distributed_data that recreates the problem of all failed queries.  This is not a high priority fix as the problem it seems to only exist when I made the mistake noted above.  Fix the mistake and the problem goes away.

wangyinz:
## Summary

- add directed Dask and Spark coverage for ensemble-parallel reads when every query yields no waveform data
- cover both a valid collection filtered by a nonexistent `data_tag` and the reported wrong/nonexistent collection case
- verify each query returns an empty, dead ensemble of the type implied by the requested collection, with the expected diagnostic

The current production path already has the intended behavior; this PR makes the reported failure modes explicit regression contracts.

## Testing

- `pytest -vv python/tests/io/test_distributed.py::test_read_distributed_ensemble_all_queries_empty[dask]` (real MongoDB)
- `pytest -vv python/tests/io/test_distributed.py::test_read_distributed_ensemble_all_queries_empty[spark]` (real MongoDB, Java 17)
- `black --check python/tests/io/test_distributed.py`
- `python -m py_compile python/tests/io/test_distributed.py`
- `git diff --check`

Fixes #651

### DIFF
--- a/python/tests/io/test_distributed.py
+++ b/python/tests/io/test_distributed.py
@@ -27,6 +27,7 @@ from mspasspy.io.distributed import (
     read_to_dataframe,
 )
 from mspasspy.db.normalize import ObjectIdMatcher
+from mspasspy.ccore.seismic import TimeSeriesEnsemble, SeismogramEnsemble
 from mspasspy.ccore.utility import ErrorSeverity
 from mspasspy.util.db_utils import MongoDBWorker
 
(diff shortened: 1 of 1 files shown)


wangyinz:
Independent review completed against exact head `c8ae72609c5cffa561eb19abdf74b88e1bdb35fb`: **no findings**.

The reviewer verified that the test inputs exercise the real list-of-dictionaries ensemble path through `read_distributed_data` and `read_ensemble_parallel`, covering both (1) matching `wf_TimeSeries` documents fully removed by a nonexistent `data_tag` and (2) the reported nonexistent/wrong `wf_Seismogram` collection. Each query checks the exact ensemble class, dead state, zero members, and `Database.read_data` diagnostic. The reviewer independently ran both Dask and Spark/Java 17 variants with real MongoDB (`2 passed`) and confirmed fixture/database isolation.

The existing production cursor path already constructs the intended typed empty ensemble and the wrapper marks it dead. Since #651 explicitly asks for directed regression tests after the downstream error could not be reproduced, the reviewer concluded this test-only change fully addresses the issue without speculative production changes.

wangyinz:
CI gate completed for exact head `c8ae72609c5cffa561eb19abdf74b88e1bdb35fb`: all 6 pull-request workflows passed.

- Python package: success on Python 3.10 and 3.13, including the full pytest suite; documentation build also passed on 3.10
- CMake: success
- Docker: success
- code-format: success
- Dependency Alignment: success
- Publish-to-PyPI validation: success

The PR remains draft, open, and unmerged.

PR #774, opened 2026-08-09
Preserve ensemble read failure diagnostics

### LINKED ISSUE #670: Minor bug in ensemble reader
There is a minor bug in `mspasspy.db.database.Database.read_data` for the case when arg0 is a MongoDB command cursor - the standard way to read an ensemble of data indexed in MongoDB.   The bug is "minor"  because it doesn't cause a failure.  The problem the bug creates is read error messages posted to an ErrorLogger container in the method of the `Database` class called `doclist2mdlist` are not posted to the failed datum but the function returns a single ErrorLogger container with error messages from all read failures for the ensemble data (cursor content) it was handling.   The output from `doclist2mdlist` is then not handled correctly either and any read error message get dropped.   The net result is that if a workflow generates a bunch of read error, which translate in the language of the `Database` class to "abortions", the error messages posted for those abortions if these data are passed through the system vanish.   That makes it challenging to figure out what kind of read problem was causing the reader to fail.   

A less serious additional issue is that data in ensembles with read failures should be saved in an "abortions" collection but I find they are actually being placed in the "cemetery" collection.   Fixing that requires properly handling of the "is_abortion" boolean in the appropriate methods.   

It is easier to fix this bug than design tests to validate it has been fixed.   I think the easiest solution is to have the `mspasspy.db.database.Database.doclist2mdlist` function return a list instead of doing what it does not which is post to the ensemble error log container.  Let me be specific.  The offending lines are these (currently 8193 to 8198):
```
        if live:
            mdlist.append(md)
        else:
            bodies.append(doc)
        if elog.size() > 0:
            ensemble_elog += elog
```
That should be:
```
        if live:
            mdlist.append(md)
        else:
            bodies.append(doc)
            eloglist.append(elog)
```
That would have to then be handled differently downstream and the documentation changed as currently elog is a single ErrorLogger container not a list of ErrorLogger containers.    There is a minor flaw in this algorithm as if a reader posts a warning message but doesn't set "live" to False the warning message would be dropped.   Fixing that would require changing the returns of doclist2mdlist further.   Since we need to do that anyway we might consider changing it completely.   One approach would be to return a tuple with tuples.  i.e. instead of the current return line:
```
return [mdlist, live, ensemble_elog, bodies]
```
it would be 
```
return [ [mdlist,read_warnings],[bodies,eloglist] ]
```
where `read_warnings` would be a list set exactly like eloglist but for would either contain None or ErrorLogger containers with warnings.   It would then be a parallel list to mdlist while eloglist would be the parallel container for bodies.   

One of us needs to fix this but I emphasize it isn't critical as it just produces mysterious results and doesn't cause a workflow to abort.

wangyinz:
## Summary
- keep each schema-rejected waveform document paired with its own `ErrorLogger`
- save rejected TimeSeries and Seismogram documents as typed abortions with `is_abortion` and their read diagnostics intact
- retain warnings from accepted documents on the returned ensemble and fix the unreachable error-log guard
- add a MongoDB-backed regression test covering both waveform types, warning retention, error-log isolation, abortions, and cemetery behavior

## Validation
- `python/tests/db/test_database.py::TestDatabase`: **42 passed**
- directed TimeSeries/Seismogram regression after final assertion: **1 passed**
- Black checks for the changed test and production-code ranges
- Python syntax compilation and `git diff --check`

Fixes #670

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -988,8 +988,17 @@ class Database(pymongo.database.Database):
             # kind of data returned in the "abortions" list.
             # See Undertaker docstring or User Manual for concepts
             if len(abortions) > 0:
-                for md in abortions:
-                    self.stedronsky.handle_abortion(md)
+                for doc, abortion_elog in abortions:
+                    if object_type is TimeSeries:
+                        abortion = TimeSeries()
+                    else:
+                        abortion = Seismogram()
+                    for key in doc:
+                        abortion[key] = doc[key]
+                    abortion.elog += abortion_elog
+                    abortion["is_abortion"] = True
+                    abortion.kill()
+                    self.stedronsky.handle_abortion(abortion)
             if live and len(mdlist) > 0:
                 # note this function returns a clean ensemble with
                 # no dead data.  It will be considered bad only if is empty
@@ -1062,7 +1071,7 @@ class Database(pymongo.database.Database):
                     ensemble = SeismogramEnsemble()
             # elog referenced here comes from doclist2mdlist - a bit confusing
             # but delayed to here to construct ensemble first
-            if elog.size() < 0:
+            if elog.size() > 0:
                 ensemble.elog += elog
 
             # We post this even to dead ensmebles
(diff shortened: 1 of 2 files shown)


wangyinz:
Independent review completed against exact HEAD `02e6ba8fe7c22ba8ec19a5d560c24dc6446638e2`: **no findings**.

The reviewer confirmed that PR #774 sufficiently addresses #670, including:
- per-document `doc2md` errors remain paired with the correct failed document;
- typed TimeSeries/Seismogram abortion records preserve tombstones, `is_abortion`, and `logdata`;
- accepted-document warnings reach the ensemble while failed-document diagnostics do not leak into it;
- partial-failure and all-failed paths produce abortions rather than cemetery records;
- the only repository call site and the updated `doclist2mdlist` return contract agree.

Independent validation passed the new regression, all 42 `TestDatabase` tests, and an additional real-MongoDB all-failed probe with two distinct schema errors for both waveform types (2 correctly paired abortions, 0 cemetery in each case).

wangyinz:
CI completed successfully for exact HEAD `02e6ba8fe7c22ba8ec19a5d560c24dc6446638e2`: **6/6 workflows passed**.

- CMake: success
- Python package: success (Python 3.10 and 3.13 pytest jobs both succeeded)
- Docker: success
- code-format: success
- Dependency Alignment: success
- Publish to PyPI: success

The PR remains draft, open, and unmerged.

PR #885, opened 2026-08-12
Fix sliding-window pipeline result and resource contracts

### LINKED ISSUE #783: [P1][Workflow] Restore sliding-window result and resource contracts
## Baseline

## Problem
Default mode never appends processing results, the advertised `Iterable` is later sized/indexed, non-positive window/task values are accepted, failures leave futures outstanding, and membership-only tests pass vacuously.

## Impact
Callers silently lose results and generator/error paths fail or retain cluster resources.

## Acceptance criteria
Materialize the public `dlist` Iterable exactly once into one list before validation or submission, and use only that list afterward. `sliding_window_size` must be literal `"auto"` or a positive integer other than `bool`; `task_per_worker` must be a positive finite real number other than `bool`, and an automatic window with zero workers or a rounded size below one is a `ValueError`. Define exactly four modes: with neither completion nor accumulator, return the processing results as a list; with completion and no accumulator, return the completion results as a list; with both, return the final accumulator value; with accumulator but no completion, raise `ValueError` before submitting work. Lists and accumulation follow Dask Future completion order and do not promise input order. On submit, Future result, completion, or accumulator failure, cancel every outstanding Future and re-raise the original exception. Limit the PR to `sliding_window_pipeline` and its tests.

## Required tests
For all three valid modes, cover empty input and exact output count/content with no omission or duplication; force out-of-input-order completion and assert list/accumulator completion order. Use a one-shot generator that records exactly one iteration. Cover bool, zero, negative, NaN, infinite, zero-worker, rounded-zero, and invalid mode cases plus submit, processing, completion, and accumulator exceptions on a local Dask cluster, asserting all outstanding Futures are cancelled.

## Related
#713/#714 fixed explicit size and #723/#766 fixed progress logging, not these contracts.

wangyinz:
## Summary

- Materialize each input iterable exactly once and preserve duplicate inputs.
- Return processing or completion results in actual Future completion order.
- Bind as_completed to the explicitly supplied Dask client loop, including when another client is the process default.
- Validate window settings before submission and cancel every outstanding Future on failures.
- Preserve the function-scoped Dask client lifecycle introduced by #1004.

## Validation

- python/tests/test_workflow.py: 32 passed
- Black, Python byte compilation, and git diff --check

Closes #783

### DIFF
--- a/python/mspasspy/workflow.py
+++ b/python/mspasspy/workflow.py
@@ -1,6 +1,8 @@
-import dask.distributed as ddist
-
+import math
 from collections.abc import Mapping, Iterable
+from numbers import Integral, Real
+
+import dask.distributed as ddist
 
 
 def sliding_window_pipeline(
@@ -63,7 +65,7 @@ def sliding_window_pipeline(
     The function uses dask Futures (note Futures are supported only in dask
     so this function cannot be used with a pyspark workflow) and submits
     and runs only sliding_window_size instances of the task defined by
-    the "completion_function" argument.  The default behavior runs
+    the "processing_function" argument.  The default behavior runs
     approximately 2 instances of the function per worker.  The number is
     "approximate" because the scheduler algorithm often queues up more
     instances on startup and the load is unbalanced until all workers are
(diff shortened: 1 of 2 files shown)


wangyinz:
Review status: Draft. Before merge: (1) rebase must retain the function-scoped Dask fixture and isolation regressions from #1004; (2) construct `as_completed` with the supplied `dask_client.loop` and add a non-default-client regression. The current no-loop construction calls `default_client()` and can bind to a stale or closed client. No maintainer behavior decision is required.

wangyinz:
Resolved the review blocker: as_completed is now bound to the explicitly supplied client loop, even when a different default client is active. The #1004 function-scoped client lifecycle and post-close compute regression remain intact. The focused workflow suite passes all 32 tests; no further human decision is required for this PR.

PR #891, opened 2026-08-12
Fix AGC window indexing and gain-function state

### LINKED ISSUE #843: Repair AGC window indexing and returned gain-function state
## Audit baseline
`origin/master@61b06b965265a1327b84a8c6b6d7de491d9db97e`; findings CXX-3, CXX-4, CXX-5, CXX-6.

## Problem
AGC mixes ramp and sliding-window indices, clears the gain vector while sizing it, offsets its time axis, and leaves invalid and zero-energy cases without one result rule.

## Impact
Impulses can be erased, callers can receive an empty or shifted gain function, and bad inputs can partly mutate data.

## Acceptance criteria
- Before mutation, throw `MsPASSError` with `ErrorSeverity::Invalid` when `twin` or input `dt` is nonfinite or nonpositive, or when `npts == 0`.
- For every valid input set `half_window = min(floor(round(twin / dt) / 2), floor((npts - 1) / 2))`; gain sample `i` uses the inclusive interval `[max(0,i-half_window), min(npts-1,i+half_window)]`.
- For a window containing `N` three-component samples, set `energy = sum(x[k,j]^2)` over all three components and all `N` samples; use gain `1/sqrt(energy/(3*N))` when `energy>0`, otherwise use gain `0.0`.
- Apply gain sample `i` once to all three components at input sample `i`; return exactly `npts` gain samples with the input `t0` and `dt`.

## Required tests
- Compare every gain and output sample with the formula above for constant, all-zero, and edge/interior impulse data.
- Cover `npts` 0, 1, and 2; nonzero `t0`; invalid `twin`/`dt`; equal-length and oversized positive windows; assert no mutation on every rejected input.

## Related issues

## Re-audit disposition

The published acceptance criteria authorized replacing the documented Seismic Unix cumulative/ramp AGC with an uncited centered-RMS algorithm and changed the public error contract. PR #891 was therefore closed. This issue is closed as not planned; any future issue must preserve the established AGC definition and detailed documentation while narrowly fixing bounds, sizing, time origin, and zero-energy handling.

wangyinz:
Fixes #843

## Summary

- validate `twin`, `dt`, and empty inputs before mutating the waveform
- derive the bounded half-window from `round(twin/dt)` and compute each inclusive three-component RMS window exactly once
- define zero-energy gain as zero and apply each gain to the matching sample only
- preserve the gain function time grid and commit the output only after a successful full computation
- add Release-safe native contract tests and Python binding regressions

## Validation

- Release full native build: passed
- CTest: 13/13 passed (`test_agc` 1/1)
- rebuilt Python bindings, confirmed loaded from the isolated build tree: 19/19 passed
- `clang-format --dry-run --Werror`: passed
- Black check: passed
- `py_compile`: passed
- `git diff --check`: passed

## Independent agent review

A separate agent that did not implement this change re-read #843, audited the mutation/error/window/gain/time-axis contracts, rebuilt the Release target and bindings, reran the native and Python tests, and returned **REVIEW PASS** with no blocker. No merge was performed.

### DIFF
--- a/cxx/include/mspass/algorithms/algorithms.h
+++ b/cxx/include/mspass/algorithms/algorithms.h
@@ -13,16 +13,10 @@ namespace mspass::algorithms {
 /* \brief Apply agc operator to three component seismogram data.
 
    Automatic gain control (agc) is a standard operation in seismic
-reflection processing.  The algorithm used her is a variant of that in
-seismic unix but applied to vector data.   That is scaling is no
-determined by absolute value of each sample but th vector amplitude of
-each sample.  Scaling is determined by the average vector amplitude
-over a specified time window length.  There isa  ramp in and ramp off
-range of size equal to the window length.   agc was notorious in the
-early days of seismic processing for making it impossible to recover
-true amplitude.   We remove that problem here by returning a TimeSeries
-object whose contents contain the gain applied to each sample of the
-original data.
+reflection processing.  Each three-component sample is scaled by the inverse
+RMS amplitude in a centered, inclusive window.  Windows are truncated at data
+boundaries.  A zero-energy window has zero gain.  The gain applied at every
+sample is returned so callers can inspect the operation.
 
 \param d - data to apply the operator to.  Note it is altered.
 \param twin - length of the agc operator in seconds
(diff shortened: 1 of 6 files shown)


wangyinz:
Closing this implementation after the renewed scientific/API review.

The original public contract says this is the Seismic Unix-style cumulative/ramped vector-energy AGC, returns the gain needed to recover amplitude, and reports failures through the datum ErrorLogger without throwing. This PR removes that detailed contract and replaces the implementation with an uncited centered inclusive RMS algorithm plus a different exception ABI. Its formula test reproduces the new implementation rather than providing an independent seismic-processing oracle; even requested even-length windows become odd-length inclusive windows.

The bounds/vector-sizing defects remain valid, so Issue #843 should stay open. A replacement PR should preserve the documented cumulative/ramp algorithm and error contract while repairing its indexing/state defects, unless a separately reviewed scientific proposal supplies an authoritative algorithm reference, migration rationale, and independent golden data.

PR #896, opened 2026-08-12
Reject WindowData requests beyond the parent sample grid

### LINKED ISSUE #847: Reject C++ WindowData requests past the parent sample grid
## Audit baseline
`origin/master@61b06b965265a1327b84a8c6b6d7de491d9db97e`; findings CXX-12.

## Problem
WindowData can map its end index to `npts` and fabricate a live zero sample beyond the parent.

## Impact
A synthetic sample can look like valid scientific data and propagate through processing.

## Acceptance criteria
- Map requested endpoints with C++ `round((t-t0)/dt)` semantics.  An end at the final real sample succeeds and includes it; an end just below `endtime+0.5*dt` succeeds, while exactly `endtime+0.5*dt` and any later time reject because they round to `npts` or greater.
- CoreTimeSeries and CoreSeismogram overloads throw `MsPASSError` with `ErrorSeverity::Invalid` before constructing a child on rejection.
- TimeSeries and Seismogram history/logging overloads return a copy marked dead with `npts==0`, preserve metadata and processing history, and append exactly one `Invalid` elog entry.

## Required tests
- Exercise every Core and history/logging overload at the final sample, one representable value below the positive half-sample tie, the exact tie, and final sample plus `dt`; assert data/state/history/elog and absence of fabricated samples.

## Related issues
#551/#557 concern short-segment policy, not this index boundary.

## Re-audit disposition

The acceptance criteria restored independent endpoint rounding that the source history explicitly records as previously rejected for unequal sub-sample endpoint residuals. PR #896 was therefore closed. This issue is closed as not planned; a future narrow fix should preserve duration-derived sample count and correct only the out-of-range `ie >= npts` checks, with unequal-residual regression coverage.

wangyinz:
Fixes #847

## Summary

- map both requested endpoints through the waveform `sample_number` rule
- reject any end index at or beyond `npts`, including the exact positive half-sample tie
- make Core overloads throw `MsPASSError(Invalid)` before child construction
- make history/logging overloads return a dead zero-sample copy while preserving metadata/history and appending exactly one `Invalid` elog entry
- add Release-safe native coverage for all four overloads plus Python binding parity

## Validation

- Release focused CTest: passed
- rebuilt utility/seismic/basic bindings, confirmed loaded from the isolated worktree: new binding tests 4/4 passed
- baseline-red proof: old binding fabricated a live sixth sample at the half tie
- independent full-state probe confirmed parent immutability and returned metadata/history/elog contracts
- existing window module: 5 passed, 1 unrelated baseline-identical merge-order failure
- clang-format, Black, `py_compile`, and whitespace/diff checks: passed

## Independent agent review

A different agent re-read #847, rebuilt Release and three bindings, checked every overload and edge, independently reproduced the original fabricated sample, and returned **REVIEW PASS** with no blocker. No merge was performed.


wangyinz:
Closing this implementation after the renewed sampling-contract review.

The old source contains an explicit maintenance note explaining why end index is derived from the requested duration (outns = round((end-start)/dt)+1; ie=is+outns-1): independently rounding start and end had already been found to differ by one sample when their sub-sample residues differ. This PR deletes that warning and restores the discarded independent-end-rounding algorithm. Its tests cover final/half-tie boundaries but not differing fractional start/end residues, so they cannot establish compatibility.

The narrow defect in Issue #847 is still valid: bounds must reject ie >= npts so no fabricated terminal zero sample is returned. Keep the duration-derived endpoint, fix the bounds in all overloads, and add a non-congruent fractional-residue regression in a replacement PR.

PR #897, opened 2026-08-12
Post ensemble metadata from the first live member

### LINKED ISSUE #865: Post ensemble metadata from the first live member
## Audit baseline
`origin/master@61b06b965265a1327b84a8c6b6d7de491d9db97e`; findings PY-28.

## Problem
Default ensemble metadata posting iterates through all live members and leaves values from the last one.

## Impact
The fast path violates its first-live contract and makes output depend on input length.

## Acceptance criteria
- With `check_all_members=False`, copy each requested key from the first live member only; if that member lacks a key, raise `MsPASSError(full_message, ErrorSeverity.Invalid)` before ensemble metadata or member cleanup changes.
- If there is no live member, return the same ensemble with metadata and members unchanged.
- With `check_all_members=True`, require every live member to contain each key and compare values with `==`; any absence or mismatch raises the same exception before changes.  On success copy the common values, then erase member copies only when `clean_members=True`.

## Required tests
- Use first/last live members with different values, leading dead members, all-dead, missing keys, equal/mismatched all-member mode, and clean true/false; assert pre-error immutability and exact metadata source.

## Related issues

wangyinz:
Fixes #865

## Summary

- select the first live member once for the fast metadata-posting path
- return an all-dead ensemble unchanged
- prevalidate every requested key and every live-member equality before updating ensemble metadata or cleaning members
- use `ErrorSeverity.Invalid` consistently and remove the obsolete comparison helper
- clean member copies only after successful all-member validation and only when requested

## Validation

- focused TimeSeriesEnsemble/SeismogramEnsemble contract matrix: 16/16 passed
- existing converter regressions: 12/12 passed
- baseline-red proof with unmodified `master`: 6 focused failures for first-live/all-dead/atomicity behavior
- Black, `py_compile`, and `git diff --check`: passed

## Independent agent review

A different agent re-read #865, expanded the tests to both ensemble types and second-key failures, audited all write/error/cleanup ordering, reran focused/legacy/baseline tests, and returned **REVIEW PASS** with no blocker. No merge was performed.

### DIFF
--- a/python/mspasspy/util/converter.py
+++ b/python/mspasspy/util/converter.py
@@ -843,42 +843,6 @@ def Stream2SeismogramEnsemble(stream):
 obspy.core.Stream.toSeismogramEnsemble = Stream2SeismogramEnsemble
 
 
-def _all_members_match(ens, key):
-    """
-    This is a helper function for below.  I scans ens to assure all members
-    of the ensemble have the same value for the requested key. It uses
-    the python operator == for testing.  That can fail for a variety of
-    reasons the "match" may be overly restrictive for some types of data
-    linked to key.
-
-    :param ens:  ensemble data to scan.  Function will throw a MsPASS error if
-      the data this symbol is associated with is not a mspass ensemble object.
-    :param key:  key whose values are to be tested for all members of ens.
-    :return:  True of all members match, false if there are any differences.
-      Note if a key is not defined in a live member the result will be false.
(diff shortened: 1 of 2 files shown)


wangyinz:
Final maintainer review: an all-dead input remains unchanged and returns `None`, preserving the function's established in-place return contract. Live input still copies from the first live member, and validation/cleanup remain atomic.

PR #899, opened 2026-08-12
Reject misaligned grids in waveform arithmetic

### LINKED ISSUE #848: Reject misaligned grids in waveform arithmetic
## Audit baseline
`origin/master@61b06b965265a1327b84a8c6b6d7de491d9db97e`; findings CXX-17.

## Problem
TimeSeries and Seismogram `+=`/`-=` pair array indices after checking only time-reference type.

## Impact
Samples at different physical times can be silently combined.

## Acceptance criteria
- Treat sample intervals as equal only when `abs(dt1-dt2) <= 1e-6*max(abs(dt1),abs(dt2))`; otherwise throw `MsPASSError` with `ErrorSeverity::Invalid` before mutation.
- With equal intervals, let `q=(t02-t01)/dt1` and `k=round(q)` using C++ rounding; require `abs(q-k)<=1e-6`, otherwise throw the same error before mutation.
- For a valid integer offset, add or subtract only samples whose physical-time indices overlap; leave nonoverlap samples of the left operand unchanged.  Do not resample either operand.

## Required tests
- For both classes and both operators cover equal grids, positive/negative integer offsets, no overlap, offsets at and just beyond `1e-6` sample, and `dt` differences at and just beyond the relative tolerance; assert exact left-object state after every rejection.

## Related issues

wangyinz:
Closes #848

## Summary
- Reject incompatible time references, sample intervals, and non-integer grid offsets before waveform arithmetic mutates either operand.
- Apply arithmetic only on the physical overlap; never resample inside `TimeSeries` or `Seismogram` operators.
- In MCXcor's explicit alignment path, move each shifted start time to the nearest beam-grid sample without changing samples, `dt`, or `npts`.

## Maintainer decision
Each `TimeSeries` remains a dense uniform series. Different sampling intervals must be resampled explicitly. MCXcor uses whole-trace nearest-grid placement as its default alignment policy; general waveform arithmetic remains strict. No interpolation is performed implicitly.

## Validation
- Focused Python arithmetic and complete MCXcor regressions: 21 passed
- Full native CTest: 17 passed
- Black, Python byte compilation, and `git diff --check`: passed

### DIFF
--- a/cxx/include/mspass/seismic/CoreSeismogram.h
+++ b/cxx/include/mspass/seismic/CoreSeismogram.h
@@ -223,8 +223,8 @@ public:
   the ends of the rhs are inside the range of the lhs.
 
   \param d is other signal to add to this.
-  \exception MsPASSError can be thrown if lhs and rhs do not have matching
-  time standards.
+  \exception MsPASSError if lhs and rhs have incompatible time standards,
+  sample intervals, or start-time grids.
   **/
   CoreSeismogram &operator+=(const CoreSeismogram &d);
   /*! Addition operator.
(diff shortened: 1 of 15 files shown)


wangyinz:

An MCXcor owner must choose how strict sampled-grid arithmetic handles the existing 0.312835-sample output-window mismatch: documented resampling, nearest-index accumulation with a justified tolerance, or rejection. Do not restore unconditional time-grid snapping; this PR stays draft until that algorithm decision is made.

PR #902, opened 2026-08-12
Record cumulative calibration state and forward ensemble options

### LINKED ISSUE #869: Record cumulative calibration conversion and forward ensemble options
## Audit baseline
`origin/master@61b06b965265a1327b84a8c6b6d7de491d9db97e`; findings PY-33, PY-34.

## Problem
Calibration can scale samples without recording the cumulative applied conversion factor, and ensemble recursion drops caller matching/error options.

## Impact
Metadata can disagree with transformed samples and ensemble behavior can differ from direct atomic calls.

## Acceptance criteria
- On a successful cache match, multiply samples exactly once by `this_calib`.  Write `calib=old_calib*this_calib` when a finite numeric old value exists, and `calib=this_calib` when it is absent; `calib` records the cumulative conversion factor already applied to the samples.
- A present nonnumeric or nonfinite old `calib` appends one `Invalid` elog entry, kills the datum, and leaves samples and the old calibration metadata unchanged.
- Ensemble recursion forwards the caller's `id_key` and `kill_if_undefined` unchanged to every member; each member's outcome equals a direct atomic call with those values.

## Required tests
- Use missing, unit, nonunit, string, NaN, and Inf old calibration; assert sample multiplication and exact metadata product or unchanged invalid state; cross custom `id_key` and cache miss with both kill options for atomic/ensemble parity.

## Related issues

## Re-audit disposition

The cumulative calibration rule conflicts with the repository invariant `physical amplitude = calib * data` and can apply the same conversion twice. PR #902 was therefore closed. This issue is closed as not planned; a replacement must first define whether `calib` is a pending conversion or historical provenance, with applied-factor history stored separately.

wangyinz:
Fixes #869.

## Summary
- record successful calibration as the cumulative conversion factor already applied to the samples
- reject nonnumeric or nonfinite existing calibration metadata atomically with one Invalid elog entry
- forward custom identifier and undefined-calibration policy through ensemble recursion
- add direct atomic/ensemble parity regression coverage

## Verification
- focused public-path calibration tests: 12 passed
- master-baseline proof: 10 failures and 2 passes
- independent agent review: PASS, zero blockers
- local Mongo-backed legacy test was not claimed because no Mongo listener was available; GitHub CI remains required

This PR is ready for human review. It must not be merged automatically.

### DIFF
--- a/python/mspasspy/algorithms/calib.py
+++ b/python/mspasspy/algorithms/calib.py
@@ -1,4 +1,6 @@
 import pickle
+import math
+import numbers
 import numpy as np
 from mspasspy.ccore.seismic import TimeSeries, TimeSeriesEnsemble
 from mspasspy.ccore.utility import MsPASSError, ErrorSeverity
@@ -186,9 +188,10 @@ class ApplyCalibEngine:
         loaded in the cache by the constructor.   By default the key
         used is "channel_id".  That can be changed with the "id_key"
         argument.  If a match is found all
-        the sample valued are multiplied by calib AND the calib attribute is
-        set.   Note a complaint will be issued if calib was found to already
-        be defined in  AND is not 1.0 (a default sometimes appropriate)
+        the sample values are multiplied by calib AND the calib attribute is
+        set to the cumulative conversion factor applied to the samples.  Note
+        a complaint will be issued if calib was found to already be defined
+        AND is not 1.0 (a default sometimes appropriate).
 
         This method can easily become a mass murderer.   By default any
         datum with an undefined calib value in the cache of this object will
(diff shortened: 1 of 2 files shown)


wangyinz:
Closing this implementation after the renewed dimensional/API review.

The repository-wide public invariant is that physical amplitude is recovered as calib * data: the scale API explicitly preserves that product by dividing calib when samples are multiplied, and the database schema defines calib as the conversion factor from stored sample data to ground-motion units. ApplyCalibEngine itself says it converts counts to nm/s.

This PR multiplies samples by this_calib and then stores old_calib * this_calib as calib (or this_calib when absent). A downstream consumer following the public invariant applies the same response factor again; with no prior calib, the recovered physical amplitude is proportional to this_calib squared. The new tests encode the issue-authored cumulative-factor rule but do not supply a dimensional or round-trip oracle. Although the baseline function is internally inconsistent too, extending that inconsistency is not a safe fix.

Issue #869 should remain open. A replacement needs a domain-owner decision: if calib remains the pending sample-to-physical conversion, applying it to samples should normally leave calib normalized to 1 (with any applied-factor provenance stored separately). If calib is to be redefined as processing history, that requires a schema/API migration and coordinated changes to scale and all calib consumers. The ensemble option-forwarding and invalid-metadata atomicity fixes can be preserved independently.

PR #904, opened 2026-08-12
Close ExtractComponent input and output type contracts

### LINKED ISSUE #332: Problem with ExtractComponent and implications for ensemble handling

wangyinz:
Fixes #332.

## Summary
- reject TimeSeries, TimeSeriesEnsemble, and non-MsPASS inputs with `TypeError` before the generic wrapper can swallow the failure
- keep live/dead Seismogram output typed as TimeSeries
- preserve SeismogramEnsemble member count, order, metadata, ensemble state, and per-member state, including dead containers
- eliminate every implicit-`None` path

## Verification
- new contracts plus complete existing `test_basic.py`: 15 passed
- audited master baseline: 4 failures and 3 passes on the new contracts
- unsupported-input tests prove data, metadata, ordering, states, and elogs remain unchanged
- independent agent review: PASS, zero blockers

This PR is ready for human review. It must not be merged automatically.

### DIFF
--- a/python/mspasspy/algorithms/basic.py
+++ b/python/mspasspy/algorithms/basic.py
@@ -41,6 +41,21 @@ def _validate_lqt_angle_pair(func):
     return guarded
 
 
+def _reject_unsupported_extract_component_input(func):
+    """Reject unsupported inputs before the generic wrapper can return them."""
+
+    @wraps(func)
+    def checked(data, *args, **kwargs):
+        if not isinstance(data, (Seismogram, SeismogramEnsemble)):
+            raise TypeError(
+                "ExtractComponent only accepts Seismogram or SeismogramEnsemble"
+            )
+        return func(data, *args, **kwargs)
+
+    return checked
+
+
+@_reject_unsupported_extract_component_input
 @mspass_func_wrapper
 def ExtractComponent(
     data,
--- a/python/tests/algorithms/test_extract_component_contracts.py
+++ b/python/tests/algorithms/test_extract_component_contracts.py
@@ -1,0 +1,142 @@
+import numpy as np
+import pytest
+
+from mspasspy.algorithms.basic import ExtractComponent
+from mspasspy.ccore.seismic import (
+    Seismogram,
+    SeismogramEnsemble,
+    TimeSeries,
+    TimeSeriesEnsemble,
+)
+
+
+def _seismogram(marker, live=True):
+    result = Seismogram(4)
+    result["marker"] = marker
+    for component in range(3):
+        for sample in range(result.npts):
+            result.data[component, sample] = marker * 100 + component * 10 + sample
+    if live:
+        result.set_live()
+    else:
+        result.kill()
+    return result
+
+
+def _timeseries(marker, live=True):
+    result = TimeSeries(4)
+    result["marker"] = marker
+    for sample in range(result.npts):
+        result.data[sample] = marker * 100 + sample
+    if live:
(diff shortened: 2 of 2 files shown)


wangyinz:
Final maintainer review: a dead `SeismogramEnsemble` retains the existing short-circuit behavior and returns an empty dead `TimeSeriesEnsemble` without processing members. The PR now tightens unsupported-input `TypeError` behavior while preserving live-ensemble conversion.

PR #928, opened 2026-08-12
Allocate history job IDs atomically

### LINKED ISSUE #782: [P1][History] Allocate globally unique job IDs with one atomic Mongo counter
## Baseline

## Problem
Job IDs use a racy read-maximum-then-add-one sequence, and current tests do not exercise concurrent allocation.

## Impact
Concurrent workflows can receive the same supposedly unique job ID.

## Acceptance criteria
Allocate every automatic job ID with one Mongo `find_one_and_update` counter operation using `$inc`, `upsert`, and return-after semantics. Protect the counter document/key contract with a unique index and never fall back to scanning history. Keep this PR limited to job-ID allocation and tests.

## Required tests
Run concurrent real-Mongo allocation from multiple threads/processes; assert all IDs are unique and monotonic with no gaps caused by duplicate returns; verify counter initialization and unique-index enforcement.

## Related
Closed #61/#456 are broader design discussions, not this race.

wangyinz:
## Summary

- Allocate history job IDs from one atomic MongoDB counter.
- Keep legacy scanning out of normal allocation; provide an explicit, idempotent bootstrap for the historical high-water mark.
- Reserve explicit IDs and reject booleans/non-integers before database access.
- Add contract, concurrency, and real-Mongo coverage.

Closes #782

### DIFF
--- a/python/mspasspy/history.py
+++ b/python/mspasspy/history.py
@@ -34,6 +34,42 @@ def get_jobid(db):
     return counter["value"]
 
 
+def bootstrap_history_jobid_counter(db):
+    """Raise the job-id counter to the largest legacy history job id.
+
+    This is an explicit, idempotent deployment step for databases created
+    before job ids were allocated from ``history_counters``.  Runtime job-id
+    allocation does not call this function or scan the history collection.
+
+    :param db: database handle
+    :type db: top level database handle returned by a call to
+      MongoClient.database
+    :return: the resulting counter high-water mark
+    """
+    legacy_max_document = db.history.find_one(
+        {"jobid": {"$type": ["int", "long"]}},
+        projection={"_id": False, "jobid": True},
+        sort=[("jobid", pymongo.DESCENDING)],
+    )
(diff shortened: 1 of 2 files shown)


wangyinz:

A persistence owner must define how the atomic counter is bootstrapped above job IDs already stored by legacy deployments, including explicit-ID behavior. Fresh-database concurrency tests are insufficient; this PR stays draft until a migration or namespace policy guarantees global uniqueness.

wangyinz:
Review status: Draft. Required code fix: validate every explicit job ID as a non-bool integer before any Mongo operation; invalid BSON values can otherwise mutate and corrupt the numeric counter. Deployment decision: use an idempotent one-time bootstrap to the maximum legacy `history.jobid` before rollout, while the runtime allocator never scans history. Maintainer confirmation requested.

wangyinz:
Review decision: ready after final CI. Runtime allocation uses one atomic counter and never scans legacy history. Legacy high-water initialization is explicit and idempotent; explicit IDs reserve the counter, and invalid ID types fail before database access. The branch now contains only the three #782 commits rebased on current master.

PR #931, opened 2026-08-12
Support atomic SectionPlotter inputs with correct orientation

### LINKED ISSUE #789: Support atomic SectionPlotter inputs with correct orientation
## Baseline

Verified on `master`: `SectionPlotter.plot` advertises atomic `TimeSeries` and `Seismogram` inputs, but the scalar path passes a one-dimensional array to two-dimensional section renderers and the three-component path uses component-major data without orienting samples as rows.

## Problem

Atomic `TimeSeries` raises during plotting, while atomic `Seismogram` can be oriented incorrectly.

## Required behavior

- support atomic `TimeSeries` as one vertical section trace
- support atomic `Seismogram` as three vertical traces ordered component 0, 1, and 2 from left to right
- keep samples on matrix rows and traces/components on columns
- preserve the downward-running section time axis and all four existing styles: `wt`, `wtva`, `img`, and `wtvaimg`
- preserve ensemble behavior

## Required tests

Exercise both atomic types through all four public styles. Verify exact sample coordinates, one versus three traces, `npts x 1` versus `npts x 3` image orientation, and unchanged ensemble dispatch.

wangyinz:
## Summary

- support atomic `TimeSeries` as one vertical SectionPlotter trace
- orient atomic `Seismogram` data as `npts x 3`, with components 0, 1, and 2 from left to right
- preserve the downward-running time axis and all four existing styles
- keep ensemble dispatch unchanged
- document the atomic SectionPlotter behavior

## Validation

- raw-grid, atomic SectionPlotter, and adjacent graphics tests: 67 passed
- both atomic types exercised through `wt`, `wtva`, `img`, and `wtvaimg`
- `git diff --check` and Python compilation: passed
- Black reports the changed Python files unchanged

Closes #789

### DIFF
--- a/docs/source/user_manual/graphics.rst
+++ b/docs/source/user_manual/graphics.rst
@@ -215,23 +215,20 @@ types differs in a few important ways:
 2.  A :py:class:`Seismogram <mspasspy.ccore.seismic.Seismogram>` is displayed
     by ``SeismicPlotter`` in one figure, with components 0, 1, and 2 at equal
     vertical intervals from bottom to top.
-3.  :py:class:`TimeSeriesEnsemble <mspasspy.ccore.seismic.TimeSeriesEnsemble>`
+3.  :py:class:`SectionPlotter <mspasspy.graphics.SectionPlotter>` uses a
+    downward-running vertical time axis.  It displays an atomic ``TimeSeries``
+    as one trace and an atomic ``Seismogram`` as three traces, with components
+    0, 1, and 2 ordered from left to right.
+4.  :py:class:`TimeSeriesEnsemble <mspasspy.ccore.seismic.TimeSeriesEnsemble>`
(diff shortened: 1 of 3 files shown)


wangyinz:

A graphics API owner must decide whether atomic waveform plotting is repaired or formally deprecated with transition guidance. The current PR removes an advertised capability instead of repairing it, so it remains draft pending that API decision.

wangyinz:
Review resolution: atomic inputs are supported rather than rejected. TimeSeries is plotted as one trace; Seismogram data is transposed so components 0/1/2 become three left-to-right traces, with samples on rows. Existing ensemble dispatch is unchanged. The rebased implementation passes all 67 raw-graphics, SectionPlotter, and adjacent graphics tests locally.

PR #932, opened 2026-08-12
Correct graphics coordinates and allocation guards

### LINKED ISSUE #787: [P1][Graphics] Correct raw sample coordinates, bounds, normalization, and allocation guards
## Baseline

## Problem
Raw/atomic axes end one sample late; `wtva_raw` fixes a time bound at zero; constant normalization divides by zero; empty ensemble conversion divides by `dt=0`; allocation guards bound rows but not total array size.

## Impact
Plots mislabel/clip data, emit NaN/Inf, or attempt unsafe allocations.

## Acceptance criteria
Coordinate sample `i` as `t0+i*dt`, so the last coordinate is `t0+(npts-1)*dt`, and use the actual first/last coordinates for both wiggle and image bounds. For an all-constant normalized section, emit finite zero amplitude offsets for every sample. Empty input or a non-finite/non-positive `dt` raises `ValueError`. Before allocation, compute `rows*columns*itemsize` with checked integer multiplication; multiplication overflow or a result greater than 536870912 bytes (512 MiB) raises `MemoryError`, while exactly 536870912 bytes is allowed. Limit this PR to raw conversion/plotting and tests.

## Required tests
Assert every line/image coordinate and bound for positive and negative nonzero `t0`; assert all-constant traces produce only finite zero offsets. Cover empty, NaN, infinite, zero, and negative `dt`. Exercise allocation at 536870912 bytes, one byte above it, and dimensions whose multiplication would overflow, asserting allocation is never attempted for rejected cases.

## Related
Closed #65 is general plotting design, not these defects.

wangyinz:
## Summary

- coordinate raw and atomic plots on the exact `t0 + i*dt` sample grid
- use the true first and final sample for line and image bounds
- normalize constant sections to finite zero offsets
- reject non-finite/non-positive sample intervals before plotting or allocation
- check matrix dimension multiplication against the platform index size before allocation
- preserve the established ten-million-row absolute-time-span sanity policy

## Approved API policy

This PR does not add the fixed 512 MiB limit proposed in the original Issue. It preserves existing public empty-input behavior and the existing 10,000,000-row ensemble time-span guard, while adding platform integer-overflow checks before allocation. Ordinary resource exhaustion remains NumPy/the host's responsibility.

## Validation

- raw, atomic, allocation, and adjacent graphics tests: 51 passed
- `git diff --check`: passed
- Black reports the changed files unchanged

Closes #787

### DIFF
--- a/python/mspasspy/graphics.py
+++ b/python/mspasspy/graphics.py
@@ -16,6 +16,42 @@ from mspasspy.algorithms.basic import ExtractComponent
 # not sure that is necessary but this makes context clearer
 from mspasspy.algorithms.window import scale as mspass_scale_function
 
+_MAX_PLOT_ARRAY_BYTES = 536870912
+
+
+def _validate_plot_grid(npts, dt):
+    if npts <= 0:
+        raise ValueError("plot input is empty")
+    if not numpy.isfinite(dt) or dt <= 0.0:
+        raise ValueError("plot sample interval must be finite and positive")
+
+
+def _checked_plot_allocation(rows, columns, itemsize):
+    max_index = numpy.iinfo(numpy.intp).max
+    if rows < 0 or columns < 0 or itemsize <= 0:
+        raise ValueError("plot array dimensions and item size must be positive")
+    if rows and columns > max_index // rows:
+        raise MemoryError("plot array dimensions overflow the platform index size")
+    elements = rows * columns
+    if elements and itemsize > max_index // elements:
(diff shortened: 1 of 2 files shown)


wangyinz:
Maintainer decision applied: preserve existing empty-input behavior and the 10,000,000-row ensemble time-span guard; do not add a fixed 512 MiB allocation limit. Platform integer-overflow checks remain.

PR #949, opened 2026-08-12
Harden AWS Lambda invocation without breaking response compatibility

### LINKED ISSUE #806: [P1][AWS example] Enforce one region-aware, size-safe Invoke response protocol
## Baseline

## Problem
The AWS Lambda example masks some invocation failures, can use the wrong region, estimates the response limit from raw output rather than the serialized response, and does not reliably clean up streams and temporary files.

## Compatibility constraint
Preserve the response formats already consumed by MsPASS and example users:

- Lambda wire response: `ret_type="content"` carries base64 in `ret_value`; `ret_type="key"` carries `bucket::key` in `ret_value`.
- `AwsLambdaClient.call_lambda_function` returns the existing public fields `return_type` and `ret_value`.
- Additional wire-response fields remain forward-compatible and do not invalidate an otherwise valid response.

## Acceptance criteria
Use the configured boto3 Session/client region and remove the hardcoded region. Treat Lambda `FunctionError`, malformed UTF-8/JSON, invalid base64, malformed required fields, or an unknown response type as contextual `RuntimeError`, never success. Decide whether content fits from the final UTF-8 JSON response using serialization aligned with the Lambda Python runtime; final sizes of 5,999,999 and 6,000,000 bytes remain inline, while 6,000,001 bytes uploads the output and returns `bucket::key`. Close every response/archive stream and remove every owned temporary file in `finally` on success or failure. Limit this issue to the AWS Invoke example/protocol and tests.

## Required tests
Cover both response types and an alternate configured region; FunctionError; malformed UTF-8/JSON; missing/wrong fields; invalid base64; unknown type; an additional response field; exact final serialized size boundaries; and download, processing, serialization, upload, and client-decode cleanup failures. Include regression assertions for the existing `bucket::key` wire value and public `return_type` field.

## Related
Runtime packaging is isolated in PLAT-AWS-RUNTIME / #803.

wangyinz:
Fixes #806.

## Summary
- preserve the existing Lambda wire protocol (`content` or `bucket::key`) and the public `return_type` client field
- keep otherwise valid responses forward-compatible when they contain additional fields
- use the configured AWS region, surface `FunctionError` and malformed payloads as contextual `RuntimeError`, and close response/archive streams
- size inline results from the final runtime-style UTF-8 JSON response and upload larger results to S3
- clean up every owned temporary file on success and failure; explicit S3 output skips unnecessary file reads and base64 serialization
- regenerate the deterministic Python 3.13 deployment bundle with the updated handler

## Verification
- AWS Invoke protocol tests: 32 passed
- Python 3.13 bundle/runtime tests: 4 passed
- bundle imports/native ABI and ObsPy miniSEED round trip passed
- embedded handler sources match the repository sources byte-for-byte

The PR remains draft only while the refreshed GitHub checks run.

### DIFF
--- a/python/tests/test_aws_lambda_invoke_protocol.py
+++ b/python/tests/test_aws_lambda_invoke_protocol.py
@@ -1,0 +1,562 @@
+import base64
+import importlib.util
+import io
+import json
+import os
+import sys
+import tempfile
+import zipfile
+from pathlib import Path
+
+import pytest
+from botocore.response import StreamingBody
+from botocore.stub import Stubber
+
+REPOSITORY_ROOT = Path(
+    os.environ.get("MSPASS_TEST_REPOSITORY_ROOT", Path(__file__).resolve().parents[2])
+)
+EXAMPLE_DIR = REPOSITORY_ROOT / "scripts" / "aws_lambda_examples"
+
+
+def _load_module(name, filename):
+    spec = importlib.util.spec_from_file_location(name, EXAMPLE_DIR / filename)
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module
+
+
+@pytest.fixture
+def client_module():
+    return _load_module("mspass_aws_lambda_client_test", "AwsLambdaClient.py")
+
+
+@pytest.fixture
+def process_module(monkeypatch):
+    function_module = _load_module("aws_lambda_func_def", "aws_lambda_func_def.py")
+    monkeypatch.setitem(sys.modules, "aws_lambda_func_def", function_module)
+    return _load_module("mspass_aws_lambda_process_test", "process.py")
+
+
+class TrackingStream(io.BytesIO):
+    def __init__(self, value):
+        super().__init__(value)
+        self.was_closed = False
+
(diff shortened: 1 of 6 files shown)


wangyinz:

A Lambda API owner must define the wire contract: field names, S3 reference shape, unknown-field behavior, and the exact serialized-size boundary near 6 MB. Keep this PR draft until runtime interoperability evidence validates that contract.

wangyinz:
Compatibility blocker confirmed against current repository callers:

- `Database._download_windowed_mseed_file` still parses key responses as `bucket::key`; this PR returns only the key.
- `AwsLambdaClient.call_lambda_function` currently exposes `return_type`; this PR changes that public result field to `ret_type`.
- The 5,999,999/6,000,000/6,000,001-byte tests measure compact JSON generated inside `process.py`, but the AWS Python runtime serializes the returned object with its own encoder and separators. They therefore do not prove the transmitted Invoke boundary.

Recommendation: merge #950 first, then rebase this PR; preserve the existing wire and caller-facing shapes, retain the region/`FunctionError`/cleanup fixes, and use a runtime-aligned or conservative inline-size test. Keep this PR draft pending that compatibility decision.

Issue #807, opened 2026-08-12
[P2][IU launcher] Repair HPC host, command, readiness, and process-state contracts


wangyinz:
## Baseline
Verified on `master` commit `61b06b965265a1327b84a8c6b6d7de491d9db97e`.

## Problem
All-explicit hosts leave attributes unset; scheduler assignment is misspelled; tunnel executes stale argv; scheduler/worker endpoints are wrong or incomplete; remote env lacks a delimiter and contains literal `&`; status treats exit 0 as running and polls absent workers; frontend omits endpoints and blocks on `communicate`.

## Impact
Common HPC configurations fail, launch against wrong services, misreport stopped processes, or block interactive use.

## Acceptance criteria
Initialize `primary_node`, `database_host`, `scheduler_host`, `worker_hosts`, and every process handle for all-explicit, all-auto, and mixed host configurations. Execute the newly constructed tunnel/container/worker argv exactly, pass configured Mongo and Dask endpoints in complete comma-delimited environment values, and use neither literal `&` nor `shell=True`. Only `process.poll() is None` is running; skip absent optional worker/frontend handles without polling or treating them as failures. Use `MSPASS_STARTUP_TIMEOUT_SECONDS` default 120 and `MSPASS_STARTUP_POLL_SECONDS` default 2. Before workers or frontend, require Mongo `ping` on the configured endpoint and a Dask probe that creates `distributed.Client(configured_endpoint, timeout='2s')`, calls `scheduler_info()`, and closes it. On owned-child early exit or timeout, start no worker/frontend, terminate and wait for all already-started owned children, and raise `MsPASSError` with `ErrorSeverity.Invalid`. Interactive launch includes both endpoints, returns immediately without `communicate()`, and retains its process handle. Limit this PR to `HPCClusterLauncher` and tests.

## Required tests
Mock exact argv/environment for all-explicit, all-auto, mixed, tunnel, primary-worker, remote-worker, batch, and interactive paths, asserting no shell metacharacter or `shell=True`. Cover status for `None`, 0, nonzero, and absent optional handles; default/overridden timeout/poll settings; delayed Mongo/Dask readiness; each child early exit and timeout; Dask close; no later launch; owned-child terminate/wait; exact Invalid exception; nonblocking frontend and retained handle; YAML parse and `py_compile`.

## Related
Merged #640 introduced the prototype but does not repair it.

wangyinz:
[...]

PR #951, opened 2026-08-12
Fix IU HPC launcher lifecycle

wangyinz:
Related to #807

## Summary

- initialize explicit/auto/mixed host and every process-handle state
- execute exact tunnel/container/worker argv without shell metacharacters or `shell=True`
- probe the configured MongoDB and Dask endpoints before workers/frontends
- detect every owned-child early exit and startup timeout
- terminate and wait for already-started owned children before reporting failure
- retain a failed process handle when cleanup itself fails so shutdown can be retried
- keep frontend launch nonblocking and retain its process handle

## Renewed error-contract review

The issue text changed all startup failures to `MsPASSError(Invalid)`, but this class is an infrastructure launcher and its baseline public behavior uses ordinary Python runtime/configuration exceptions.  Commit `60be0795` restores that boundary:

- malformed timeout/poll settings raise `ValueError` before any child starts
- owned-child early exit/readiness timeout raises `RuntimeError`
- process-creation/system exceptions are re-raised with their original identity after cleanup
- if cleanup fails, that failure is explicit and the failed handle is retained rather than silently discarded

The long `launch` and `shutdown` docstrings now document these boundaries.

## Validation

- focused launcher lifecycle contract: 29/29 passed
- added original-exception identity and cleanup-retry-handle tests
- Python 3.12/3.13 `py_compile` and `git diff --check`: passed
- the new test file is Black 26.5.1 clean; the pre-existing launcher file has legacy whole-file formatting, so unrelated reformatting was deliberately excluded

This remains draft pending fresh CI and human infrastructure/API review.

### DIFF
--- a/python/tests/test_iu_hpc_launcher_contract.py
+++ b/python/tests/test_iu_hpc_launcher_contract.py
@@ -1,0 +1,693 @@
+import importlib.util
+import os
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+import yaml
+
+REPOSITORY_ROOT = Path(
+    os.environ.get("MSPASS_TEST_REPOSITORY_ROOT", Path(__file__).resolve().parents[2])
+)
+LAUNCHER_PATH = REPOSITORY_ROOT / "scripts" / "IU_examples" / "python" / "launcher.py"
+
+
+@pytest.fixture
+def launcher_module():
+    spec = importlib.util.spec_from_file_location(
+        "mspass_iu_launcher_test", LAUNCHER_PATH
+    )
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    assert Path(module.__file__).resolve() == LAUNCHER_PATH.resolve()
+    return module
+
+
+def _configuration(**cluster_updates):
+    cluster = {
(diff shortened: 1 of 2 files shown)


wangyinz:
Renewed review found three blockers before merge:

1. Drop the `test_workflow.py` commit; that lifecycle fix was already superseded by #1004/#885 on `master`.
2. Normalize an auto-discovered Dask host to the same complete endpoint used by workers (default `tcp://HOST:8786`) before the readiness probe. A bare host can hang beyond the advertised 2-second boundary.
3. Do not pipe long-lived child stdout/stderr unless they are drained. Inheriting the batch job output is the smallest safe default.

Rebase and merge this PR before #952 because both edit `launcher.py`. No maintainer API decision remains for #951.

wangyinz:
Implemented the renewed review fixes on current `master`:

- dropped the superseded `test_workflow.py` lifecycle commit
- normalized auto Dask readiness addresses to `tcp://HOST:8786` (or `DASK_SCHEDULER_PORT`)
- removed undrained pipes from long-lived owned children
- parsed configured runtime arguments with `shlex`

Final local validation: 30 focused launcher tests, `py_compile`, and `git diff --check` passed. The PR remains Draft until this final head passes CI.

PR #952, opened 2026-08-12
Repair DesktopLauncher configuration and Compose lifecycle

### LINKED ISSUE #808: [P2][IU launcher] Repair DesktopLauncher configuration and Compose lifecycle
## Baseline

## Problem
DesktopLauncher queries the Jupyter URL before readiness, leaves browser command undefined for common OS names, uses inconsistent frontend service/config names, ignores Compose return codes, and does not implement the base run/status contract.

## Impact
Startup can silently fail, open no/incorrect URL, or raise platform-dependent errors; lifecycle state is unreliable.

## Acceptance criteria
Use exactly `data/yaml/compose.yaml` as the default Compose file and `mspass-frontend` as the service in up/logs/ps/exec/down operations. Accept canonical `platform.system()` names `Linux`, `Darwin`, and `Windows`; launch browser argv as `[browser, url]`, `["open", "-a", browser, url]`, and `["cmd", "/c", "start", "", browser, url]` respectively, without `shell=True`. Any other OS raises `MsPASSError` with `ErrorSeverity.Invalid` before Compose starts. After successful `docker compose up -d`, poll service logs for a real Jupyter HTTP(S) URL using `MSPASS_STARTUP_TIMEOUT_SECONDS` default 120 and `MSPASS_STARTUP_POLL_SECONDS` default 2; do not return a tokenless fallback URL. A nonzero up/logs/ps/exec/down result, early service exit, URL timeout, or browser-launch failure raises Invalid. If this launcher started the stack, every post-up failure runs `docker compose down`, terminates/waits any owned browser process, and clears ownership. `launch()` is idempotent and returns the URL; `status()` returns 1 only when `compose ps --status running --services` contains `mspass-frontend`, otherwise 0; `run(python_file)` executes `docker compose exec -T mspass-frontend python python_file`, returns its `CompletedProcess` on zero, and raises Invalid on nonzero; `shutdown()` is repeatable and downs only its owned stack. Limit this PR to `DesktopLauncher` and tests.

## Required tests
Mock exact Compose and browser argv for Linux, Darwin, Windows, and unsupported OS; exact config/service consistency; delayed valid URL, early service exit, timeout, and no fallback; default/overridden timing; every Compose/browser failure; owned browser/stack cleanup; caller-owned stack preservation; repeated launch/shutdown; exact status/run return and exception contracts; base method signatures and `py_compile`.

## Related
Merged #640 introduced the prototype only.

wangyinz:
## Summary

- standardize `DesktopLauncher` on `data/yaml/compose.yaml`, canonical Linux/macOS/Windows browser argv, and argv-only subprocess execution
- start the complete Compose project rather than only `mspass-frontend`, and report ready only when every configured service is running
- wait for a real Jupyter HTTP(S) URL with configurable startup timing; never fabricate a tokenless fallback
- preserve caller-owned projects, make launch/shutdown idempotent, and deterministically terminate/wait (or kill) owned browser processes
- retain ordinary Python API boundaries: invalid arguments raise `ValueError`; Compose, readiness, and browser failures raise `RuntimeError`
- document topology, ownership, lifecycle, return values, and failure behavior in the public class and method docstrings

## Root cause

The prototype queried logs before readiness, mixed Compose configuration and service names, built platform-specific browser shell strings, ignored command return codes, and did not track stack/browser ownership consistently.

The first version of this PR followed an issue-authored contract too literally: it targeted and checked only the frontend service, which can leave a scheduler with no worker, and changed ordinary launcher failures to `MsPASSError(Invalid)` without a compatibility basis. The scientific/API re-audit corrected both expansions.

## Impact

Desktop launches now fail explicitly, require the database/scheduler/worker/frontend topology to be running, never fabricate a fallback URL, preserve caller-owned stacks, and clean up owned resources without replacing the original exception when cleanup itself also fails.

## Validation

- focused launcher contract: **24 passed**
- exact Compose/browser argv, delayed URL, full-service readiness, partial topology, every public failure path, ownership, idempotence, and browser poll cleanup covered
- Python byte compilation and `git diff --check`: passed
- test file formatted with Black; production changes are confined to `DesktopLauncher` in a legacy file whose unrelated HPC section is not Black-formatted

This PR remains draft pending the full renewed compatibility/scientific re-audit.

Fixes #808

### DIFF
--- a/python/tests/test_iu_desktop_launcher_contract.py
+++ b/python/tests/test_iu_desktop_launcher_contract.py
@@ -1,0 +1,571 @@
+import importlib.util
+import inspect
+import os
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+from mspasspy.ccore.utility import ErrorSeverity, MsPASSError
+
+REPOSITORY_ROOT = Path(
+    os.environ.get("MSPASS_TEST_REPOSITORY_ROOT", Path(__file__).resolve().parents[2])
+)
+LAUNCHER_PATH = REPOSITORY_ROOT / "scripts" / "IU_examples" / "python" / "launcher.py"
+
+
+@pytest.fixture
+def launcher_module():
+    spec = importlib.util.spec_from_file_location(
+        "mspass_iu_desktop_launcher_test", LAUNCHER_PATH
+    )
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module
+
+
+class BrowserProcess:
+    def __init__(self, status=None):
+        self.status = status
+        self.terminate_count = 0
+        self.wait_count = 0
+
+    def poll(self):
+        return self.status
+
+    def terminate(self):
+        self.terminate_count += 1
+        self.status = 0
+
+    def wait(self):
+        self.wait_count += 1
+        return self.status
+
+
+class PollFailureBrowser(BrowserProcess):
+    def poll(self):
+        raise OSError("browser poll failed")
+
+
+class SubprocessScript:
+    def __init__(self, results):
+        self.results = list(results)
+        self.run_calls = []
+        self.popen_calls = []
+        self.browser = BrowserProcess()
+        self.popen_error = None
+
+    def run(self, argv, **kwargs):
+        self.run_calls.append((argv, kwargs))
+        assert self.results, f"unexpected subprocess.run call: {argv}"
+        result = self.results.pop(0)
+        if isinstance(result, BaseException):
+            raise result
+        return result
+
+    def popen(self, argv, **kwargs):
+        self.popen_calls.append((argv, kwargs))
+        if self.popen_error is not None:
+            raise self.popen_error
+        return self.browser
+
+
+def completed(stdout="", stderr="", returncode=0):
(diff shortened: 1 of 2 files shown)


wangyinz:
Renewed review found three blockers:

1. Prefer the loopback Jupyter URL. Real logs print the container hostname URL before `127.0.0.1`; returning the first URL opens an address the desktop host cannot resolve.
2. Treat a nonzero `compose up` as potentially partial and attempt cleanup when this launcher began from an empty project.
3. Retain `_owns_stack` and the browser handle when cleanup fails so `shutdown()` can be retried.

Maintainer decision needed: if some configured Compose services are already running on entry, should `launch()` (a) fail without changing the caller-owned partial project (recommended), or (b) complete it with `up -d` but never run `down`?

wangyinz:
Implemented the reviewed lifecycle policy and rebased onto #951/master.

- An empty Compose project is launcher-owned and is cleaned up on launch failure or shutdown.
- A partially running project is caller-owned: `up -d` fills missing services, but this launcher never runs `down`.
- Failed `down` and browser cleanup retain state/handles so cleanup can be retried.
- Jupyter log parsing prefers a loopback URL when container and host URLs are both present.
- HPC and desktop launcher contract suites pass together: 57 tests.

Keeping this PR draft until the refreshed CI is green.

PR #953, opened 2026-08-12
Persist GridFS storage mode during waveform migration

### LINKED ISSUE #813: Persist storage_mode when update_data migrates file samples to GridFS
## Baseline
Verified against origin/master commit 61b06b965265a1327b84a8c6b6d7de491d9db97e on 2026-08-12. Findings: B03, T05-file-gridfs-transition.

## Problem
When update_data moves samples from file storage to GridFS, it updates the in-memory object and gridfs_id but does not persist storage_mode=gridfs. Locations: python/mspasspy/db/database.py:3202-3240.

## Impact
A later read follows the stale file reference, hides the update, and can leave an unreachable new GridFS object.

## Acceptance criteria
- A successful migration stores storage_mode="gridfs" and a gridfs_id that resolves to the new samples; the next public read returns those samples.
- If the waveform-document update fails, the original file reference remains unchanged and readable.
- A failed migration removes the newly created unreferenced GridFS object.
- Waveforms not taking this file-to-GridFS transition retain their current storage behavior.

## Required tests
- Successful file-to-GridFS round trip with exact storage_mode/gridfs_id and sample assertions.
- Injected waveform-update failure proving the original reference remains readable and no new orphan remains.

## Related
No direct current issue covers this transition.

wangyinz:
## Summary

- persist `storage_mode="gridfs"` in the waveform document when `update_data` migrates samples from file storage
- treat `storage_mode="file"` as authoritative even when an earlier failed migration left a stale `gridfs_id`
- stage new GridFS data without deleting any pre-existing object; if the waveform reference update fails or matches no document, remove only the new blob and restore the caller's exact storage keys
- preserve the original file-backed document and verify successful results through the public `read_data` path

## Root cause

`Database.update_data` changed the in-memory object and wrote new samples to GridFS, but did not persist `storage_mode`. The first repair recognized only pristine file documents without a `gridfs_id`; it therefore could not recover the mixed `storage_mode=file + stale gridfs_id` state produced by the original bug. Using the normal overwrite path for that state would also delete the pre-existing GridFS object before the waveform reference became durable.

## Validation

- MongoDB 8.0.29 contract suite: 12 passed across TimeSeries/Seismogram, pristine/stale file states, success/exception/unmatched-update paths
- contract suite plus neighboring `TestDatabase::test_update_data`: 13 passed
- exact baseline 61b06b96 against the final suite: 10 failed, 2 passed
- failure snapshots prove the waveform document, readable file samples, caller storage keys, and every pre-existing GridFS file/chunk remain unchanged while the newly staged blob is removed
- Black 26.5.1, Python 3.12/3.13 `py_compile`, and `git diff --check`: passed

This remains a draft pending the repository-wide scientific/API re-audit.

Fixes #813

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -3343,9 +3343,11 @@ class Database(pymongo.database.Database):
                 if "storage_mode" in mspass_object
                 else None
             )
-            file_to_gridfs_transition = (
-                original_storage_mode == "file" and "gridfs_id" not in mspass_object
+            original_has_gridfs_id = "gridfs_id" in mspass_object
+            original_gridfs_id = (
+                mspass_object["gridfs_id"] if original_has_gridfs_id else None
             )
+            file_to_gridfs_transition = original_storage_mode == "file"
             if "storage_mode" in mspass_object:
                 storage_mode = mspass_object["storage_mode"]
                 if not storage_mode == "gridfs":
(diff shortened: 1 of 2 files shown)


wangyinz:
Independent compatibility audit complete on the current #955/#1002 commit/rollback path. File storage remains authoritative beside a stale `gridfs_id`; success persists the new GridFS mode/reference without deleting that unrelated stale blob, and failure removes only the staged blob and restores the caller storage keys. Python 3.10/3.13, MongoDB, formatting, and Docker CI are fully green.

PR #954, opened 2026-08-12
Replace elog records in place

### LINKED ISSUE #816: Replace elog records without deleting the previous record first
## Baseline
Verified against origin/master commit 61b06b965265a1327b84a8c6b6d7de491d9db97e on 2026-08-12. Findings: B08, T08-elog-replacement.

## Problem
_save_elog deletes an existing elog document and then inserts its replacement. Locations: python/mspasspy/db/database.py:4241-4259.

## Impact
An insert failure loses the previous audit log while waveform metadata may still reference its old ID.

## Acceptance criteria
- Use the smallest existing PyMongo path: update the existing elog document in place by its current _id, without delete-then-insert.
- Preserve the elog _id and waveform reference for an existing record.
- If the update fails, the prior elog remains readable and unchanged.
- The no-existing-elog path continues to insert one new record.

## Required tests
- Successful in-place replacement with stable _id.
- Injected update failure proving the old elog remains readable.
- New-elog insertion behavior.

## Related
No direct issue covers this replacement ordering.

wangyinz:
## Summary

- replace an existing elog document atomically at its current `_id`
- preserve the existing waveform cross-reference when the in-memory datum no longer carries `_id`
- keep new-elog insertion behavior unchanged and update the public regression expectation to the stable-ID contract

## Root cause

`Database._save_elog` previously deleted the old document and inserted a replacement.  If insertion failed, the prior audit log was already gone and the waveform could still reference its obsolete ID.

## Validation

- MongoDB 8.0.29: 8 TimeSeries/Seismogram contract cases passed
- neighboring `test_save_elogs` and public `test_update_data`: passed; 10 focused tests total
- exact baseline: 4 contract failures and the updated public stable-ID assertion fails as expected
- independent agent review: passed
- Black, Python 3.12/3.13 `py_compile`, and diff checks: passed

Fixes #816

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -4405,7 +4405,8 @@ class Database(pymongo.database.Database):
     ):
         """
         Save error log for a data object. Data objects in MsPASS contain an error log object used to post any
-        errors handled by processing functions. This function will delete the old elog entry if `elog_id` is given.
+        errors handled by processing functions.  If `elog_id` identifies an
+        existing record, that record is replaced atomically in place.
 
         :param mspass_object: the target object.
         :type mspass_object: either :class:`mspasspy.ccore.seismic.TimeSeries` or :class:`mspasspy.ccore.seismic.Seismogram`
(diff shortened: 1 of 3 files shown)


wangyinz:
Independent final-head audit complete. Existing elog records keep the same `_id` and waveform cross-reference and are replaced in one database operation, eliminating the delete/insert gap; missing-id insertion behavior remains unchanged. Current-master CI is fully green.

PR #955, opened 2026-08-12
Make GridFS waveform replacement failure-safe

### LINKED ISSUE #814: Make GridFS overwrite preserve the previous readable sample
## Baseline
Verified against origin/master commit 61b06b965265a1327b84a8c6b6d7de491d9db97e on 2026-08-12. Findings: B04, T06-gridfs-overwrite.

## Problem
GridFS overwrite deletes the old blob before writing and committing the replacement. Locations: python/mspasspy/db/database.py:6664-6673 and its update_data caller.

## Impact
A failed put or reference update can permanently lose the only readable sample while the waveform document still points to it.

## Acceptance criteria
- Write the new GridFS object first, then atomically change the waveform reference from the expected old gridfs_id to the new gridfs_id.
- If the new write or reference change fails, retain the old reference and readable old blob, remove any unreferenced new blob, and rethrow the original exception.
- Delete the old blob only after the new reference is committed.
- If old-blob deletion fails, keep the committed new reference and readable new blob, do not roll back, retain the old blob, append one Complaint to the datum elog, and return the normally updated datum.

## Required tests
- Successful overwrite with the old blob removed and the returned datum referencing readable new samples.
- Injected new-put and reference-update failures asserting the old reference/blob, absence of a new orphan, and the original exception.
- Injected old-delete failure asserting the committed new reference, both blobs, one Complaint, and a normal updated-datum return.

## Related
Issue #570 concerned read mismatch handling, not overwrite ordering.

wangyinz:
Fixes #814.

## Summary

- Stage replacement samples under a new GridFS ID.
- Switch public `update_data` and `save_data(overwrite=True)` references with compare-and-set.
- Delete the old blob only after the new waveform reference is durable.
- Keep the low-level GridFS helper stage-only.
- Reject `write_distributed_data(overwrite=True)` before scheduler or storage execution because distributed bulk insert cannot safely replace an existing waveform reference.
- Restrict rollback handling to ordinary exceptions.

## Failure contract

A failed stage or waveform-reference update preserves the previous readable GridFS object. Cleanup targets only the newly staged object. Failure to delete an old object after commit is reported without reverting the committed new reference.

## Verification

- Black and Python compilation checks passed.
- Focused distributed-overwrite rejection tests passed locally.
- MongoDB-backed overwrite contracts remain covered by the PR test suite and CI.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -3169,6 +3169,11 @@ class Database(pymongo.database.Database):
         # only for usage errors.   We test the elog size to check if there
         # are other warning messages and add a summary if there were any
         logsize0 = mspass_object.elog.size()
+        old_gridfs_id = (
+            mspass_object["gridfs_id"] if "gridfs_id" in mspass_object else None
+        )
+        if old_gridfs_id is not None:
+            mspass_object.erase("gridfs_id")
         try:
             self.update_metadata(
                 mspass_object,
@@ -3179,8 +3184,9 @@ class Database(pymongo.database.Database):
                 normalizing_collections=normalizing_collections,
                 alg_name=alg_name,
             )
-        except:
-            raise
+        finally:
+            if old_gridfs_id is not None:
+                mspass_object["gridfs_id"] = old_gridfs_id
         logsize = mspass_object.elog.size()
         # A bit verbose, but we post this warning to make it clear the
         # problem originated from update_data - probably not really needed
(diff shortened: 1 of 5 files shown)


wangyinz:

Atomic and ensemble GridFS overwrite paths are repaired, but distributed bulk overwrite still lacks compare-and-set semantics for an existing waveform reference. Keep this PR draft until the distributed overwrite contract is designed and tested.

PR #962, opened 2026-08-13
Fix inventory location grouping and lookup

### LINKED ISSUE #824: Preserve Inventory channel locations and reject ambiguous lookups
## Baseline
Verified against origin/master commit 61b06b965265a1327b84a8c6b6d7de491d9db97e on 2026-08-12. Findings: B16, B17.

## Problem
save_inventory pairs every channel with every location and uses station-wide times. get_seed_channel can return implicit None for multiple explicit-location matches. Locations: python/mspasspy/db/database.py:5163-5295,5529-5536.

## Impact
Multi-location station metadata is corrupted and duplicate lookup behavior is ambiguous.

## Acceptance criteria
- The Inventory-to-channel grouping logic emits each channel only under its own location, with that location's coordinates and validity interval; it never forms a channel-by-location Cartesian product.
- An explicit location lookup returns None for zero matches and the record for one match. More than one match raises MsPASSError with ErrorSeverity.Invalid and never selects one implicitly.
- This PR changes only grouping and lookup logic. It does not define a persistent Inventory representation or implement public save/read round-trip; DATA-33 owns that codec and integration.

## Required tests
- Unit tests with a mocked codec boundary for two locations having different coordinates, validity periods, channels, and responses, including absence of cross-location channel records.
- Explicit-location zero, one, and multiple-match tests asserting None, the sole record, and MsPASSError/ErrorSeverity.Invalid respectively.
- No serialization, pickle migration, or public save/read round-trip tests in this Issue.

## Related
Issue #481 concerned coordinate order and GeoJSON, not this grouping and lookup failure. DATA-33 owns finding B15, T15-inventory, the Inventory codec, and public round-trip integration; this grouping/lookup PR remains independently mergeable by testing against a mocked codec boundary.

wangyinz:
## Summary

- Group each StationXML channel only under its own location code and preserve its exact coordinates, depth, validity interval, orientation, and response.
- Create one site per network/station/location with the union of its channel intervals. Use channel coordinates when they agree; otherwise use Station coordinates and emit one warning.
- Return `None` for zero explicit-location matches, the document for one match, and `MsPASSError(Invalid)` for ambiguous duplicates.

This PR does not define or migrate the persistent Inventory representation; #989 owns that codec and its backward-compatible readers.

## Validation

- Local contract result: 1 database-independent test passed; 4 MongoDB-backed tests were collected and skipped because no local MongoDB listener was available.
- Python compilation, Black formatting, and `git diff --check` pass.

Closes #824

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -5496,8 +5496,8 @@ class Database(pymongo.database.Database):
                     rec = geoJSON_doc(loc_lat, loc_lon, doc=rec, key="location")
                     rec["elev"] = loc_elev
                     rec["edepth"] = loc_edepth
-                    rec["starttime"] = starttime.timestamp
-                    rec["endtime"] = endtime.timestamp
+                    rec["starttime"] = loc_stime.timestamp
+                    rec["endtime"] = loc_etime.timestamp
                     if (
                         latitude != loc_lat
                         or longitude != loc_lon
@@ -5566,11 +5566,11 @@ class Database(pymongo.database.Database):
                     # done with site now handle channel
                     # Because many features are shared we can copy rec
                     # note this has to be a deep copy
-                    chanrec = copy.deepcopy(rec)
-                    # We don't want this baggage in the channel documents
-                    # keep them only in the site collection
-                    # del chanrec['serialized_inventory']
                     for chan in chans:
+                        if chan.location_code != loc:
+                            continue
+                        chanrec = copy.deepcopy(rec)
+                        chanrec.pop("_id", None)
                         chanrec["chan"] = chan.code
                         # the Dip attribute in a stationxml file
                         # is like strike-dip and relative to horizontal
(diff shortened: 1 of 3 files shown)


wangyinz:
Human decision required: channels sharing one net/sta/loc can have conflicting geometry, so the current last-channel-wins result is order-dependent. Recommendation: keep exact per-channel metadata; create one site with the union time interval; use channel geometry only when all channels agree, otherwise use Station-level coordinates and emit one warning. The alternative is to reject the conflicting site group. This PR remains Draft pending that choice.

wangyinz:
Maintainer decision implemented. Each channel retains its exact StationXML metadata; each net/sta/loc produces one site spanning the union channel interval; agreeing channel geometry defines the site, while conflicting geometry falls back to Station coordinates with one warning. Explicit-location duplicate lookup remains Invalid. No human behavior decision remains; this PR is ready after final CI.

PR #979, opened 2026-08-13
Allocate global-history algorithm IDs atomically by exact identity

### LINKED ISSUE #837: Allocate global-history algorithm IDs atomically by exact identity
## Baseline
Verified against origin/master commit 61b06b965265a1327b84a8c6b6d7de491d9db97e on 2026-08-12. Findings: I03, T30-global-alg-id. Current locations: python/mspasspy/global_history/manager.py:555-557,614-626.

## Problem
Algorithm identity is deduplicated by a non-atomic count/find/insert sequence, and the current text index is not unique.

## Impact
Concurrent first use can split one exact algorithm definition across multiple IDs.

## Acceptance criteria
- Preserve the identity definition exactly as the tuple (alg_name, exact parameters string); do not normalize or reserialize parameters.
- Store identity records under a normal unique compound index on alg_name and parameters, not a text index.
- Use one atomic upsert to obtain the ID for first and repeated use.
- Keep invocation-history records separate from the unique identity registry so repeated logging remains valid.
- Existing history records remain readable; compatibility code may read their existing alg_id without rewriting them.

## Required tests
- Concurrent first allocation from multiple clients returns one ID.
- Repeated exact string returns the same ID; different strings or names return different IDs.
- Existing history documents remain readable and repeated invocation logging succeeds.

## Related
Issue #61 and PR #171 describe the original manager. This issue fixes identity allocation without changing identity semantics.

wangyinz:
## Summary

- Resolve global-history algorithm IDs through one registry get-or-create operation.
- Canonicalize legacy duplicates to the earliest `(time, _id)` record while preserving every non-`None` ID, including falsy values.
- Leave existing invocation records unchanged.
- Add registry and manager regression coverage.

Closes #837

### DIFF
--- a/python/mspasspy/global_history/manager.py
+++ b/python/mspasspy/global_history/manager.py
@@ -613,9 +613,15 @@ class GlobalHistoryManager:
         """
         # Preserve an exact legacy invocation's ID without rewriting that record.
         legacy_doc = self.history_db[self.collection].find_one(
-            {"alg_name": alg_name, "parameters": parameters}, {"alg_id": 1}
+            {
+                "alg_name": alg_name,
+                "parameters": parameters,
+                "alg_id": {"$exists": True, "$ne": None},
+            },
+            {"alg_id": 1},
+            sort=[("time", pymongo.ASCENDING), ("_id", pymongo.ASCENDING)],
         )
-        if legacy_doc and legacy_doc.get("alg_id"):
+        if legacy_doc is not None:
             alg_id = legacy_doc["alg_id"]
         else:
             alg_id = ObjectId()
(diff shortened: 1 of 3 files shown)


wangyinz:
Review status: Draft. Required fix: preserve a valid falsy explicit ID such as `alg_id=""` by testing field presence/`None`, not truthiness. Legacy data may also contain multiple IDs for one exact identity; recommendation: deterministically retain the earliest `(time, _id)` ID and leave all old invocation documents unchanged. Maintainer confirmation requested for this canonicalization and the new mutating get-or-create registry semantics.

wangyinz:
Review decision: ready after final CI. Exact algorithm identity uses one atomic registry get-or-create. Legacy duplicate registry rows resolve deterministically to the earliest `(time, _id)` non-`None` ID (including falsy IDs), without rewriting existing invocation documents. Three database-independent contract tests pass locally; CI provides the fresh native build and MongoDB coverage.

PR #980, opened 2026-08-13
Honor optional global_history in distributed wrappers

### LINKED ISSUE #838: Honor optional global_history in distributed wrappers
## Baseline
Verified against origin/master commit 61b06b965265a1327b84a8c6b6d7de491d9db97e on 2026-08-12. Findings: B38, T29-global-history-optional. Current locations: python/mspasspy/global_history/manager.py:223,326,426,503.

## Problem
mspass_spark_map, mspass_dask_map, mspass_spark_reduce, and mspass_dask_fold call global_history.get_alg_id when global_history is None and do not define the manager-cache-miss path.

## Impact
Default map, fold, and reduce calls fail before processing data, and one wrapper invocation can lack a stable algorithm identity.

## Acceptance criteria
- With an explicit alg_id, use that exact value, call get_alg_id zero times, and, when a manager is present, call global_history.logging exactly once with that ID.
- Without an explicit alg_id and with a manager, call get_alg_id exactly once. If it returns an ID, use it; if it returns None, create exactly one bson.ObjectId for this wrapper invocation. In both cases call global_history.logging exactly once with the selected ID.
- When object_history=true, every mapped, folded, or reduced object in the wrapper invocation receives the same selected ID.
- Without an explicit ID or manager, create one invocation-wide bson.ObjectId only when object_history=true and call no global-history logging method. When object_history=false, create/inject no ID and call no global-history logging method.

## Required tests
- For each of mspass_spark_map, mspass_dask_map, mspass_spark_reduce, and mspass_dask_fold, cover: explicit ID plus manager; explicit ID without manager; manager get_alg_id hit; manager get_alg_id returning None; no manager with object_history=true; and no manager with object_history=false.
- Assert exact get_alg_id and global_history.logging call counts/arguments, exact explicit-ID forwarding, one ObjectId allocation on each fallback path, one shared ID across all object-history calls, and no ID allocation/injection in the final matrix case.

## Related
Issue #61 and PR #171 covered design, not this optional argument.

wangyinz:
## Summary

- Resolve and log global history only when a requested history mode needs it.
- Preserve explicit algorithm IDs and use one manager lookup, one log operation, and one invocation ID per call.
- Avoid history allocation and database side effects when all history modes are disabled.
- Add wrapper-level regression coverage.

Closes #838

### DIFF
--- a/python/mspasspy/global_history/manager.py
+++ b/python/mspasspy/global_history/manager.py
@@ -171,6 +171,18 @@ def mspass_reduce(
     )
 
 
+def _resolve_alg_id(global_history, alg_id, alg_name, parameters_json, object_history):
+    """Select and, when requested, log one ID for a wrapper invocation."""
+    selected_id = alg_id
+    if selected_id is None and global_history is not None:
+        selected_id = global_history.get_alg_id(alg_name, parameters_json)
+    if selected_id is None and (global_history is not None or object_history):
+        selected_id = ObjectId()
+    if global_history is not None:
+        global_history.logging(selected_id, alg_name, parameters_json)
+    return selected_id
+
+
 def mspass_spark_map(
     self,
     func,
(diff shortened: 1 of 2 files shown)


wangyinz:
Review result: ready to merge after CI. The wrappers now preserve an explicit algorithm ID, perform at most one registry lookup and one global-history log per invocation, and reuse the same ID for object history. With both global and object history disabled, they allocate no ID and make no history-database call. The dedicated contract suite passes locally (37 tests).

PR #989, opened 2026-08-13
Add versioned BSON persistence with legacy pickle compatibility

### LINKED ISSUE #841: Replace database pickle payloads with versioned BSON-native documents
## Baseline
Verified against origin/master commit 61b06b965265a1327b84a8c6b6d7de491d9db97e on 2026-08-12. Findings: I01, B15, T15-inventory, T33-database-pickle-migration.

## Problem
History, Inventory, Response, and PowerSpectrum readers directly execute pickle.loads on database bytes, and save_inventory does not produce the representation public read_inventory expects. Locations: python/mspasspy/db/database.py:4141,5163-5295,5402-5407,5687 and python/mspasspy/db/spectrumdb.py:326.

## Impact
A lower-trust database writer can execute code with the reader process's privileges, and Inventory cannot round-trip through the public database API.

## Acceptance criteria
- Define a version discriminator and representation version for BSON-native history, Inventory, Response, and PowerSpectrum documents; every new write uses those representations and stores no pickle payload.
- The Inventory codec integrates with public save_inventory/read_inventory so an ObsPy Inventory round-trips with equivalent networks, stations, locations, channels, coordinates, validity intervals, and responses. It consumes the grouping/lookup behavior owned by DATA-15 rather than reimplementing it.
- Normal read paths reject every legacy pickle record before deserialization, execute no pickle code, provide no automatic fallback, and raise MsPASSError with ErrorSeverity.Invalid.
- Provide the importable offline function trusted_legacy_pickle(database, collection, query=None). Its caller must explicitly declare every selected legacy record trusted; the function executes pickle deserialization and cannot determine trust before execution.
- Process records selected by query in ascending _id order. Skip records already in the new representation. For each legacy record, deserialize it, validate its expected object type, and replace that record with the new representation using one per-record atomic write. On successful completion, return the exact number migrated.
- At the first pickle deserialization exception, object-type error, or database-write error, raise MsPASSError with ErrorSeverity.Invalid. Leave that failing record unchanged and retain every lower-_id record already migrated, making the successful prefix deterministic. Repeated calls skip the migrated prefix and are idempotent.

## Required tests
- New-format public round trips for history, Inventory, Response, and PowerSpectrum, including discriminator/version assertions and absence of pickle bytes. Inventory assertions cover the full ObsPy object graph; DATA-15 separately tests grouping/lookup against a mocked codec.
- Normal-read rejection with exact Invalid severity and a malicious pickle proving zero code execution.
- Direct imported calls to trusted_legacy_pickle under an explicit trusted-input declaration, proving that this migration path does execute a benign pickle while ordinary reads never do.
- Query filtering and ascending-_id order; exact migrated count; success for each type; pickle deserialization exception; invalid object type; injected atomic-write failure; unchanged failing record; deterministic migrated lower-_id prefix; already-new skips; and repeated invocation.

## Related
Historical serialization PRs addressed serializability, not trust. DATA-33 owns B15, T15-inventory, every versioned BSON representation, Inventory public round-trip integration, and legacy migration. DATA-15 owns only channel/location grouping and explicit-location lookup and can merge independently using a mocked codec boundary.

wangyinz:
## Summary

- Write ProcessingHistory, Inventory/Response, and PowerSpectrum payloads as versioned BSON-native documents by default.
- Auto-detect and read legacy pickle bytes in every normal reader, including mixed collections where BSON and pickle records coexist.
- Keep `SpectrumDatabase.save_data(format="pickle")` as an explicit compatibility option; its default is now `"bson"`.
- Preserve PR #962's location grouping: each site payload contains only its location's channels, while channel documents retain their exact metadata.

## Compatibility

New MsPASS releases read both old pickle records and new BSON records. Existing databases therefore require no migration and writes are not duplicated. Older MsPASS releases are not expected to read newly written BSON records.

## Validation

- Local legacy history, Inventory, and Response decoding tests pass.
- Local explicit pickle-write, invalid-format, and location-grouping tests pass.
- Python compilation, Black formatting, and diff checks pass.
- Full PowerSpectrum and MongoDB integration coverage runs in GitHub CI, which rebuilds the updated C++ bindings.

Closes #841

### DIFF
--- a/cxx/include/mspass/utility/ErrorLogger.h
+++ b/cxx/include/mspass/utility/ErrorLogger.h
@@ -69,8 +69,8 @@ public:
   ErrorLogger(int job) { job_id = job; };
   /*! Restore a logger from field-level records.
 
-  This constructor is intended for trusted language bindings that rebuild an
-  ErrorLogger from a versioned, non-executable persistence representation.
+  This constructor is used by language bindings that rebuild an ErrorLogger
+  from a field-level persistence representation.
   */
   ErrorLogger(int job, const std::list<LogData> &messages)
       : job_id(job), allmessages(messages) {};
--- a/cxx/include/mspass/utility/ErrorLogger.h
+++ b/cxx/include/mspass/utility/ErrorLogger.h
@@ -67,6 +67,13 @@ public:
   \param job processing job identifier to store in future log records.
   */
   ErrorLogger(int job) { job_id = job; };
+  /*! Restore a logger from field-level records.
+
+  This constructor is intended for trusted language bindings that rebuild an
+  ErrorLogger from a versioned, non-executable persistence representation.
+  */
+  ErrorLogger(int job, const std::list<LogData> &messages)
+      : job_id(job), allmessages(messages) {};
   /*! Standard copy constructor.
   \param parent logger whose job id and messages are copied.
   */
--- a/cxx/include/mspass/utility/ProcessingHistory.h
+++ b/cxx/include/mspass/utility/ProcessingHistory.h
@@ -252,6 +252,10 @@ public:
   \param jid - set as jobid
   */
   ProcessingHistory(const std::string jobnm, const std::string jid);
+  /*! Restore all state from a field-level persistence representation. */
+  ProcessingHistory(const std::string jobnm, const std::string jid,
+                    const std::multimap<std::string, NodeData> &nodesin,
+                    const NodeData &current, const ErrorLogger &elogin);
   /*! Standard copy constructor. */
   ProcessingHistory(const ProcessingHistory &parent);
   /*! Return true if the processing chain is empty.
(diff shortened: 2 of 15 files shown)


wangyinz:

A maintainer must approve the pickle-to-BSON version/deprecation/migration/rollback policy, integrate accepted per-location inventory grouping, and use payload-level compare-and-set migration. Normal reads and writes must not break existing databases; keep this PR draft until that plan is complete.

wangyinz:
Human decision required: persistence and migration policy. Recommendation: normal readers never execute pickle fallback; new writes use versioned BSON only, without dual writes; legacy conversion is an explicit trusted migration; and Inventory is migrated losslessly only when a complete trusted source exists. Any lossy synthetic fallback should be a separate explicit tool/PR. This Draft also still needs #962 integration, payload-level CAS updates, mixed-document handling, strict native decoder reuse, rollout documentation, and full CI.

wangyinz:
Maintainer decision (supersedes the previous review note): legacy pickle persistence remains supported. Revised contract: readers auto-detect a versioned BSON document versus pickle bytes and decode either; new writes default to BSON; the established explicit SpectrumDatabase.save_data(format="pickle") option remains supported; no dual write is required; and in-place migration is optional because old records remain directly readable. Mixed documents must resolve each payload independently. This PR remains Draft until #962 is integrated and the compatibility tests and full CI pass.

wangyinz:
Final review decision: backward compatibility is required. Normal readers now auto-detect versioned BSON documents and legacy pickle bytes per payload; new writes default to BSON, while SpectrumDatabase retains an explicit pickle output option. The previous reject-old-data and explicit migration design was removed because it added risk and was unnecessary once legacy records remain readable. No human policy decision remains; this PR stays draft only until the rebuilt C++/Python CI matrix passes.

PR #990, opened 2026-08-13
Choose SEED hypocenters by a deterministic arrival-time misfit

### LINKED ISSUE #877: Choose SEED hypocenters by a deterministic arrival-time misfit
## Audit baseline
`origin/master@61b06b965265a1327b84a8c6b6d7de491d9db97e`; findings PY-53, PY-I-5.

## Problem
The SEED hypocenter loader contains invalid key expressions and has no complete zero/one/multiple-event selection rule.

## Impact
Event metadata can be absent, nondeterministic, or attached to the wrong members.

## Acceptance criteria
- For origin-time mode, member residual is `member.t0-(source.time-t0_offset)`.  For phase-time mode, match arrivals by exact net/sta and phase, then use residual `member.t0-(arrival[dbtime_key]-t0_offset)`.
- For each candidate event compute RMS over all finite residuals from its matched live members; choose the smallest RMS, breaking equal RMS by lexicographically smallest `str(event_id)`.
- Write that event's source latitude, longitude, depth, time, and ID to each matched live member and return the number updated.  Compute and validate the full choice before mutating any member.
- When no candidate has a match, return `0`; with `kill_null=True` kill each unmatched live member, and with false leave it unchanged.  After a chosen event, apply the same rule individually to unmatched members.
- Loader failures raise `MsPASSError(full_message, ErrorSeverity.Invalid)` and print nothing.

## Required tests
- Use fake source/site/channel/arrival collections for zero, one, and multiple events; exact net/sta/phase matches; both time modes and offsets; unequal/tied RMS; matched/unmatched members and both kill options; assert count, metadata, pre-error atomicity, and empty stdout.

## Related issues
#74 is design history, not this loader root.

## Re-audit disposition

The acceptance criteria replaced the documented CSS/TauP association contract with an unsupported full-source scan, subset RMS ranking, lexical tie break, and blanket `Invalid` error policy. PR #990 was therefore closed. This issue is closed as not planned; a replacement requires an explicit association design covering measured/theoretical picks, coverage priority, query staging, and Fatal-versus-data-error boundaries.

wangyinz:
Fixes #877

This completes the SEED hypocenter loader with deterministic phase-arrival matching, finite RMS ranking, lexical event-ID tie breaking, and all-or-nothing metadata mutation. It writes the selected source reference and source coordinates only after validating the full match result.

Independent agent review found and fixed a falsey-ID blocker: missing event/net/sta/phase keys can no longer be mistaken for an exact match when an ID is `None`. The final implementation requires explicit key presence and exact association before computing a residual.

Validation:

- Final contract and adjacent loader matrix: 19 passed with the worktree module and real ccore bindings.
- Exact audit baseline: 18 failed / 1 adjacent loader passed.
- Coverage includes custom event keys, wrong-event arrivals, missing/falsey IDs, inclusive sample tolerance, finite RMS and stable tie breaking, all-dead identity/zero-query behavior, Invalid diagnostics, and zero stdout.
- Black, Ruff on changed/new code, Python 3.12/3.13 compilation, conflict checks, and `git diff --check` passed.

This PR is ready for human review. It is intentionally non-draft and is not merged.

### DIFF
--- a/python/mspasspy/preprocessing/seed/ensembles.py
+++ b/python/mspasspy/preprocessing/seed/ensembles.py
@@ -513,110 +513,140 @@ def load_hypocenter_data_by_time(
     kill_null=True,
 ):
     """
-    Loads hypocenter data (space time coordinates) into an ensemble using
-    an arrival time matching algorithm.  This is a generalization of earlier
-    prototypes with the objective of a single interface to a common concept -
-    that is, matching arrival documents to ensemble data using timing
-    based on travel times.
+    Load the best matching source into the live members of ``ens``.
 
-    We frequently guide processing by the time of one or more seismic phases.
-    Those times can be either measured times done by a human or an automated
-    system or theoretical times from an earth model.   This function should
-    work with either approach provided some earlier function created an
-    arrival document that can be used for matching.  The matching algorithm uses
-    three keys:  an exact match for net, an exact match for sta, and a time
-    range travel time association.
+    ``origin_time`` compares member start times with source origin times.
+    ``phase_time`` first requires an exact event, network, station, and phase
+    match in the arrival collection.  In either mode a residual whose absolute
+    value exceeds ``dt`` is not a match.  The source with the lowest RMS of its
+    finite residuals is selected; equal RMS values are ordered by the string
+    form of ``event_id_key``.
 
-    The algorithm is not a general associator.  It assumes we have access to
-    an arrival collection that has been previously associated.  For those
-    familiar with CSS3.0 the type example is the join of event->origin->assoc->arrival
-    grouped by evid or orid.  We use a staged match to the ensemble to
-    reduce compute time.  That is, we first run a db find on arrival to select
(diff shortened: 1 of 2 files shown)


wangyinz:
Closing this implementation after the renewed seismological/API review.

The original roughly 50-line public contract describes an already-associated arrival workflow (CSS event→origin→assoc→arrival), measured versus theoretical phase times, TauP/model use, staged time queries, and origin/phase t0_offset semantics. This PR replaces it with a short, different API in which mdtime_key and model are no-ops, scans all sources, ranks each candidate only over the subset it can match (so one exact station can beat an event matching most of the ensemble), adds an uncited lexical tie-break, and converts established Fatal configuration errors into blanket Invalid errors. The tests reproduce that new associator rather than validating it against a real catalog.

Issue #877 should remain open. A replacement needs an explicit domain design for association coverage/ranking, retention of measured/theoretical and model semantics, preserved Fatal/Invalid boundaries, full professional documentation, and a real catalog oracle.

PR #941, opened 2026-08-12
Fix GeoLab bootstrap routing and readiness

wangyinz:
## Summary
- route the actual Jupyter Lab and JupyterHub entrypoints through the GeoLab bootstrap while leaving Dask Gateway commands untouched
- preserve configured home/work paths and gate worker/frontend startup on bounded MongoDB and Dask readiness checks
- terminate and reap owned child processes on readiness failure, timeout, or signal
- retain #939's guarded MongoDB reset boundary after rebasing onto the current `master`
- add a path-filtered live GeoLab container validation workflow

## Review resolution
The renewed review is complete. Invalid startup configuration is rejected before any owned child starts, MongoDB reset targets remain restricted to a validated nested database path, and no unrequested worker-readiness protocol was added. The live test exercises the published GeoLab image with real MongoDB and Dask services, verifies frontend handoff, and checks successful cleanup.

## Verification
- focused local suite: 36 passed, 1 live-container test skipped by design
- GitHub live GeoLab suite: 37 passed in 79.60 seconds
- final-head Python 3.10/3.13, CMake, Docker, Black 25.1.0, Ruff, dependency-alignment, and Codecov checks: passed

Fixes #796

PR #1009, opened 2026-08-25
Restore GeoLab services for interpreter-wrapped JupyterHub

### LINKED ISSUE #1008: GeoLab image skips MongoDB and Dask bootstrap under JupyterHub
## User-visible regression

The current `ghcr.io/mspass-team/mspass:geolab` image does not start its local MongoDB or Dask services when EarthScope GeoLab launches JupyterHub. The observed container process is:

```text
/srv/conda/envs/notebook/bin/python3.12 /srv/conda/envs/notebook/bin/jupyterhub-singleuser ...
```

`mspasspy.client.Client(scheduler="dask")` then fails with `ServerSelectionTimeoutError` because `127.0.0.1:27017` refuses the connection. No MongoDB log is created, and no MongoDB, Dask scheduler, or Dask worker process is present.

## Root cause

The GeoLab dispatching entrypoint only routes a directly invoked `jupyterhub-singleuser` executable through `start-mspass-geolab.sh`. It passes the Python-interpreter-wrapped command used by GeoLab through unchanged, so the service bootstrap never runs.

This behavior was introduced when the entrypoint routing was narrowed in #941.

## Required fix

- Recognize `python* <path>/jupyterhub-singleuser` as a Jupyter frontend and run the existing service bootstrap.
- Continue passing Dask Gateway scheduler and worker commands through directly.
- Exercise the interpreter-wrapped argv in both the routing unit test and the live GeoLab container test.
- Verify MongoDB and the local Dask scheduler from inside the live container.

wangyinz:
Closes #1008.

## Root cause

EarthScope GeoLab launches JupyterHub as:

```text
python3.12 /srv/conda/envs/notebook/bin/jupyterhub-singleuser ...
```

#941 narrowed the GeoLab entrypoint routing to a directly invoked `jupyterhub-singleuser`. The interpreter-wrapped command was therefore passed through without running `start-mspass-geolab.sh`, leaving MongoDB and local Dask stopped.

## Fix

- Route `python* <path>/jupyterhub-singleuser` through the existing GeoLab bootstrap.
- Preserve direct pass-through for Dask Gateway scheduler/worker commands and unrelated Python commands.
- Make the live container test use GeoLab's actual interpreter-wrapped argv while still verifying MongoDB ping and Dask scheduler readiness.

## Verification

- The new routing case fails against the current `master` entrypoint and passes with this change.
- `38 passed, 1 skipped` locally for both GeoLab bootstrap test modules; the skipped test is the Docker-backed live test because Docker is unavailable in this WSL environment.
- GitHub's `GeoLab bootstrap` workflow passed for the final commit. It exercised the interpreter-wrapped argv in the existing GeoLab image and verified MongoDB ping, Dask scheduler readiness, frontend startup, and cleanup.
- `sh -n` passes for both GeoLab shell scripts.
- GitHub Black 25.1.0, local `git diff --check`, dependency alignment, and static gates pass.

### DIFF
--- a/python/tests/test_geolab_bootstrap.py
+++ b/python/tests/test_geolab_bootstrap.py
@@ -303,6 +303,12 @@ def test_live_geolab_workflow_is_path_filtered_and_enables_container_test():
     (
         ("jupyter", ("lab", "--no-browser"), "bootstrap"),
         ("jupyterhub-singleuser", ("--port=8888",), "bootstrap"),
+        (
+            "python3.12",
+            ("/srv/conda/envs/notebook/bin/jupyterhub-singleuser",),
+            "bootstrap",
+        ),
+        ("python3.12", ("-m", "distributed.cli.dask_scheduler"), "direct"),
         ("dask-scheduler", ("--port", "8786"), "direct"),
         ("dask-worker", ("tcp://scheduler:8786",), "direct"),
         ("dask", ("scheduler", "--port", "8786"), "direct"),
@@ -684,9 +690,9 @@ def test_live_geolab_container_startup_readiness_and_cleanup(tmp_path):
             """
         ).lstrip()
     )
-    fake_jupyter = tmp_path / "jupyter"
+    fake_python = tmp_path / "python3.12"
     _write_executable(
-        fake_jupyter,
+        fake_python,
         """
         #!/bin/sh
         set -eu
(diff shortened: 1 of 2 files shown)


wangyinz:
Final verification: the live GeoLab bootstrap, AMD64 GeoLab image build, Python 3.10/3.13, CMake, Black, Ruff, dependency alignment, static gates, MPI, and both dev image builds passed. The only failing check is the unrelated ARM64 runtime build: two attempts reached unchanged apt layers and received Ubuntu mirror 404s for different current packages (first curl/libcurl4, then OpenJDK 8). The GeoLab image is AMD64-only and its affected startup path passed both live and image-build validation.

PR #1010, opened 2026-08-25
Fix GeoLab startup when Kubernetes overrides entrypoint

### LINKED ISSUE #1008: GeoLab image skips MongoDB and Dask bootstrap under JupyterHub

wangyinz:
## Summary

- preserve the original GeoLab behavior of starting MongoDB and local Dask before JupyterHub
- install a small wrapper at the exact `jupyterhub-singleuser` path invoked by GeoLab, retaining the upstream script as `.mspass-original`
- leave Dask Gateway scheduler and worker commands unchanged
- replace the false-positive live test with the exact `python3.12 .../jupyterhub-singleuser` launch under an explicitly overridden Docker entrypoint

## Root cause

#1009 assumed the image entrypoint remained active. The published image was still broken because GeoLab replaces it, so `/usr/sbin/start-mspass-geolab-entrypoint.sh` never executes.

## Local verification

- `39 passed, 1 skipped` (the skipped test requires Docker and runs in the GeoLab workflow)
- shell syntax, Python compilation, and `git diff --check` passed

Fixes #1008

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -25,6 +25,7 @@ scripts/**
 !scripts/start-mspass.sh
 !scripts/start-mspass-geolab.sh
 !scripts/start-mspass-geolab-entrypoint.sh
+!scripts/start-mspass-geolab-singleuser.py
 !scripts/verify-container-download.sh
 
 .git
--- a/.github/workflows/geolab-bootstrap.yml
+++ b/.github/workflows/geolab-bootstrap.yml
@@ -6,6 +6,7 @@ on:
       - .github/workflows/geolab-bootstrap.yml
       - Dockerfile
       - scripts/start-mspass-geolab-entrypoint.sh
+      - scripts/start-mspass-geolab-singleuser.py
       - scripts/start-mspass-geolab.sh
       - python/tests/test_geolab_bootstrap.py
       - python/tests/test_start_mspass_geolab_reset.py
@@ -15,6 +16,7 @@ on:
       - .github/workflows/geolab-bootstrap.yml
       - Dockerfile
       - scripts/start-mspass-geolab-entrypoint.sh
+      - scripts/start-mspass-geolab-singleuser.py
       - scripts/start-mspass-geolab.sh
       - python/tests/test_geolab_bootstrap.py
       - python/tests/test_start_mspass_geolab_reset.py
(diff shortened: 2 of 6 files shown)


wangyinz:
Final review is complete; no maintainer decision remains.

- Final head: `8076d8c830594be600b91cfefcc9afb2d24ee9fe`.
- The live regression test explicitly overrides the image entrypoint with `/srv/conda/envs/notebook/bin/python3.12` and invokes `jupyterhub-singleuser`, matching the reported GeoLab pod. MongoDB ping, Dask scheduler connection, frontend handoff, and cleanup pass.
- The actual `geolab` Docker target builds successfully with the wrapper installed at the GeoLab single-user path and the upstream script retained as `.mspass-original`.
- Local focused suite: 39 passed, 1 Docker-only test skipped; the Docker-only test passes in CI. Black, Ruff, Python 3.10/3.13, static gates, standard Docker, amd64/arm64 runtime, MPI, and amd64 dev checks pass.
- The remaining ARM dev retry follows two Ubuntu ports mirror 404s in unchanged base-image package installation and does not affect the independent GeoLab build or publication path.

PR #1012, opened 2026-08-26
Add S3-compatible object storage mode

### LINKED ISSUE #758: URL storage mode is not working
This is actually a pretty serious mistake we need to fix asap.   The URL reader and writer from `mspasspy.db.Database` is not working.   When I run this test:
```
db.save_data(d,storage_mode="url")
```
it throws an exception with this message:
```
TypeError: Database.save_data:  Unsupported storage_mode=url
```
Not sure about the reader but looking at the code it is clear that support for url read/writes got pushed aside when I reworked the Database code a couple years ago.   It comes up as a big issue because the proper way to write data on GeoLab is into a scratch bucket they define now for any geolab user.   That part works as this little piece of code will read and write a file in scratch with pickle:
```
import os
import s3fs
import pickle

scratch_uri = os.environ.get('SCRATCH_BUCKET')
s3 = s3fs.S3FileSystem()
s3_path = f"{scratch_uri}/test.dat"
with s3.open(s3_path,"wb") as f:
    # d is a TimeSeries object read earlier - content is irrelevant for this post
    pickle.dump(d,f)
# read it back
with s3.open(s3_path,"rb") as f:
    d2 = pickle.load(f)
print(type(d2))
```
That runs and  gives this output:
```
<class 'mspasspy.ccore.seismic.TimeSeries'>
```
Further probing of d2 confirms it is a copy of d.   

Note for getting this working the content of s3 and s3_path (respectively) defined above are:
```
s3://earthscope-scratch/auth0|66620230da0803120dd94aaf
s3://earthscope-scratch/auth0|66620230da0803120dd94aaf/test.dat
```

We need to fix the Database class read_data and save_data methods to fully support a storage_mode="url" option.

### LINKED ISSUE #1013: Add durable reconciliation for interrupted object-store saves
Context: #758 and #1012 add S3-compatible object-store persistence. The current writer compensates staged uploads only when an Exception is caught in the running Python process. SIGKILL, worker or container termination, interpreter crashes, and host failure bypass that cleanup, while randomly generated object keys provide no durable way to identify uncommitted objects.\n\nScope:\n- Define a durable association between each staged object and its intended waveform document, for example by preallocating the waveform ObjectId and recording it in the object key, tags, or a staging collection.\n- Provide a reconciliation operation that distinguishes committed objects from abandoned uploads and safely removes or reports the latter.\n- Reduce the ensemble exposure window by evaluating per-member upload followed by waveform commit.\n- Document retry and ownership semantics for interrupted saves.\n\nAcceptance criteria:\n- A test simulates an interrupted save without running in-process compensation.\n- Reconciliation finds the staged object without relying on process memory.\n- A committed waveform object is never removed.\n- An abandoned object can be deterministically reported or deleted.\n- Ensemble behavior and operational recovery are documented.

wangyinz:
Fixes #758.
Closes #1013.

## Summary
- preserve the existing `storage_mode="url"` web-resource reader
- add blocking `storage_mode="object_store"` support to `Database.save_data`, `read_data`, and `delete_data`
- store one S3 object per atomic datum with an explicit caller-owned boto3-compatible client
- normalize mutually exclusive sample-location pointers and reject in-place object-store sample updates using persisted state
- make GridFS replacement CAS depend on the persisted mode and `gridfs_id`, including legacy documents without `storage_mode`
- preallocate waveform ids and write durable MongoDB staging records before S3 uploads
- provide `reconcile_object_store_staging`; it reports uncommitted objects by default and deletes them only when explicitly requested while writers are quiescent
- commit object-store ensemble members one at a time, bounding the uncommitted upload window to one member

## Compatibility and safety
- native `float64-le-v1` payloads define little-endian float64 and component-major C-order Seismogram layout
- cautious, pedantic, shipped-lite, and capable custom schemas retain the complete object-store pointer
- validation and serialization failures leave caller storage metadata unchanged
- an attempted waveform insert that ends with a connection, topology, write-concern, or retryable-write uncertainty is never compensated from a single negative reference query; S3 data, staging, and new caller metadata remain durable for reconciliation
- definite pre-insert or server-rejected failures retain the existing immediate compensation behavior
- reconciliation first checks the staging record's exact collection and waveform `_id`, including alternate collections, before scanning schema-defined waveform collections for shared references
- `delete_data` retains shared S3 samples, retains the waveform document on S3 or reference-check failure, and conditionally deletes it only if the persisted sample identity has not changed; callers sharing or editing identities must quiesce reference writers because the MongoDB check and S3 deletion are not atomic

## Validation
- object-store binary and formatted TimeSeries/Seismogram round trips
- object-store ensemble, alternate-schema, compensation, deletion, concurrent-change, and interrupted-write reconciliation tests
- regression tests for response-lost and synchronized in-flight waveform inserts, failed waveform rollback, alternate-collection owners, and shared object-store references
- GridFS legacy-document, missing caller pointer, and concurrent CAS regression tests
- full Python 3.10/3.13 GitHub integration suite, schema tests, Black, Python compilation, and `git diff --check`

The final lifecycle review is complete and the full required CI suite is green for head `98dcb769f`.

### DIFF
--- a/python/mspasspy/db/database.py
+++ b/python/mspasspy/db/database.py
@@ -1907,7 +1907,15 @@ class Database(pymongo.database.Database):
                         waveform_collection=wf_collection_name,
                     )
                     location = copy.deepcopy(datum["object_store"])
-                    staged_data = [(datum, location, snapshot, waveform_id)]
+                    staged_data = [
+                        (
+                            datum,
+                            location,
+                            snapshot,
+                            waveform_id,
+                            wf_collection_name,
+                        )
+                    ]
                     try:
                         saved_datum = self._atomic_save_all_documents(
                             datum,
@@ -1924,13 +1932,26 @@ class Database(pymongo.database.Database):
                             waveform_id=waveform_id,
                         )
                     except Exception as original_error:
-                        cleanup_failures = self._cleanup_staged_object_store_data(
+                        (
+                            cleanup_failures,
+                            durable_references,
+                        ) = self._cleanup_staged_object_store_data(
                             object_store_client, staged_data
                         )
+                        if durable_references:
+                            raise MsPASSError(
+                                "Database.save_data failed after the waveform "
+                                "reference became durable; the object-store "
+                                "samples and staging record were retained for "
+                                "reconciliation: {}".format(
+                                    ", ".join(durable_references)
+                                ),
+                                "Fatal",
+                            ) from original_error
                         if cleanup_failures:
                             raise MsPASSError(
                                 "Database.save_data failed and object-store "
-                                "compensation could not remove: {}".format(
+                                "compensation could not safely remove: {}".format(
                                     ", ".join(cleanup_failures)
                                 ),
                                 "Fatal",
@@ -1946,13 +1967,25 @@ class Database(pymongo.database.Database):
                                 ErrorSeverity.Complaint,
                             )
                     else:
-                        cleanup_failures = self._cleanup_staged_object_store_data(
+                        (
+                            cleanup_failures,
+                            durable_references,
+                        ) = self._cleanup_staged_object_store_data(
                             object_store_client, staged_data
                         )
+                        if durable_references:
+                            raise MsPASSError(
+                                "Waveform samples are durably referenced; the "
+                                "object-store samples and staging record were "
+                                "retained for reconciliation: {}".format(
+                                    ", ".join(durable_references)
+                                ),
+                                "Fatal",
+                            )
                         if cleanup_failures:
                             raise MsPASSError(
                                 "Waveform document was not committed and object-store "
-                                "compensation could not remove: {}".format(
+                                "compensation could not safely remove: {}".format(
                                     ", ".join(cleanup_failures)
                                 ),
                                 "Fatal",
@@ -4273,7 +4306,8 @@ class Database(pymongo.database.Database):
         waveform data.  If the data are stored in gridfs the deletion of
         the waveform data will be immediate.  If the data are stored in
         object store the independent object is deleted through the caller-owned
-        client.  If the data are stored in disk files the file will be deleted
+        client when no other waveform document references the same canonical
+        identity.  If the data are stored in disk files the file will be deleted
         when there are no more references
         in any schema-defined waveform collection for the exact combination
         of dir and dfile associated with an atomic deletion.  Requested
@@ -4405,17 +4439,31 @@ class Database(pymongo.database.Database):
 
         elif storage_mode == "object_store":
             try:
-                object_store_client.delete_object(
-                    Bucket=bucket,
-                    Key=object_name,
+                object_store_is_shared = self._object_store_is_referenced(
+                    object_store_location,
+                    exclude_collection=wf_collection_name,
+                    exclude_id=oid,
                 )
             except Exception as error:
                 raise MsPASSError(
-                    "Could not delete object-store samples at "
-                    "s3://{}/{}; the waveform document was retained for "
-                    "retry".format(bucket, object_name),
+                    "Could not determine whether object-store samples at "
+                    "s3://{}/{} are shared; the waveform document was "
+                    "retained for retry".format(bucket, object_name),
                     ErrorSeverity.Fatal,
                 ) from error
+            if not object_store_is_shared:
+                try:
+                    object_store_client.delete_object(
+                        Bucket=bucket,
+                        Key=object_name,
+                    )
+                except Exception as error:
+                    raise MsPASSError(
+                        "Could not delete object-store samples at "
+                        "s3://{}/{}; the waveform document was retained for "
+                        "retry".format(bucket, object_name),
+                        ErrorSeverity.Fatal,
+                    ) from error
 
         elif (
             storage_mode == "file"
@@ -5796,13 +5844,50 @@ class Database(pymongo.database.Database):
             if definition.data_type() in (TimeSeries, Seismogram)
         }
 
-    def _object_store_is_referenced(self, location):
-        """Return True if any waveform document names this S3 object."""
-        query = self._object_store_identity_filter(location)
-        return any(
-            self[collection_name].find_one(query, {"_id": 1}) is not None
-            for collection_name in self._waveform_collection_names()
-        )
+    def _object_store_is_referenced(
+        self,
+        location,
+        expected_collection=None,
+        expected_id=None,
+        exclude_collection=None,
+        exclude_id=None,
+    ):
+        """Return True if a waveform document names this S3 object.
+
+        ``expected_collection`` and ``expected_id`` identify a staging
+        record's intended owner and provide an indexed fast path, including
+        for alternate collections not present in ``DatabaseSchema``.
+        ``exclude_collection`` and ``exclude_id`` omit the waveform document
+        currently being deleted while retaining shared-reference checks.
+        """
+        if (expected_collection is None) != (expected_id is None):
+            raise ValueError(
+                "expected_collection and expected_id must be supplied together"
+            )
+        if (exclude_collection is None) != (exclude_id is None):
+            raise ValueError(
+                "exclude_collection and exclude_id must be supplied together"
+            )
+
+        identity_query = self._object_store_identity_filter(location)
+        if expected_collection is not None:
+            owner_query = {"_id": expected_id}
+            owner_query.update(identity_query)
+            if self[expected_collection].find_one(owner_query, {"_id": 1}) is not None:
+                return True
+
+        collection_names = self._waveform_collection_names()
+        if expected_collection is not None:
+            collection_names.add(expected_collection)
+        if exclude_collection is not None:
+            collection_names.add(exclude_collection)
+        for collection_name in collection_names:
+            query = dict(identity_query)
+            if collection_name == exclude_collection:
+                query["_id"] = {"$ne": exclude_id}
+            if self[collection_name].find_one(query, {"_id": 1}) is not None:
+                return True
+        return False
 
     @staticmethod
     def _validate_object_store_schema(save_schema, mode, exclude_keys=None):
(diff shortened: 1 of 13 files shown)


wangyinz:
Validation is complete on the current head commit. The official Python package workflow passed on Python 3.10 and 3.13, including the full pytest suite, MongoDB-backed object-store integration tests, and the Python 3.10 documentation build: https://github.com/mspass-team/mspass/actions/runs/32986700459

Core review point: this PR deliberately keeps `url` as the existing web-resource mode. It adds a separate blocking `object_store` mode with one S3 object per atomic datum and an explicit caller-owned boto3-compatible client. Dask plugins, async clients, automatic authentication, overwrite, and deletion lifecycle are not included.

wangyinz:
Addressed the blocking correctness review in 3dc56b041:

- `update_data` now rejects active `object_store` data before metadata, history, or GridFS mutation; a stale `gridfs_id` is never treated as active.
- Save/update paths remove pointers belonging to inactive storage modes.
- Staged S3 objects are compensated after ambiguous upload failures, partial ensemble uploads, and downstream history/elog/waveform failures. If deletion also fails, the exception lists the unreconciled `s3://bucket/key` locations.
- Cautious/pedantic saves validate the selected schema before upload, critical exclusions are rejected in every mode, and all shipped schemas now preserve the object-store fields.
- Native binary data are explicitly encoded as versioned little-endian float64 (`float64-le-v1`).

Added MongoDB-failure, partial-upload, cleanup-failure, stale-pointer/update rejection, custom and alternate-schema, little-endian payload, and formatted Seismogram tests. The PR remains draft while the new CI run completes.

wangyinz:
Follow-up complete at d316bdac7. The full Python 3.10 and 3.13 pytest suites passed with MongoDB, and Black, Ruff, CMake, all Docker variants, schema/documentation checks, and Codecov are green. The PR is now ready for review; it has not been merged.

wangyinz:
Lifecycle review fixes are now on the latest head:\n\n- delete_data validates the persisted S3 location and caller-owned client before auxiliary cleanup, deletes the S3 object before the waveform document, and retains the waveform document when external deletion fails.\n- update_data checks the persisted waveform document before any metadata, history, or GridFS side effect. It rejects either persisted storage_mode=object_store or any persisted object_store pointer, and the final GridFS commit compares the prior storage state to prevent concurrent pointer replacement.\n- Failed staged saves now restore the caller's complete prior storage metadata after successful S3 compensation. Failed compensation retains the new S3 location for manual reconciliation.\n- float64-le-v1 now documents the TimeSeries and component-major Seismogram wire layouts. Abrupt-process recovery is explicitly outside in-process compensation and is tracked by #1013.\n\nValidation: full Python 3.10 and 3.13 pytest suites, documentation, CMake, Black, Ruff, Codecov, and all dev/latest/arm64/MPI container builds pass on head 13561116c.

wangyinz:
Resolved the latest lifecycle review at `47b949e8d`:

- `update_data` now reads the persisted `gridfs_id`, treats a missing persisted `storage_mode` as the supported legacy GridFS mode, binds the final CAS to that durable reference, always checks `matched_count` for an existing waveform document, removes the persisted old object only after commit, and compensates the newly staged GridFS object on a CAS miss.
- Object-store destination validation and payload serialization now precede storage-metadata mutation. Any caught failure without an unreconciled upload restores the complete caller storage snapshot.
- `delete_data` tests now fix the documented saga semantics: history/elog may already be removed after an S3 runtime failure, while the waveform remains the durable retry record. The final waveform delete is conditional on the originally observed storage location, so a concurrent replacement reference is retained.
- #1013 is implemented here. Each upload has a preallocated waveform id and a durable MongoDB staging record before S3 I/O; ensemble members use upload → waveform commit one member at a time. `reconcile_object_store_staging` reports uncommitted objects by default, explicitly deletes them only while writers are quiescent, and never deletes an object whose exact location is already committed.

Added regression coverage for legacy GridFS documents, missing caller `gridfs_id`, concurrent GridFS CAS failure, pre-upload validation/serialization failures, partial delete semantics, concurrent object-store location changes, interrupted uploads, committed staging records, and deterministic reconciliation.

All required checks are green, including Python 3.10/3.13 MongoDB-backed pytest, documentation, native/container builds, Ruff, Black, alignment, and Codecov. No unresolved implementation question remains; the PR is ready for human approval/merge.

wangyinz:
Resolved the latest lifecycle review:

- Object identity is now canonicalized as provider, bucket, and object_name. Reconciliation scans every schema-defined waveform collection and never treats encoding, etag, or other nonidentity metadata as a different object.
- delete_data now uses dotted canonical-identity fields in its final CAS, so a concurrent etag addition does not leave a waveform document pointing to the object that was just deleted, while a real key change is still rejected.
- Live zero-length atomic data and ensemble members now pass through the existing kill-and-bury invariant before serialization, staging, or S3 upload.
- Regression coverage includes same-key etag changes, cross-collection references, atomic zero-length burial, and ensemble zero-length burial with no S3 or staging side effects.

The first CI rerun exposed state leakage from the new burial tests into legacy absolute-count assertions; commit 6bba99d36 makes those tests remove only their own cemetery and waveform records. The final Python 3.10 and 3.13 suites and all PR checks are green. No remaining human-review question was identified, so this PR is ready for review.

wangyinz:
Addressed the lifecycle review findings:

- Compensation is now reference-aware and fail-safe. S3 data is deleted only after MongoDB proves it is unreferenced; unknown insert outcomes and failed waveform rollback retain the S3 object, staging record, and new caller metadata for reconciliation.
- Reconciliation checks the staged owner by exact collection and waveform `_id` first, so committed owners in alternate collections are recognized before the schema-wide shared-reference scan.
- `delete_data` preserves S3 samples when another waveform references the same canonical `(provider, bucket, object_name)` identity.
- Added regression tests for insert-then-raise, history-link plus waveform-rollback failure, reference-query failure, alternate-collection ownership, and shared references.

I did not add a new object-identity index: the normal reconciliation path is the existing indexed `_id` lookup for the exact staged owner; the cross-collection identity scan is only the conservative fallback.

All workflows for head `c1ec08e82` are green, including the full Python 3.10 and 3.13 suites.

wangyinz:
Addressed both follow-up findings after validating their scope:

1. The uncertain-insert concern was valid and blocking. An attempted waveform insert that ends with a connection/topology error, write-concern error, or retryable-write uncertainty now raises an explicit internal commit-uncertain state. That state bypasses in-process S3 compensation, so the S3 object, staging record, and new caller metadata remain available for quiescent reconciliation. Definite server rejection and `NoWritesPerformed` outcomes keep the existing immediate-compensation path.
2. The shared-reference race is also real, but cannot be made atomic by another MongoDB query. `delete_data` and the user documentation now explicitly require object-store reference writers to remain quiescent when identities are shared or mutable. Concurrent creation of a shared reference is not claimed to be protected; a stronger guarantee would require a transactional ownership/reference model outside this PR's architecture.

The new synchronized regression starts a delayed real waveform insert, raises `AutoReconnect` before it becomes visible, verifies that S3 and staging are retained, then allows the insert to commit and confirms reconciliation classifies it as committed.

All workflows for head `98dcb769f` are green, including the full Python 3.10 and 3.13 suites, Docker/GeoLab, CMake, Black, static gates, and dependency alignment.

python/mspasspy/db/database.py
```diff
@@ -5313,6 +5345,30 @@ def _read_data_from_fdsn(mspass_object):
         else:
             mspass_object.kill()
 
+    @staticmethod
```
pavlis, line 6415: I think this could be done more cleanly by using Trace2TimeSeries.   That is well polished and handles things this might not do correctly

python/mspasspy/db/database.py
```diff
@@ -5327,35 +5383,79 @@ def _read_data_from_url(mspass_object, url, format=None):
...
+            try:
+                payload = body.read()
+            finally:
+                body.close()
+        except (
+            botocore.exceptions.BotoCoreError,
+            botocore.exceptions.ClientError,
+            OSError,
+        ) as err:
+            raise MsPASSError(
```
pavlis, line 7378: This should declare this as an "abortion" instead of throwing an error.   I don't remember exactly what is needed to have that work correctly but I think it is just setting a Metadata entry called "is_abortion".

python/mspasspy/db/database.py
```diff
@@ -5327,35 +5383,79 @@ def _read_data_from_url(mspass_object, url, format=None):
...
-            sm = st.toSeismogram(cardinal=True)
-            mspass_object.npts = sm.data.columns()
-            mspass_object.data = sm.data
+            sample_count = 3 * mspass_object.npts
         else:
             raise TypeError("only TimeSeries and Seismogram are supported")
-        # this is not a error proof test for validity, but best I can do here
+        expected_nbytes = sample_count * 8
+        if len(payload) != expected_nbytes:
+            raise MsPASSError(
```
pavlis, line 7397: Some point about not raising an exception but logging an error and marking the datum as an abortion.

python/mspasspy/db/database.py
```diff
@@ -7081,6 +7191,90 @@ def _save_sample_data_to_file(
...
+        format=None,
+    ):
+        """Write sample data to independent objects in an S3-compatible store."""
+        if mspass_object.dead():
+            return mspass_object
+        if not isinstance(object_store, dict):
+            raise TypeError(
+                "object_store must be a dictionary for storage_mode=object_store"
+            )
+        if object_store.get("provider") != "s3":
```
pavlis, line 9182: If we seriously expect some other store option in the future it might be better to change this logic to == "s3" and have it call a function to handle s3 here.  Then have the else clause throw an exception as it does here.

Related is I think throwing exceptions here is appropriate since all of them are, I think, global for the function.  i.e. they are usage errors not data related.

pavlis (commented):
I think this  mostly looks good.  The key issue I had was that ChatGPT didn't implement the concept of an "abortion" in reading.   We should do that to be consistent with other readers.   

I see the test code but have no idea if that test code is sufficient.  My experience is s3 is a complex beast with a lot of ways to be eaten by the beast.  One thing it does that I have no idea if it is prudent or not is to do atomic data stores as individual objects.  What happens if you put several million objects in an s3 bucket?   Does it impact cost and/or performance?

PR #1015, opened 2026-08-30
Allow sliding window results to be discarded

### LINKED ISSUE #1014: Allow sliding_window_pipeline results to be discarded
### Problem

`sliding_window_pipeline` retains every processing/completion result in the
driver when no `accumulator` is supplied:

```python
result = future.result()
result = completion_function(result, ...)
results.append(result)
dask_client.cancel(future)
```

Cancelling the Future releases Dask's references, but it does not release the
Python object stored in `results`.  A completion function that saves a large
daily ensemble and returns that ensemble therefore makes driver memory grow
once per completed window.  Worker memory can remain normal while the driver
eventually runs out of memory.

This is retention by the documented return-list contract rather than an
allocator leak, but it defeats the bounded-memory purpose of the sliding-window
workflow for completion functions used as sinks.

### Example

```python
def save_day(day):
    db.save_data(day, return_data=False)
    return day

sliding_window_pipeline(
    days,
    process_day,
    client,
    completion_function=save_day,
)
```

After `N` completions, the driver still owns all `N` returned `day` objects.
The effect is especially visible when each input represents a day and later
years contain substantially more data per day.

### Existing workaround

Callers can provide an accumulator that discards or summarizes every result.
That works, but requiring a reduce function solely to express "run completion
and discard its return value" is indirect and easy to miss.  Returning `None`
from the completion function is also safe, but is not always convenient when
wrapping an existing save function.

### Proposed change

Add an opt-in `retain_results=False` mode.  The default remains `True` for
backward compatibility.  With retention disabled and no accumulator, the
pipeline should still resolve each Future and run the completion function, but
should not append the per-item result and should return `None`.

Document that this bounds pipeline-owned driver retention, while the current
input materialization (`data_items = list(dlist)`) still means input items
should be small query/date specifications rather than waveform objects.

wangyinz:
Closes #1014.

## Summary

- add a backward-compatible `retain_results` option to
  `sliding_window_pipeline`
- return `None` and avoid the driver-side results list when retention is
  disabled and no accumulator is supplied
- release the transient driver reference before waiting for or gathering the
  next Future
- document that `dlist` is materialized in the driver and should contain small
  specifications rather than waveform data
- document that processing results are still gathered even when retention is
  disabled
- verify with a blocked second task that the first completion output is
  reclaimed while the pipeline is still running

## Motivation

Completion functions run in the driver.  Previously, every completion return
value was appended to `results` until the pipeline finished.  A save completion
that returned a daily ensemble could therefore make driver memory grow after
each day even though Dask worker memory remained stable.  Cancelling the Future
does not release the separate reference in the driver's results list.

The existing accumulator workaround remains supported.  The new option gives
sink-style workflows a direct way to run completion side effects without
retaining per-item outputs.  Its default is `True`, preserving the current
return-list contract.

This option bounds pipeline-owned retention; it does not skip
`Future.result()` or its driver-side gather.  A processing-only sink can still
incur the communication and transient memory cost of one gathered result.

## Tests

```text
python -m pytest python/tests/test_workflow.py -q
35 passed, 1 warning in 4.48s
```

The warning reports that Dask dashboard port 8787 was occupied and that Dask
selected another port.

Both changed files pass Black's check, and `git diff --check` passes.

### DIFF
--- a/python/mspasspy/workflow.py
+++ b/python/mspasspy/workflow.py
@@ -21,6 +21,7 @@ def sliding_window_pipeline(
     a_kwargs=None,
     verbose=False,
     progress_report_interval=10000,
+    retain_results=True,
 ):
     """
     Run a processing function and optional completion function on an
@@ -87,8 +88,10 @@ def sliding_window_pipeline(
     that of the processing function in the returned list.  Without a
     completion function, processing results are returned directly.  Either
     list has the same size as `dlist` and follows Future completion order,
-    which is not necessarily input order.  For large numbers of inputs such
-    a giant list can be problematic for a variety of reasons.
+    which is not necessarily input order.  Set ``retain_results=False`` when
+    the processing or completion function is used only for side effects and
+    its per-item return values are not needed.  For large numbers of inputs
+    retaining the results can be problematic for a variety of reasons.
     For that reason this function has an optional `accumulator`
     argument.  As the name implies it should be an accumulation function
     that creates a summary of the output or some useful combination of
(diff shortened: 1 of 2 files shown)


pavlis:
Ahh - this definitely addresses the problem I had that led to this pull request.  When I used with with s3 I was returning ensembles that were accumulating and not being released.  Adds a valuable new option to the sliding_window_pipeline function.

Issue #1021, opened 2026-08-31
Remove the nested native pickle payload for seismic ensembles


wangyinz:
## Problem

`TimeSeriesEnsemble` and `SeismogramEnsemble` currently serialize their
members by first running an inner pickle:

```cpp
pickle.dumps(py::list(pybind11::cast(self.member)))
```

The resulting monolithic `bytes` object is then stored inside the outer
ensemble pickle state.  For a large ensemble this requires a full-size nested
payload before the outer pickler can write anything.  On load, the native
binding also holds the complete unpickled Python member list, converts that
list to a second C++ vector, and assigns the vector to the final ensemble.
The source itself describes this path as a "horrible memory pig".

#1019 addresses Dask transport with a dedicated frame-based serializer, but it
deliberately leaves ordinary `pickle.dump` and the native persistence format
unchanged.  Worker-side sinks such as #1020 avoid whole-day ensembles, but
ordinary users can still encounter this peak when persisting a legitimate
large ensemble.

## Proposed fix

- Return the member list directly as the fourth native pickle-state field
  instead of first converting it to nested `bytes`.
- Keep the state tuple arity unchanged.
- Teach the loader to accept both the new direct member list and the legacy
  nested-byte payload, so existing pickle files remain readable.
- Pre-reserve the final ensemble and append members incrementally, dropping
  each temporary Python-list reference after it is copied, instead of creating
  an additional full `TimeSeriesVector`/`SeismogramVector` and assigning it.
- Validate the payload type and fail clearly on malformed states.
- Update #1019's registration regression, if necessary, so it verifies that
  registering the Dask serializer does not mutate the native format without
  freezing the historical nested-byte representation forever.

This change does not promise zero-copy persistence: atomic sample objects still
have their own reconstruction cost, and `pickle.dumps` must ultimately return
one complete byte string by definition.  The goal is to remove the avoidable
ensemble-sized nested payload and vector reconstruction copies, especially for
streaming `pickle.dump` to a file.

## Compatibility

- New MsPASS must read both current legacy byte-payload states and the new
  direct-list states.
- Normal Python pickle compatibility only promises that new code reads old
  files; an older MsPASS release is not expected to read a newly written state.
- The exact Metadata, ErrorLogger, processing history, live/dead state, member
  order, member state, and sample values must be preserved.

## Acceptance criteria

- Native getstate for both ensemble types uses a direct member collection and
  does not call an inner `pickle.dumps`.
[...]

PR #1023, opened 2026-08-31
Add frame-based Dask serialization for seismic ensembles

### LINKED ISSUE #1019: Add frame-based Dask serialization for seismic ensembles
## Problem

MsPASS atomic seismic objects expose protocol-5 NumPy buffers, but the native
`TimeSeriesEnsemble` and `SeismogramEnsemble` pickle state first executes:

```cpp
pickle.dumps(py::list(pybind11::cast(self.member)))
```

That nested pickle turns all members into one monolithic `bytes` field before
Dask sees the object.  It hides the atomic sample buffers from Dask's
frame-based protocol and adds full-payload copies on serialization and
deserialization.

Measured with the current implementation:

- `TimeSeries(1000)` serializes through Dask as a small control frame plus an
  8,000-byte sample frame.
- `Seismogram(1000)` produces a small control frame plus a 24,000-byte sample
  frame.
- The corresponding one-member ensembles produce one monolithic frame of
  roughly 8.6 KB and 24.9 KB, respectively.
- A three-member 10,000-sample TimeSeries ensemble is one ~241 KB frame;
  a prototype ensemble serializer produces a ~1.4 KB control frame and three
  80 KB sample frames.

This is a general Dask transport problem.  It is not required for the bounded
worker-side sink added in #1017: when `completion_on_worker=True` and
`retain_results=False`, the ensemble should not cross the Dask boundary at all.
It also does not affect an ordinary `pickle.dump()` performed inside a worker.

## Proposed fix

Add a Python-level Dask serialization family for exact
`TimeSeriesEnsemble`/`SeismogramEnsemble` objects while leaving the native
pickle format unchanged:

- Serialize a member-free shell carrying Metadata, ErrorLogger, and live/dead
  state.
- Serialize members with Dask/protocol-5 so each sample array remains an
  independent frame.
- Reconstruct into a pre-reserved final ensemble and release temporary member
  references incrementally.
- Include an explicit wire-format version in the serializer header.
- Register lazily from `mspasspy` so both the driver and raw
  `distributed.Client` workers install the dispatch without depending on the
  MongoDB worker plugin.  Importing MsPASS must continue to work when Dask is
  not installed.
- Reject or deliberately fall back for Python subclasses rather than silently
  reconstructing them as a base ensemble.

Ordinary `pickle.dumps` bytes and the on-disk pickle contract must remain
unchanged.  Mixed MsPASS versions in one cluster are unsupported and should
fail clearly rather than silently decode a different wire format.

## Risks and limits

One sample frame per member can be expensive for ensembles containing very
large numbers of tiny members.  The change needs representative benchmarks for
both member count and sample size; frame count, control-header size, elapsed
time, and peak allocation should be recorded.  The receiver still needs Dask
frames plus reconstructed C++ data, so this does not make an arbitrarily large
day object safe.

## Acceptance criteria

- Both ensemble types use the registered Dask serializer and expose multiple
  protocol frames for non-empty sample data.
- Metadata, non-empty ErrorLogger data, processing state, live/dead state,
  member order, and sample values round-trip.
- Empty ensembles and a list containing ensembles round-trip.
- Process-based `LocalCluster(processes=True)` tests prove registration works
  on independent workers for worker-to-driver and driver-to-worker transfer;
  threaded tests alone are insufficient.
- Explicit pickle-only Dask serialization remains monolithic and standard
  `pickle.dumps(..., protocol=5)` is byte-for-byte unchanged by registration.
- MsPASS imports and its non-Dask client mode still work without `distributed`.
- A small benchmark documents the frame-count tradeoff and demonstrates lower
  monolithic-control allocation for representative large members.

wangyinz:
## Summary

- lazily register exact native `TimeSeriesEnsemble` and
  `SeismogramEnsemble` Dask serializers in drivers and independent workers
- keep sample arrays in protocol-5 frames instead of hiding them in one
  ensemble-sized control payload
- reconstruct into a pre-reserved target while releasing temporary member
  references incrementally
- preserve Metadata modification state, ErrorLogger data, processing history,
  member order/state, samples, and ensemble lifecycle
- leave ordinary Python pickle behavior unchanged by registration and fall
  back to pickle for subclasses

The wire format carries an explicit version.  Importing `mspasspy` remains safe
without `distributed`, including `mspasspy.client.Client(scheduler="none")`.

## Validation

- 14 serializer tests passed, including a forced-`spawn`
  `LocalCluster(processes=True)` test with a worker PID distinct from the
  driver
- worker-to-driver and driver-to-worker transfers both use the registered
  serializer
- first lazy registration is isolated in a fresh subprocess and proven not to
  change native state or ordinary pickle bytes
- full member samples and lifecycle/timing/history state are compared
- no-`distributed` client mode is exercised in a subprocess
- independently reviewed; no implementation blocker remained

The documented benchmark records the benefit for several large members and
the frame-count tradeoff for many small members.  Tests condition the legacy
single-frame assertion on the native nested-byte state so the intentional
native-format work in #1021 will not create a brittle future failure.

Closes #1019

### DIFF
--- a/docs/source/user_manual/parallel_processing.rst
+++ b/docs/source/user_manual/parallel_processing.rst
@@ -556,7 +556,7 @@ That same workflow using `sliding_window_pipeline` is this:
    data_out = sliding_window_pipeline(doclist,
                        atomic_example,
                        dask_client,
-                       db.name)
+                       pfunc_args=(db.name,))
 
 For this case the example includes the code to instantiate a MsPASS Client,
 which was omitted in the map example.  It illustrates that `sliding_window_pipeline`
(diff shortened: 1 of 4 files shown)


wangyinz:
Closing this PR after re-evaluating it against the direct-member native pickle change in #1025.

Once #1025 is applied, Dask's default protocol-5 pickle path already exposes the member sample buffers as out-of-band frames. Combined benchmarks found the same frame count and effectively the same wire size with and without this custom serializer. The custom path can still reduce native peak RSS by avoiding the safe C++ member snapshot used by ordinary pickle, but elapsed-time results vary by ensemble shape and it does not reduce network volume.

That narrow benefit does not justify the current global cost: importing `mspasspy` eagerly imports `distributed`, fresh-process measurements added roughly 50 MB of baseline RSS and about 0.24 seconds of startup time, and the PR introduces automatic global registration, a second ensemble state implementation, a private wire-format version, and mixed-version cluster compatibility risk.

It is also not needed for the workflow that motivated this work when `completion_on_worker=True` and `retain_results=False`, because the large ensemble should not cross the Dask boundary.

If a demonstrated production use case later requires large ensembles to cross Dask boundaries, a smaller follow-up should benchmark against #1025, avoid the top-level `distributed` import, and use explicit/Dask-scoped registration. The relevant claim should be limited to native peak-RSS reduction, not fewer frames or bytes.

PR #1024, opened 2026-08-31
Add a bounded EarthScope S3 day-window workflow example

### LINKED ISSUE #1020: Add a bounded-memory EarthScope S3 day-window example workflow
## Context

A collaborator supplied an out-of-tree prototype used to index EarthScope
station-day miniSEED objects, build event windows, and write one output object
per day.  The prototype is not currently part of MsPASS, so the findings below
are not bugs in an existing MsPASS API.  This issue proposes adding a sanitized,
tested example workflow that preserves the useful design while correcting the
confirmed data-integrity and memory-lifecycle defects.

No saved notebook outputs, credentials, local database tables, or private data
should be committed.

### Data integrity

- Day selection is taken from the first arrival in a group.  A later arrival
  crossing midnight can therefore omit its adjacent station-day holding.
- Arrivals are merged with holdings only on `(net, sta)`.  When two daily
  holdings exist, the merge creates an arrival-by-holding Cartesian product and
  duplicates windows.
- S3 indexing calls `list_objects_v2` once.  More than 1,000 objects are
  silently omitted because continuation tokens are never followed.
- The annual query uses an inclusive next-year boundary, duplicating an arrival
  exactly at January 1 in adjacent years.
- One preparation notebook deletes `distinct()[0]` on the assumption that the
  first network is invalid; MongoDB does not guarantee that ordering.
- A station-normalization fallback says it preserves the original station but
  fails to write `sta`.

### Processing correctness and resource lifecycle

- The detrend condition compares the imported function to a string and applies
  detrend to the empty output ensemble instead of the decoded input ensemble.
- Multi-rate input is reconverted from the growing merged stream on every rate
  iteration instead of once after the loop.
- `get_object_with_iris_versions` returns from `finally`, masking original
  failures and even control exceptions, and never explicitly closes the S3
  `StreamingBody`.
- An indexing helper turns its real error into a later unpacking `TypeError` by
  implicitly returning `None`.
- Driver completion creates an unused S3 client every day.
- `nfailed` is unbound for a verbose zero-failure day.
- Worker-plugin teardown calls `.close()` even when setup never installed a
  client.

### Unbounded memory path

The prototype currently builds every station ensemble for a day on one worker,
returns the whole list to the driver, deep-copies all members into another
daily ensemble, and then pickles that entire ensemble.  The default sliding
window may keep about two tasks per worker in flight.  #1017 can keep a
completion on the worker, but simply toggling it is insufficient: the worker
would still hold the full day list, merged copy, and pickle payload at once.

## Proposed feature

Add a maintained EarthScope S3 example with pure, unit-testable helpers and a
bounded worker-side sink:

- Build day/station requests without row-level arrival/holding cross products.
  Union every required day and keep one arrival request per input identity.
- Use the S3 paginator and deterministic ordering/deduplication.
- Reuse one worker-local S3 client, close each response body in `finally`, and
  preserve original exceptions.
- Apply detrend to the decoded input and convert a multi-rate stream once.
- Filter dead/empty members before serialization.
- Process and write one station or bounded batch at a time.  A day output may
  contain a small header followed by independent pickle records; do not create
  a second whole-day ensemble.
- Upload from a bounded temporary file/multipart configuration, return only a
  small status dictionary, and use deterministic object keys so retries are
  idempotent.
- Start the documented pipeline at `sliding_window_size=1`; explain how to
  increase it only after measuring the largest day.
- Correct the year boundary, network filtering, station fallback, zero-failure
  accounting, and plugin teardown behavior.

## Acceptance criteria

- Unit tests with fake MongoDB collections and fake paginated S3 clients cover
  one-day and cross-midnight arrivals, no omissions, no duplicates, and more
  than 1,000 indexed objects.
- Tests prove response bodies are closed on success and error, original
  exceptions/control exceptions are not masked, and failed setup has safe
  teardown.
- Tests prove detrend targets decoded data and multi-rate conversion runs once.
- A process-based Dask test uses an unpickleable station payload and a fake sink
  to prove the large intermediate remains inside one worker while only a small
  status reaches the driver.
- The writer never constructs a whole-day merged ensemble and reads its record
  stream incrementally.
- Examples contain no credentials or saved notebook output and document
  required environment-based authentication.
- Documentation clearly labels this as an example workflow, not a universal
  EarthScope data API.

wangyinz:
## Summary

Add a sanitized, maintained EarthScope S3 example that replaces the supplied
out-of-tree notebook path with a bounded worker-side workflow.

The example:

- paginates and deterministically indexes all station-day objects
- uses half-open year queries and includes adjacent days required by padded
  windows
- groups arrivals without an arrival-by-holding Cartesian product and unions
  every day required by each window
- preserves unmatched station codes and filters invalid network values without
  relying on MongoDB ordering
- reads versioned miniSEED with explicit body cleanup and original-exception
  propagation
- converts multi-rate streams once, detrends the decoded data, and filters
  dead/empty records
- processes bounded station/arrival batches, consumes the generator in the
  same worker, writes independent pickle records to a temporary file, and
  uploads with single-threaded multipart transfer
- returns only small status dictionaries and uses deterministic object keys so
  retries overwrite rather than duplicate output
- uses one worker-local boto3 default-chain client to read the public Open Data
  Program bucket required for TA and write the worker's authorized GeoLab
  scratch bucket
- defaults to `sliding_window_size=1`

The last record/window references are explicitly dropped before constructing
the next window or starting the upload, avoiding a transient overlap through
pybind member wrappers.

## Security and scope

No notebook, saved output, credential, CSS/Antelope table, or private data is
included.  AWS credentials, if needed for the output bucket, are resolved and
refreshed in the worker by boto3's standard provider chain; they are never task
arguments.  The README limits the input path to public ODP networks and labels
this as a focused example rather than a universal EarthScope API.

## Validation

- all 15 tests passed in one run
- the suite includes a real
  `LocalCluster(processes=True)` and an intentionally unpickleable intermediate,
  proving the station payload stays in the worker while the driver receives a
  small dictionary
- coverage includes >1,000 paginated objects, version fallback order,
  cross-midnight and leap-year boundaries, duplicate prevention, body cleanup,
  error/control-exception propagation, detrend/multi-rate behavior, idempotent
  keys, bounded writer behavior, default-chain client creation, and
  failed-plugin setup/teardown
- `py_compile`, Black's formatting check, credential-pattern scan, and
  `git diff --check` passed
- independently reviewed; no blocker remained

Live ODP/scratch access, real CSS/Mongo data, GeoLab disk limits, and production
multipart throughput are intentionally not claimed by the fake-backed tests.
The README requires a measured small-day run first.

Closes #1020

### DIFF
--- a/python/tests/test_earthscope_s3_example.py
+++ b/python/tests/test_earthscope_s3_example.py
@@ -134,6 +134,7 @@ def _holding(net, sta, year, jday, key):
 
 
 def test_paginated_index_is_complete_deduplicated_and_deterministic():
+    assert workflow.EARTHSCOPE_BUCKET == "earthscope-geophysical-data"
     prefix = "miniseed/TA/2014/001/"
     first = [{"Key": f"{prefix}S{i}.TA.2014.001"} for i in range(1000)]
     second = [
--- a/python/tests/test_earthscope_s3_example.py
+++ b/python/tests/test_earthscope_s3_example.py
@@ -1,6 +1,8 @@
+import sys
 from dataclasses import replace
 from io import BytesIO
 from pathlib import Path
+from types import SimpleNamespace
 
 from botocore.exceptions import ClientError
 import dask.distributed as ddist
@@ -427,7 +429,7 @@ def test_failed_worker_setup_has_safe_teardown(monkeypatch):
     worker = FakeWorker()
     plugin = EarthScopeS3Worker()
 
-    def fail_setup():
+    def fail_setup(**kwargs):
         raise RuntimeError("setup failed")
 
     monkeypatch.setattr(worker_support, "create_earthscope_s3_client", fail_setup)
@@ -440,6 +442,58 @@ def test_failed_worker_setup_has_safe_teardown(monkeypatch):
     plugin.teardown(worker)
     assert client.closed
     assert plugin.key not in worker.data
+
+
+def test_earthscope_sdk_session_owns_refresh_and_receives_role(monkeypatch):
+    captured = {}
+    s3_client = object()
+
+    class FakeSession:
+        def client(self, service, config=None):
+            captured["client"] = (service, config)
+            return s3_client
+
+    class FakeUser:
+        def get_boto3_session(self, *, role):
+            captured["scope"] = {"role": role}
+            return FakeSession()
+
+    class FakeEarthScopeClient:
+        def __init__(self):
+            self.user = FakeUser()
+
+    monkeypatch.setitem(
+        sys.modules,
+        "earthscope_sdk",
+        SimpleNamespace(EarthScopeClient=FakeEarthScopeClient),
+    )
+
+    assert (
+        worker_support.create_earthscope_s3_client(role="s3-miniseed-v2") is s3_client
+    )
+    assert captured["scope"] == {"role": "s3-miniseed-v2"}
+    assert captured["client"] == ("s3", worker_support.S3_CONFIG)
+
+    assert worker_support.create_earthscope_s3_client() is s3_client
+    assert captured["scope"] == {"role": "s3-miniseed-v2"}
+
+
+def test_worker_plugin_forwards_role(monkeypatch):
+    captured = {}
+    client = CloseTracker()
+
+    def create_client(**kwargs):
+        captured.update(kwargs)
+        return client
+
+    monkeypatch.setattr(worker_support, "create_earthscope_s3_client", create_client)
+    worker = FakeWorker()
+    plugin = EarthScopeS3Worker(role="test-role")
+
+    plugin.setup(worker)
+
+    assert worker.data[plugin.key] is client
+    assert captured == {"role": "test-role"}
 
 
 def test_worker_completion_keeps_unpickleable_payload_in_process():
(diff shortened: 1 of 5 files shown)


wangyinz:
Closing this PR because it is neither a validated fix for the supplied out-of-tree workflow nor ready to be maintained in the core repository.

The current Python 3.10 and 3.13 CI jobs fail during collection with `ModuleNotFoundError: No module named 'scripts'`; `MANIFEST.in` also prunes `scripts`, so the added test never validates an installed package. The local tests use fake S3/MongoDB objects and do not exercise the real EarthScope/GeoLab authentication, miniSEED-to-MsPASS path, worker plugin installation, scratch upload, native serialization, or peak RSS.

The documented GeoLab invocation also passes `SCRATCH_BUCKET` directly as boto3's `Bucket=` value. GeoLab supplies a full `s3://bucket/user-prefix` path, so the implementation must parse the bucket and preserve the user prefix; the original `s3fs.open(full_uri)` contract did this correctly.

The replacement is not drop-in compatible with the original workflow: it changes daily `TimeSeriesEnsemble` pickle output into station-batch `.pklstream` output. Its bounded-memory claims also remain unverified with real native data, and review found remaining overlap between source/decoded streams and adjacent generator records.

A project-specific EarthScope/CSS workflow of this size belongs in `mspass_tutorial` or a dedicated workflow repository, where it can be tested on a real GeoLab small day, representative day, and heaviest day before publication. Small, independently reusable core or documentation changes can return as separate PRs.

Issue #1037, opened 2026-09-13
Undocumented change to how we handle Response data


pavlis:
Somewhere in the recent flurry of changes made with the AI to fix some old problem an unexpected change happened.   Previously the method `Database.save_inventory` saved the response data for a channel to an attribute with the key "serialized_channel_data".   It still does that but the (default) format it uses changed completely.   Previously the obspy Response object was serialized with pickle.dumps.   It seems we now have a new module in `mspasspy.db.serialization` called `encode_inventory`.   Apparently that converts the response data back to StationXML using an obspy function and saves that as the value associated with "serialized_channel_data".   The new method to restore "serialized_channel_data" from a document in channel is this code snippet:
```
from mspasspy.db.serialization import decode_response
# example just pulling first document in channel
doc = db.channel.find_one()
response_doc = doc["serialized_channel_data"]
resp = decode_response(response_doc)
```
Poking around the code it seems there is some way to make the backward compatible to handle data stored previously with pickle.   I haven't really dug into that as I don't think it is that important.   I would always recommend most uses should use web services to build the channel metadata anyway to assure it is always as current as possible since thing do get changed there.   The only way that would enter then is if reading old channel collection data saved in something like a json dump.   

I am posting this as there are two things we need to do to follow up on this:
1.  I haven't checked but presume the user manual is out of date on this topic.  Somewhere, I know, it discusses handling response data and the whole issue of using web services top populate the channel and site collections.   We need to clean any misinformation there. 
2. The mspass_tutorials are a bigger problem as unless something was done there is misinformation there too.

The good news is it looks like the calib.py module addressed this change and that is not part of this issue.  For the record, that comes up because the "calib" attribute to scale data from counts to nm/s is contained in the Response object that is created by `decode_response`.   It appears that was changed when `save_inventory` was changed, 

This not a high priority issue as it seems to be restricted to a documentation problem but we do need to address it at some point.  For now just keeping it as an issue will address the misinformation.

pavlis:
I see something else this code change did that is worth discussion.  I see the Inventory object is serialized into the site collection.   I think that is a really bad idea for three reasons:
1.  That monstrosity seriously bloats that collection.  It increases it size by at least one order of magnitude. 
2. It is a collision in concept.  The obspy.Inventory object was not designed to serve that role.   It was designed to be a combination of what we do with site and channel.  
[...]

PR #1038, opened 2026-09-13
Remove serialized Inventory payloads from site records

wangyinz:
## Summary

- stop `Database.save_inventory` from writing `serialized_inventory` into `site` documents
- keep the versioned Response payload in `channel.serialized_channel_data` unchanged
- update inventory/location contract tests to require that `site` documents do not contain `serialized_inventory`
- retain the existing `read_inventory` decoder path for backward compatibility with databases that already contain legacy/new `serialized_inventory` records

## Rationale

This follows the latest discussion in #1037. The `site` collection is intended to carry receiver metadata used for Seismogram normalization; embedding an ObsPy Inventory there duplicates channel/response data, substantially increases document size, and mixes the site and channel concepts.

The Response serialization change introduced by #989 is intentionally left intact. New channel records continue to use the versioned Response codec and existing legacy pickle records remain readable.

## Validation

- Python syntax compilation passed for the modified database module and tests.
- Repository CI should exercise the MongoDB-backed contract tests.

Refs #1037


python/mspasspy/db/database.py
```diff
@@ -7693,9 +7691,6 @@ def save_inventory(self, inv, networks_to_exclude=["SY"], verbose=False):
                         rec["edepth"] = next(iter(location_depths))
                     rec["starttime"] = loc_stime.timestamp
                     rec["endtime"] = loc_etime.timestamp
-                    rec["serialized_inventory"] = encode_inventory(
-                        inventory_subset(inv, x, station, loc_channels)
-                    )
```
pavlis, line 7698: Digging into this I think completely deleting contents of inventory goes too far.  I discovered there is some important Metadata stored in the Inventory object we are currently dropping because it is part of obspy's "Station" object that is between the Network and Channel layer of Inventory.   That is, Inventory is this nasty hierarchy that is just an image of the hierarchy of stationxml that is the son of SEED, which we all know is a 4 letter word.   That commentary aside, note we are currently storing the "Response" object image serialized to channel data.  That is appropriate as the Response object in Inventory is at the level:  Inventory->Network->Station->Channel.   The problem we have is that a Station object (one up in the hierarchy) has Metadata we are currently dropping that we do need to save.   dir of a Station object Ioaded into memory has this content (dropping all the private attributes with _):
```
'alternate_code',
 'channels',
 'code',
 'comments',
 'copy',
 'creation_date',
 'data_availability',
 'description',
 'elevation',
 'end_date',
 'equipments',
 'external_references',
 'geology',
 'get_contents',
 'historical_code',
 'identifiers',
 'is_active',
 'latitude',
 'longitude',
 'operators',
 'plot',
 'restricted_status',
 'select',
 'selected_number_of_channels',
 'site',
 'source_id',
 'start_date',
 'termination_date',
 'total_number_of_channels',
 'vault',
 'water_level'
```
I'm not sure we need to save all of those.   We especially don't want the `channels` attribute as that is actually a python list of Channel objects.   It turns out we are dropping some useful Metadata from Channel objects as well.   dir of a channel object shows this (again without the private attributes):
```
'alternate_code',
 'azimuth',
 'calibration_units',
 'calibration_units_description',
 'clock_drift_in_seconds_per_sample',
 'code',
 'comments',
 'copy',
 'data_availability',
 'data_logger',
 'depth',
 'description',
 'dip',
 'elevation',
 'end_date',
 'equipment',
 'equipments',
 'external_references',
 'historical_code',
 'identifiers',
 'is_active',
 'latitude',
 'location_code',
 'longitude',
 'plot',
 'pre_amplifier',
 'response',
 'restricted_status',
 'sample_rate',
 'sample_rate_ratio_number_samples',
 'sample_rate_ratio_number_seconds',
 'sensor',
 'source_id',
 'start_date',
 'storage_format',
 'types',
 'water_level']
```
We are currently extracting latitude, longitude, and elevation and saving those is channel complaining when those coordinates do not match those in Station.  I think we also extract the start_data and orientation data.  The point is, however, that there are some obvious things we aren't extracting from the Channel object that we are actually dropping.   (note saving the Inventory object does store those, but the original motivation of this change is doing that bloated site unnecessarily which is still true)  One's I see are:  sample rate, equipment, and the sensor attribute (which actually points to a Sensor object at the bottom of this hierachy).   

This is messy enough I recommend we do this:
1.  Change the channel code to add a serialization of the Channel object it is parsing to create a channel document.   It should continue to post the Response object inside the Channel object but serialize the higher level data structure as well (Channel) to save metadata that may be important for some applications.  
2. We need to at least add an extract of the sample rate for each channel.   Otherwise you have to dig into the Response object and infer it.   
3.  Instead of serializing Inventory, which is the issue that initiated this pull request change, we should actually serialize the Station object and make that an attribute of the site collection.   That would be completely consistent with serializing Channel for a channel  collection.

pavlis (commented):
See the lengthy comment posted in the review.  When I dug a bit deeper I realized completely dropping serialization of Inventory went to far.   Inventory preserves all station metadata but doesn't really organize it consistently.  See my suggestions for what I think is the right way to do this for consistency.

pavlis:
Fellow mspass group members.  I will volunteer to fix the suggestion I made in the review posted yesterday.  I suggest we finalize a plan in our next weekly meeting and if you all agree I'll work to finalize this pull request.

